Issue #53, opened 2026-02-24
Fix intermittent test failure


gunnarmorling:
Seeing this failure intermittently during the release process (e.g. [here](https://github.com/hardwood-hq/hardwood/actions/runs/22366172037/job/64732651723#step:5:284))

```
 Error: [ERROR] Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.028 s <<< FAILURE! -- in dev.hardwood.internal.reader.ColumnAssemblyBufferTest
Error: [ERROR] dev.hardwood.internal.reader.ColumnAssemblyBufferTest.testAssemblyThreadExceptionPropagatedToConsumer -- Time elapsed: 0.010 s <<< FAILURE!
[INFO] java.lang.AssertionError: 
[INFO] [First batch should be available] 
[INFO] Expecting actual not to be null
[INFO] 	at dev.hardwood.internal.reader.ColumnAssemblyBufferTest.testAssemblyThreadExceptionPropagatedToConsumer(ColumnAssemblyBufferTest.java:55)
```

I thought fixing #51 was enough, but not quite so.

Issue #51, opened 2026-02-24
Fix race condition in ColumnAssemblyBufferTest


gunnarmorling:
Just ran into this test failure while [testing the release pipeline](https://github.com/hardwood-hq/hardwood/actions/runs/22361723914/job/64716787224):

```
Error:  Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.034 s <<< FAILURE! -- in dev.hardwood.internal.reader.ColumnAssemblyBufferTest
Error:  dev.hardwood.internal.reader.ColumnAssemblyBufferTest.testAssemblyThreadExceptionPropagatedToConsumer -- Time elapsed: 0.023 s <<< ERROR!
java.lang.NullPointerException: Cannot invoke "java.util.List.size()" because "this.pages" is null
	at dev.hardwood.internal.reader.ColumnAssemblyBufferTest$ErrorInducingPageCursor.hasNext(ColumnAssemblyBufferTest.java:126)
	at dev.hardwood.internal.reader.PageCursor.runAssemblyThread(PageCursor.java:160)
	at java.base/java.lang.VirtualThread.run(VirtualThread.java:456)
```

I think this happens when the thread in the parent `PageCursor` fetches data before the constructor of `ErrorInducingPageCursor` has completed.

PR #69, opened 2026-02-26
#53 Decouple PageCursor Assembly Thread Initiation to Avoid Race Conditions

rionmonster:
### Description
This pull request addresses #53 which details a potential race-condition that would intermittently arise, specifically during CI builds. The underlying issue itself was the result of the `PageCursor` constructor starting the assembly thread _before_ the construction was complete. 

### Key Changes
- Added two new static `PageCursor.create()` factory methods (single-file, multi-file) which are now responsible for construction of the instance _and_ subsequent thread initialization.
- Updated existing test cases within the `ColumnAssemblyBufferTest` class to no longer use a static, subclassed instance in favor of a simpler message buffer to handle the processing.
  - Removed some of the previously `@Disabled` test cases that were updated within #51 prior to this fix.

### Verification
To reproduce the initial issue a new `ColumnAssemblyBufferTest.testAssemblyThreadRaceCondition` test was temporarily introduced ([fe4e0b8](https://github.com/hardwood-hq/hardwood/pull/55/commits/fe4e0b8e361069b4ac45124b2a54bfcc33d49e84)) in conjunction with an intentionally slow `PageCursor` to simulate the race-condition successfully:

Screenshot 2026-02-24 at 1 58 00 PM

After verifying the test failed as expected, the previously mentioned `PageCursor.create()` factory methods were introduced to verify the fix and all previously disabled tests were passing as expected:

image

#### Regression Verification
There appears to be no noticeable regressions in terms of performance as it relates to these changes and the project successfully builds:

```
PERFORMANCE TEST RESULTS

Environment:
  CPU cores:       14
  Java version:    25.0.2

Data:
  Files processed: 119
  Total rows:      651,209,003
  Total size:      9,241.1 MB
  Runs per contender: 5

Performance (all runs):
  Hardwood (multifile) [1]               2.28     285,994,292         20,428,164       4058.5
  Hardwood (multifile) [2]               2.39     272,130,799         19,437,914       3861.7
  Hardwood (multifile) [3]               2.14     304,303,272         21,735,948       4318.3
  Hardwood (multifile) [4]               2.13     305,588,458         21,827,747       4336.5
  Hardwood (multifile) [5]               2.14     304,587,934         21,756,281       4322.3
  Hardwood (multifile) [AVG]             2.22     293,999,550         20,999,968       4172.1
                                   min: 2.13s, max: 2.39s, spread: 0.26s
```

### DIFF
--- a/core/src/main/java/dev/hardwood/internal/reader/PageCursor.java
+++ b/core/src/main/java/dev/hardwood/internal/reader/PageCursor.java
@@ -64,52 +64,39 @@ public class PageCursor {
     private final ColumnAssemblyBuffer assemblyBuffer;
 
     /**
+     * Creates a PageCursor for single-file reading with optional eager assembly.
+     * Starts the assembly thread before returning if an assembly buffer is provided.
+     */
+    public static PageCursor create(List<PageInfo> pageInfos, HardwoodContextImpl context,
+                                    String fileName, ColumnAssemblyBuffer assemblyBuffer) {
+        PageCursor cursor = new PageCursor(pageInfos, context, null, -1, fileName, assemblyBuffer);
+        cursor.startAssemblyThread();
+        return cursor;
+    }
+
+    /**
+     * Creates a PageCursor with multi-file support and optional eager assembly.
+     * Starts the assembly thread before returning if an assembly buffer is provided.
+     */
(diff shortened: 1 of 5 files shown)


gunnarmorling (approved):
LGTM, merging.

Thanks a lot, @rionmonster!

Issue #60, opened 2026-02-25
Adjust to relocation of org.lz4:lz4-java:jar:1.8.1


gunnarmorling:
Getting this warning in the build:

```
 The artifact org.lz4:lz4-java:jar:1.8.1 has been relocated to at.yawk.lz4:lz4-java:jar:1.8.1
```

PR #72, opened 2026-02-27
#60 Updated lz4-java Dependency and Related Test Suite

rionmonster:
### Description
This pull request addresses #60 which pertains to a minor dependency relocation for `lz4-java` which was recently relocated (from `org.lz4:lz4-java` to `at.yawk.lz4`) to get rid of the following build warning:

```
 The artifact org.lz4:lz4-java:jar:1.8.1 has been relocated to at.yawk.lz4:lz4-java:jar:1.8.1
```

### Key Changes
- Updated all `lz4-java` dependencies to use the correct group (e.g. `<groupId>at.yawk.lz4</groupId>`).
- Added a new suite of LZ4-related decompression tests within the newly created `Lz4DecompressorIT` (following the pattern established in the similarly named Libdeflate tests.
- Updated parent package for all LZ4-related classes under a new `lz4` package (again similar to the pattern established by Libdeflate).
  
### Verification
The newly added suite of LZ4-related tests were all passing as expected in addition to the complete build and all existing tests:

image


core/src/main/java/dev/hardwood/internal/compression/lz4/Lz4Loader.java
```diff
@@ -0,0 +1,51 @@
...
+ *  Copyright The original authors
+ *
+ *  Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package dev.hardwood.internal.compression.lz4;
+
+/**
+ * Checks availability of the LZ4 library on the classpath.
+ */
+public final class Lz4Loader {
```
gunnarmorling: Out of curiosity, is there any specific need for introducing this, over just keep using the existing `checkClassAvailable()` pattern in the factory? The reason for the libdeflate case being handled differently is that we need to manage the lifecycle of the pool object in the Hardwood context instance (I'm not too happy with that particular design, it all screams "tight coupling", but it does the job for now).

gunnarmorling (commented):
Nice one, @rionmonster! Great to add some unit tests, too (this aspect definitely has been a bit neglected so far in favor of relying on the end-to-end tests). A few suggestions for your consideration inline.

Issue #76, opened 2026-02-28
Support page CRC validation


gunnarmorling:
Add support for validating page integrity using the optional CRC field in [`PageHeader`](https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift) (field 4).

When present, this is a CRC-32C checksum of the serialized page (header + compressed data). Currently skipped by `PageHeaderReader`.

PR #86, opened 2026-03-01
#76 Add Support for Parquet CRC Verification

rionmonster:
### Description
This pull request addresses #76 which adds support for validating the CRC-32 checksum that is optionally provided in Parquet files (specifically Field 4 of [`PageHeader`](https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift)) to ensure data integrity.

### Key Changes
- Added a new `crc` field for the existing `PageHeader` class to represent the optional CRC header value, which would be used as a flag to trigger validation paths with associated reading in `PageHeader` initialization.
- Introduced a new static `CrcValidator.assertCorrectCrc()` function for comparing the expected page data against the provided CRC value (throwing an `IOException` when invalid)
- Added CRC validation to the following paths:
  - Dictionary-page scanning (in `PageScanner.scanPagesSequentially()`)
  - Index-based scanning (in `PageScanner.scanPagesFromIndex()`)
- Added the new `CrcValidationTest` test suite for verifying the expected behaviors (both happy path and otherwise) as well as datagen-related cases (specifically cases for plain and dictionary encoding)
  
### Verification
As mentioned the newly added `CrcValidationTest` covers the following cases:
- `testReadFileWithCrc` to verify a simple, happy-path CRC test case for plain file requiring validation.
- `testCorruptedDataDetected` to simulate corruption within a file requiring validation (via explicit byte-flipping).
- `testReadDictionaryFileWithCrc` to verify happy-path CRC validation for a dictionary file that requires validation.
- `testCorruptedDictionaryPageDetected` to simulate corruption for dictionary files requiring validation.
- `testFilesWithoutCrc` to verify non validation behavior still works as expected (regression check).

All of which pass as expected:

image

### DIFF
--- a/core/src/main/java/dev/hardwood/internal/reader/CrcValidator.java
+++ b/core/src/main/java/dev/hardwood/internal/reader/CrcValidator.java
@@ -14,10 +14,11 @@ import java.util.zip.CRC32;
 class CrcValidator {
 
     /**
-     * Validates that the CRC-32 checksum of the given page data matches the expected value
-     * from the page header. The buffer's position and limit are not modified.
+     * Asserts that the CRC-32 checksum of the given page data matches the expected value
+     * from the page header. Throws if the checksum does not match.
+     * The buffer's position and limit are not modified.
      */
-    static void validate(int expectedCrc, ByteBuffer pageData, String columnName) throws IOException {
+    static void assertCorrectCrc(int expectedCrc, ByteBuffer pageData, String columnName) throws IOException {
         CRC32 crc = new CRC32();
         crc.update(pageData.duplicate());
         int actualCrc = (int) crc.getValue();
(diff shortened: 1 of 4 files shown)


core/src/main/java/dev/hardwood/internal/reader/CrcValidator.java
```diff
@@ -0,0 +1,30 @@
...
+import java.nio.ByteBuffer;
+import java.util.zip.CRC32;
+
+class CrcValidator {
+
+    /**
+     * Validates that the CRC-32 checksum of the given page data matches the expected value
+     * from the page header. The buffer's position and limit are not modified.
+     */
+    static void validate(int expectedCrc, ByteBuffer pageData, String columnName) throws IOException {
```
gunnarmorling: How about naming it `assertCorrectCrc()` or something along these lines, implying that the method raises an exception if there's a mismatch?

Issue #31, opened 2026-02-17
Support reading files off of object storage


gunnarmorling:
It's a common requirement to parse files on object storage (S3), and instead of pulling them down completely to a local file system, only the required file segments (specific columns when using projections, in the future also specific row groups via predicate push-down) should be fetched. We currently use the JDK NIO `Path` API for accessing files, which by default assumes that files are local.

I can see two ways for reading file segments from object storage:

* Establish a Hardwood abstraction over file locations, with the current `Path` implementation and an alternative S3 implementation
* Use the [AWS Java NIO SPI for S3](https://github.com/awslabs/aws-java-nio-spi-for-s3), which provides S3 access via NIO

I am leaning towards the latter, as it may be a relatively easy way for achieving our goal here, although I am not sure how it would work with our current scheme of using memory mapped file access.

# Dependencies

Both options will add a dependency to the project (the AWS S3 SDK in either case, additionally, the NIO SPI implementation in the latter case), which always should be a decision carefully made for this project. I think it is acceptable for this case, as it is going to be an optional dependency (similar to the supported compression libraries), to be added by a user only if and when they actually need it.

Alternatively, we could explore accessing files on S3 via HTTPS using the JDK's built-in HTTP client; file changes are retrievable via HTTP range requests, AFAIK. There may be complexities around authentication for this approach, though.

PR #87, opened 2026-03-01
Allowed FileReader to open memory buffer

ngrondin:
The objective was to be able to read small(ish) Parquet files stored in memory, received from a stream. A common use case would be to read files stored in S3 or Azure Blobs. 

Previously, the application would have needed to stream the file in and then save it to a temporary disk before reading it with ParquetFileReader, which may not be best practice in stateless docker containers. 

With this change, the bytes can be stored in a direct ByteBuffer and then this buffer is used for reading the parquet file without having to write to local disk.

For this change, I had to refactor all classes that were expecting a MappedByteBuffer to use a ByteBuffer instead, which is the super class of MappedByteBuffer. All methods used were already belonging to ByteBuffer, so no functionality change was required passed the ParquetFileReader class.

In the ParquetFileReader class, a new version of open() was added with a new private constructor.

I left the Path attribute, but it is null when the file is read from memory.

### DIFF
--- a/core/src/main/java/dev/hardwood/internal/reader/ParquetMetadataReader.java
+++ b/core/src/main/java/dev/hardwood/internal/reader/ParquetMetadataReader.java
@@ -38,7 +38,7 @@ public final class ParquetMetadataReader {
     /**
      * Reads file metadata from a memory-mapped buffer covering the entire file.
      *
-     * @param fileMapping the memory-mapped buffer of the entire file
+     * @param buffer the buffer of the entire file
      * @param path the file path (used for error messages)
      * @return the parsed FileMetaData
      * @throws IOException if the file is not a valid Parquet file
--- a/performance-testing/micro-benchmarks/src/main/java/dev/hardwood/benchmarks/PageHandlingBenchmark.java
+++ b/performance-testing/micro-benchmarks/src/main/java/dev/hardwood/benchmarks/PageHandlingBenchmark.java
@@ -8,6 +8,7 @@
 package dev.hardwood.benchmarks;
 
 import java.io.IOException;
+import java.nio.ByteBuffer;
 import java.nio.MappedByteBuffer;
 import java.nio.channels.FileChannel;
 import java.nio.file.Path;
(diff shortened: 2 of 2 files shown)


gunnarmorling:
Hey @ngrondin, thanks a lot for proposing this. I don't remember from the top of my head, but I thought there was a reason why it's declared as `MappedByteBuffer` everywhere. Let's see how CI handles the change.

> A common use case would be to read files stored in S3 or Azure Blobs.

Note we have #31 specifically for this feature request already. @iamkarasik is looking into this atm. I think it's gonna be a larger change than what's proposed here, as we'd want to avoid having to fetch files entirely, rather we'd just do range requests for the actually required segments of the file (currently, as per the specified column projections, in the future also considering push-down predicates). The design for this is WIP right now.

iamkarasik:
@gunnarmorling this change is going to be necessary even for the S3 implementation I'm working on. Currently even the `Decompressor` interface is too tightly coupled to the Memory-mapped file implementation (`MappedByteBuffer`). I've made the same change locally as well, and I believe it should be a backwards compatible change to use a `ByteBuffer`

gunnarmorling:
Thanks all!

I have created https://github.com/hardwood-hq/hardwood/pull/99 which supersedes this PR. It contains your commit, @ngrondin, for supporting of any `ByteBuffer`-backed files. I have slightly reworked that commit to wrap up a few lose ends.

In addition, there's now an `InputFile` abstraction which can represent different kinds of files, with `Path`-backed and `ByteBuffer`-backed implementations being the two currently provided options. In the future, there'll be an S3-backed `InputFile`, too.

Your usage becomes this one:

```
ByteBuffer buffer = ...;

try (ParquetFileReader reader = ParquetFileReader.open(InputFile.of(buffer))) {
  // ...
}
```

> I acknowledge that this wouldn't work for larger files where partial access or full streaming is required

I think the `InputFile` abstraction will take care of it, as it provides a method `ByteBuffer readRange(long offset, int length)` which lets the reader request any required sections of the file.

Gonna close this PR in favor of the new one; any feedback/comments on it welcome! Thanks again!

Issue #73, opened 2026-02-27
Use Offset Index for direct page location lookup


gunnarmorling:
When a Parquet file contains an [Offset Index](https://parquet.apache.org/docs/file-format/pageindex/#offset-index), use it to locate data pages by direct lookup instead of sequentially scanning all page headers through the column chunk.

This eliminates the sequential I/O overhead of reading every Thrift-encoded page header, which becomes significant for large column chunks with many pages.

Scope: Offset Index only (not Column Index, which is for predicate pushdown).

PR #93, opened 2026-03-01
#73 Using offset index if present for discovering page boundaries

gunnarmorling:
While it yields a nice speed-up to the actual scanning, I didn't observe any relevant difference for end-to-end parsing. This this will become more significant when retrieving files (or rather, chunks) from remote storage. Here are the JMH results from the new benchmark:

```
Benchmark                                                                    (dataDir)                    (fileName)  Mode  Cnt  Score   Error  Units
PageScanBenchmark.scanPages  performance-testing/test-data-setup/target/benchmark-data  page_scan_with_index.parquet  avgt   10  0.059 ± 0.005  ms/op
PageScanBenchmark.scanPages  performance-testing/test-data-setup/target/benchmark-data    page_scan_no_index.parquet  avgt   10  0.172 ± 0.003  ms/op
```


gunnarmorling:
Tagging you for awareness, @rionmonster, if you're interested.

PR #127, opened 2026-03-21
MultiFileParquetReader test runner

mghildiy:
Change summary:

- Added new test class(MultiFileParquetReaderComparisonTest) to compare results of MultiFileParquetReader wrt parquet java.
- Moved code to Utils.java to make it reusable (still to be reflected in ParquetComparisonTest).
- Change in FileManager to use column index instead of column name for schema lookup in method scanAllProjectedColumns to correctly handle files with same leaf column names for multiple columns


gunnarmorling:
Hey @mghildiy, thanks a lot! Great work, this uncovered an actual issue for nested schemas. I realized the fix needed to be a bit broader (in general, field look-ups need to be done using the full path, not just the leaf name), so I've created PR #134 on top of yours. Gonna close this one here, the other one contains your commits. Thanks again, this is very useful!

PR #172, opened 2026-03-28
Add in predicate support for row group filter pushdown

### LINKED ISSUE #177: Add IN predicate support for filter pushdown
Add support for IN predicates in row group and pagelevel filter pushdown.

ArnavBalyan:
- Enable filtering with row group pruning via column statistics
 - A row group is skipped only when all values in the set fall outside the column's min/max range. 
 - Supports int, long, and binary/string columns.
 - Closes #177

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -19,7 +19,7 @@ Design documents describe the intended end state. Do not include references to t
 
 # Coding
 
-When adding or changing public APIs, update the usage documentation in the README.md accordingly.
+When adding or changing public APIs, update the usage documentation under docs/ accordingly.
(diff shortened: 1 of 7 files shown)


gunnarmorling:
Thanks a lot, @ArnavBalyan! Pushed a commit for a few minor improvements and merged.

gunnarmorling:
Hey @ArnavBalyan, (late) follow-up question on this one: why is it `inStrings()` vs. `in()` for all the others? I didn't notice back then, but it feels a bit irregular and I don't see a specific reason for it right now. Thanks for any reasoning, if you remember it!

Issue #78, opened 2026-02-28
Support geospatial statistics and bounding box metadata


gunnarmorling:
Add support for reading [geospatial metadata](https://github.com/apache/parquet-format/blob/master/Geospatial.md) from Parquet files:

- **`GeospatialStatistics`** in `ColumnMetaData` (field 17) — per-column geospatial stats
- **`BoundingBox`** — min/max bounds for X, Y, and optional Z/M dimensions
- **`geospatial_types`** — list of geometry type codes present in the column

This is a prerequisite for spatial predicate pushdown on Geometry/Geography columns.

mghildiy:
Raised [PR](https://github.com/hardwood-hq/hardwood/pull/173) for this.

gunnarmorling:
Great stuff, thank you, @mghildiy!

PR #173, opened 2026-03-28
Support geospatial statistics and bounding box metadata

mghildiy:
Added support for reading geospatial metadata #78 :

- GeospatialStatisticsReader to read geospatial metadata captured by GeospatialStatistics
- BoundingBoxReader to read boundingbox metadata captured by BoundingBox
- simple-datagen.py extended to generate test parquet file containing geospatial metadata


gunnarmorling (commented):
Thanks a lot! Some comments inline; in addition, could you also update the docs (under /docs) and exclude the regenerated Parquet test files, this should not be needed. Thx!

mghildiy:
> Thanks a lot! Some comments inline; in addition, could you also update the docs (under /docs) and exclude the regenerated Parquet test files, this should not be needed. Thx!

I couldn't find any /docs directory, but I can see /_designs. Should I add a new doc here explaining how to extend simple-datagen.py, excluding changed parquet files from commit  and so on?

gunnarmorling:
@mghildiy, pushed a commit with some minor fixes as I was taking another look.

Taking a step back though, what this PR is missing are the actual geo logical types; without those, reading the stats isn't really useful; i.e. we're taking step 2 before step 1 here. Can you amend this PR accordingly?

gunnarmorling:
@mghildiy, what should we do with this one? DYT it's good for another review round (apart from the needed rebase)?

mghildiy:
> @mghildiy, what should we do with this one? DYT it's good for another review round (apart from the needed rebase)?

We can have another round of review for it. I would commit the changes after rebase.

gunnarmorling:
Thanks a lot, @mghildiy! I've applied some fixes to resolve a few loose ends, the result is https://github.com/hardwood-hq/hardwood/pull/413, superseding this PR. Gonna close the one here and merge the other one, it retains your authorship.

Issue #182, opened 2026-03-29
Record-level filtering for scalar predicates


gunnarmorling:
## Summary

Predicate pushdown currently operates at row group and page level only. Rows from surviving pages are returned to the caller without evaluating the filter against individual records. This means `eq("id", 42)` returns all rows from pages whose min/max range includes 42, not just the matching rows.

parquet-java applies the filter at three levels: row group, page, and record. Users passing a filter to `createRowReader(filter)` would reasonably expect only matching rows to be returned.

## Proposal

Add record-level filter evaluation for scalar predicates (`eq`, `notEq`, `lt`, `ltEq`, `gt`, `gtEq`, `in`) after page decoding. For each decoded row, evaluate the predicate and skip non-matching rows before returning them to the caller.

## Acceptance criteria

- Scalar filter predicates are evaluated at the record level
- `createRowReader(filter)` returns only rows matching the filter
- No regression in bulk-scan performance when no filter is set

rionmonster:
@gunnarmorling 

I'll carve out some time and take a look at this. I agree it's an important add, and hopefully not a ton of effort involved.

PR #186, opened 2026-03-29
#182 Add Support for Record-Level Scalar Predicate Filtering

rionmonster:
### Description 
This pull request adds record-level filtering for scalar predicates (#182), the third and finest level of the predicate pushdown hierarchy. A new `RecordFilterEvaluator` decides whether to skip record-level decoded row values (similar to `PageFilterEvaluator` for pages and `RowGroupFilterEvaluator` for row groups).

### Key Changes 
The primary change related to this effort is the previously mentioned introduction of the `RecordFilterEvaluator` class, responsible for performing the record-level filtering operations (and the related reader-level plumbing to apply the filters). 

A high-level summary of the changes can be found below: 
- Added a new `RecordFilterEvaluator` class responsible for evaluating a FilterPredicate against individual row values using the flat cached arrays from AbstractRowReader.
  - Supports all scalar types (int, long, float, double, boolean, binary), IN predicates, and compound predicates (AND/OR/NOT).
  - Null values conservatively don't match; unknown columns conservatively pass through.
- Extended reader support for record-level predicate evaluation (via newly introduced AbstractRowReader.filterPredicate)   
- Added a new `RecordFilterEvent` JFR event to capture record-level filtering metadata (e.g. records evaluated, retained, and skipped) to gauge filtering efficiency. 

### Verification & Testing 
Added a large suite of tests (and extensions of previously existing tests:
- Added a new `RecordFilterEvaluatorTest`, which covers all operators and types (parameterized) as well as IN predicates (int, long, binary), null handling, unknown column behavior, and compound predicates (AND/OR/NOT).
- Added `PredicatePushDownTest`, which extended with record-level integration tests verifying exact row values returned (not just counts) for filters that fall within a row group's range, across both single-file and multi-file reader paths. 
- Added a new `RecordFilterBenchmarkTest` to gauge three separate tiers of filtering (no-filter, match-all filter, and a selective filter) to provide an apples-apples comparison between the execution paths.
- Updated `S3InputFileTest` with a case to verify explicit record-level filtering (e.g. return just a single filtered record)

### Performance
In terms of throughput using the typical performance test suite, there appeared to be no regressions (for non-filtered test cases). A new `RecordFilterBenchmarkTest` was introduced to compare performance/overhead between which yielded the following results:

| Contender                       | Avg Time (ms) |        Rows | Records/sec   | vs Baseline      |
|---------------------------------|---------------|-------------|---------------|------------------|
| No filter (baseline)            |          19.8 |  10,000,000 | 669,946,816   | —                |
| Match-all filter (worst case)   |          65.3 |  10,000,000 | 202,588,583   | +230% overhead   |
| Selective filter (id < 1%)      |           4.3 |     100,000 |  31,091,817   | 4.6x faster      |

After the initial batch-based bitset optimization (a stepping-stone towards a more optimal vectorized approach), the worst-case overhead was reduced _significantly_ (230% overhead to 133% overhead):

| Contender                       | Time (ms) |        Rows | Records/sec   | vs Baseline      |
|---------------------------------|-----------|-------------|---------------|------------------|
| No filter (baseline)            |      16.6 |  10,000,000 | 601,434,409   | —                |
| Match-all filter (worst case)   |      38.6 |  10,000,000 | 258,948,007   | +133% overhead   |
| Selective filter (id < 1%)      |       1.9 |     100,000 |  53,399,777   | 8.7x faster      |

## Checklist

### DIFF
--- a/core/src/main/java/dev/hardwood/internal/reader/RecordFilterEvaluator.java
+++ b/core/src/main/java/dev/hardwood/internal/reader/RecordFilterEvaluator.java
@@ -32,13 +32,34 @@ import dev.hardwood.reader.FilterPredicate.Or;
 /// matching the predicate are returned to the caller.
 public class RecordFilterEvaluator {
 
+    /// Evaluates the predicate against all rows in a batch, returning a BitSet
+    /// where set bits indicate matching rows. This enables batch-level iteration
+    /// in `AbstractRowReader` instead of per-row `matches()` calls.
+    ///
+    /// @param predicate   the filter predicate to evaluate
+    /// @param batchSize   number of rows in the batch
+    /// @param valueArrays cached primitive arrays per projected column (e.g., `int[]`, `long[]`, `double[]`).
+    /// @param nulls       null bitmaps per projected column
+    /// @param nameCache   column name to projected index mapping
+    /// @return BitSet with bits set for matching rows
+    public static BitSet matchBatch(FilterPredicate predicate, int batchSize,
+            Object[] valueArrays, BitSet[] nulls, StringToIntMap nameCache) {
+        BitSet result = new BitSet(batchSize);
+        for (int i = 0; i < batchSize; i++) {
+            if (matches(predicate, i, valueArrays, nulls, nameCache)) {
+                result.set(i);
+            }
+        }
+        return result;
+    }
+
     /// Returns `true` if the row at `rowIndex` matches the predicate.
     ///
     /// @param predicate   the filter predicate to evaluate
     /// @param rowIndex    the row position within the current batch
     /// @param valueArrays cached primitive arrays per projected column
     /// @param nulls       null bitmaps per projected column (may contain null entries for required columns)
-    /// @param nameCache   column name → projected index mapping
+    /// @param nameCache   column name to projected index mapping
     public static boolean matches(FilterPredicate predicate, int rowIndex,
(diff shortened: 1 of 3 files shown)


core/src/main/java/dev/hardwood/internal/reader/RecordFilterEvaluator.java
```diff
@@ -0,0 +1,270 @@
...
+
+    /// Returns `true` if the row at `rowIndex` matches the predicate.
+    ///
+    /// @param predicate   the filter predicate to evaluate
+    /// @param rowIndex    the row position within the current batch
+    /// @param valueArrays cached primitive arrays per projected column
+    /// @param nulls       null bitmaps per projected column (may contain null entries for required columns)
+    /// @param nameCache   column name → projected index mapping
+    public static boolean matches(FilterPredicate predicate, int rowIndex,
+            Object[] valueArrays, BitSet[] nulls, StringToIntMap nameCache) {
```
gunnarmorling: `Object[] valueArrays` -- is any boxing of values happening there?

gunnarmorling (commented):
This should be fine function-wise. Do you have any idea in regards to the performance overhead? Say I have a file with 10M records without any row group or page level statistics and I apply an eq predicate against each one of them, what would be the impact? I think eventually we'd want to vectorize that filtering, but we can leave that as a follow-up.

Issue #184, opened 2026-03-29
Add logical-type-aware filter predicate factory methods


gunnarmorling:
## Summary

Filter predicate factory methods only accept raw physical types (`int`, `long`, `float`, `double`, `boolean`, `String`, `byte[]`). Users filtering on logically typed columns must manually convert values:

```java
// User has to know DATE is days-since-epoch as INT32
FilterPredicate.gt("date_col", (int) LocalDate.of(2026, 1, 1).toEpochDay())

// User has to know TIMESTAMP(MICROS) is microseconds-since-epoch as INT64
FilterPredicate.gt("created_at", someInstant.toEpochMilli() * 1000)
```

This is error-prone — e.g. passing nanoseconds to a TIMESTAMP(MILLIS) column is physically valid (both are INT64) but semantically wrong.

## Proposal

Add factory method overloads that accept Java types corresponding to Parquet logical types:

- `LocalDate` for DATE (INT32, days since epoch)
- `Instant` for TIMESTAMP (INT64, millis/micros/nanos since epoch depending on unit)
- `LocalTime` for TIME (INT32/INT64 depending on unit)
- `BigDecimal` for DECIMAL (INT32/INT64/FIXED_LEN_BYTE_ARRAY depending on precision)

Validate that the target column's logical type matches the factory method used, throwing `IllegalArgumentException` on mismatch.

## Depends on

- #183 (physical type validation)

PR #188, opened 2026-03-29
#184 Add logical-type-aware filter predicate factory methods

gunnarmorling:


gunnarmorling:
Hey @rionmonster, another improvement related to filtering, in case you want to take a look. This is about accepting the corresponding Java type for filters of columns with a logical type. So you e.g. can filter a `DATE` column by passing in a `LocalDate` rather than an `int` (days since epoch).

Issue #146, opened 2026-03-25
Use Quarkus' native binary testing infra


gunnarmorling:
Currently, the CLI native binary is manually tested (see README), and there are several shell scripts for doing so, too. Instead, can we move to the Quarkus infra for testing native binaries:

- https://quarkus.io/guides/getting-started-testing#quarkus-integration-test
- https://quarkus.io/guides/building-native-image#testing-the-native-executable

This will run tests as part of the native binary, reducing manual efforts and friction.

gunnarmorling:
Woot, woot, awesome! I'm gonna put this issue tentatively into the CR1 milestone then.

gunnarmorling:
Merged; thanks a lot, @brbrown25!

PR #235, opened 2026-04-09
#146 Implementation of QuarkusMainIntegrationTest so we can test the native binary, catching any issues early in the cycle.

brbrown25:
## Checklist


brbrown25:
I tested this with 
```
sdk use java 25.0.2-graalce
./mvnw -Dnative verify -pl cli
```
and it passed 

```
❯ uname -a
Darwin SSC-T9240GX4DF 25.4.0 Darwin Kernel Version 25.4.0: Thu Mar 19 19:30:44 PDT 2026; root:xnu-12377.101.15~1/RELEASE_ARM64_T6000 arm64

❯ sw_vers
ProductName:            macOS
ProductVersion:         26.4
BuildVersion:           25E246

OS: macOS Tahoe 26.4 (25E246) arm64
Host: MacBook Pro (16-inch, 2021)
Kernel: Darwin 25.4.0
CPU: Apple M1 Pro (10) @ 3.23 GHz
GPU: Apple M1 Pro (16) @ 1.30 GHz [Integrated]                                                                                            
```
I'll dig more into the failures but this is a start :)

brbrown25:
Ok I think I've figured out what's up and hope to have an updated commit in the morning.

.github/workflows/maven.yml
```diff
@@ -209,12 +209,16 @@ jobs:
         env:
           MAVEN_OPTS: '--add-modules jdk.incubator.vector'
 
-  # Verify the native CLI build on PRs (Linux only, no release)
```
gunnarmorling, line 212: This was actually Linux only intentionally; these native builds really add up and I'm concerned we might be pushed beyond the resources we're getting for free. So the idea is to build the native binary for one platform here for validation, and then have another, manually triggered, job for building and publishing early-access builds for all supported platforms when we feel it makes sense.

cli/src/test/java/dev/hardwood/cli/command/ConvertCommandIT.java
```diff
@@ -0,0 +1,74 @@
...
+    public String listFile() {
+        return getClass().getResource("/list_basic_test.parquet").getPath();
+    }
+
+    @Override
+    public String nonexistentFile() {
+        return "nonexistent.parquet";
+    }
+
+    @Test
```
gunnarmorling, line 48: There are quite a few additional tests here; is there any reason for having them only in the IT, rather than in the contract classes?

gunnarmorling (commented):
Thanks a lot, @brbrown25! A few comments/questions inline.

.github/workflows/maven.yml
brbrown25: Good call let me switch that back!

cli/src/test/java/dev/hardwood/cli/command/ConvertCommandIT.java
brbrown25: So I added the IT tests so that we could have the existing contract tests are about the JVM side and the IT should match the non IT but have the annotations to actually try the binary. That being said Claude may have added some extra so let me take another look!

brbrown25:
I started addressing the feedback but got pulled into somethings, but will get this pushed up in the next couple days

brbrown25:
got a few more tweaks to make, almost ready

test-bom/pom.xml
```diff
@@ -92,7 +92,7 @@
       <dependency>
         <groupId>org.testcontainers</groupId>
         <artifactId>testcontainers-bom</artifactId>
-        <version>1.20.6</version>
+        <version>1.21.4</version>
```
brbrown25, line 95: Without this I found compatibility issues with getting test containers to work happily in linux

.github/workflows/maven.yml
```diff
@@ -209,12 +211,16 @@ jobs:
...
+  # Verify the native CLI build on PRs
   native-build-check:
     if: github.event_name == 'pull_request'
     needs: build
-    name: 'Verify native CLI build'
-    runs-on: ubuntu-latest
+    name: 'Verify native CLI build on ${{ matrix.os }}'
```
brbrown25, line 218: I made this change so that in the future if we want to support multiple os it's there, but I'm happy to back that out.

brbrown25:
@gunnarmorling I have pulled in the lasted main changes and we successfully have native integration tests testing the binary :) sorry for the delay on this one.

.github/workflows/maven.yml
gunnarmorling: That's alright, fine to keep.

gunnarmorling:
Hey @brbrown25, sorry that this took a bit longer to review. I was heads-down into a large scale refactoring of the decoding pipeline last week, so I just got back to this one now.

I noticed that as per this PR, essentially all CLI tests were only running after the native binary has been built, which would extend the feedback cycle quite a bit. So I've reworked things, leaving the existing (JVM-based) integration tests in place (for faster feedback), only adding two smoke tests--one for a local file, one for an S3 file--for the native binary (for asserting general functionality of the native binary). I've also removed the manual test shell scripts as they shouldn't be needed any longer (_NATIVE_BUILD.md_ still has instructions for some basic manual testing, when needed).

The result is in PR https://github.com/hardwood-hq/hardwood/pull/267, gonna close this one here in favor of it. Thanks again!

brbrown25:
@gunnarmorling no worries! I like that approach. They actually do still run in the regular tests but it builds the jar first so your approach actually makes a lot more sense for a quite a few reasons.

gunnarmorling:
Woot, awesome, thanks a lot, @brbrown25!

PR #241, opened 2026-04-14
#90 Include file name in all exceptions raised during reading

### LINKED ISSUE #90: Include file name in all exceptions raised during reading
## Summary

When Hardwood throws an exception during file reading (decoding errors, schema issues, corrupt data, etc.), the exception message often does not include the name of the file that triggered it. This makes debugging difficult, especially when processing multiple files.

All exceptions surfaced to the caller should consistently include the file name. This should be handled in one centralized place (e.g., wrapping or amending exceptions at the reader entry points) rather than sprinkling file-name context throughout individual decoders and internal components.

## Acceptance Criteria

- Every exception that escapes a Hardwood reader includes the originating file name
- The solution is centralized (single wrapping/amending point), not scattered across internal classes
- Existing exception messages and causes are preserved

brunoborges:
## Summary

Closes #90 — every exception escaping a reader now includes the originating file name in the format `[fileName] original message`.

### ExceptionContext utility
Centralised in `dev.hardwood.internal.ExceptionContext.addFileContext()` — single source of truth for exception wrapping used by `FlatRowReader`, `NestedRowReader`, and `ColumnReader`. Type-preserving logic:

- **Exact-class checks** for `IllegalArgumentException`, `IllegalStateException`, `NullPointerException`, `IndexOutOfBoundsException` — avoids downgrading subclass types
- **`instanceof`** for `UncheckedIOException` (preserves `IOException` cause)
- **CompletionException** — preserves original cause chain (not the CompletionException wrapper itself)
- **Reflection fallback** for other types with `(String, Throwable)` constructor
- **`RuntimeException` wrapper** as last resort
- **Idempotency** — skips wrapping if `[fileName]` prefix is already present

### File name tracking through the v3 pipeline

```
PageSource (tracks currentWorkItem with InputFile)
  → ColumnWorker (fileNameBuffer[] parallel to reorderBuffer, file boundary detection in drain)
    → BatchExchange.Batch / NestedBatch (each batch carries its fileName)
      → FlatRowReader / NestedRowReader / ColumnReader (reads fileName from batch)
```

- **PageSource:** exposes `getCurrentFileName()` from the current work item's `InputFile.name()`
- **ColumnWorker:** the retriever thread stores the file name in a parallel `fileNameBuffer[slot]` before submitting decode tasks. The drain thread detects file boundary changes and flushes the current batch, ensuring each batch maps to exactly one file. Errors from retriever, decode, and drain threads are enriched with file context before `signalError()`
- **FlatColumnWorker / NestedColumnWorker:** `publishCurrentBatch()` stamps `currentBatchFileName` onto the outgoing batch
- **FlatRowReader / NestedRowReader:** `loadNextBatch()` reads `currentFileName` from the first column's batch. All ~29/40 public accessor methods are wrapped with try-catch that delegates to `ExceptionContext`
- **ColumnReader:** `nextBatchInternal()` reads `currentFileName` from the batch. All typed getters, null accessors, and offset methods are wrapped

### Memory ordering

The `fileNameBuffer[]` in `ColumnWorker` is a plain `String[]` (no volatile/atomic). Safety is guaranteed by the existing happens-before chain: retriever writes `fileNameBuffer[slot]` before submitting the decode task, which does a volatile write to `reorderBuffer[slot]` via `AtomicReferenceArray.set()`, and the drain thread does a volatile read via `getAndSet()`. Slot reuse is safe because the retriever throttles when `nextSeq - consumePosition >= MAX_INFLIGHT_PAGES`.

### Coverage

All public accessor methods in `FlatRowReader` (~29 methods) and `NestedRowReader` (~40 methods) are wrapped with try-catch blocks that delegate to `ExceptionContext`. `ColumnReader` wraps `nextBatch()`, all typed value getters, null accessors, and offset methods.

`FilteredRowReader` is a pure delegate — no wrapping needed since the underlying reader already wraps.

## Tests

15 new tests:

**FileNameInExceptionTest** (8 tests):
- Single-file RowReader: type mismatch, missing column
- Multi-file RowReader: correct attribution across files
- Single-file ColumnReader: type mismatch, no-batch state
- Multi-file ColumnReader: correct attribution across batches
- Exception type preservation (IllegalArgumentException)
- Bracketed format verification (`[fileName] ...` regex)

**ExceptionContextTest** (7 tests):
- Type preservation for common exception types
- Idempotency (already-wrapped exceptions returned as-is)
- Null file name passthrough
- Null message handling
- Exotic exception type fallback

Updated `PqRowApiTest` null-column assertions to use `hasMessageContaining()` to accommodate the file name prefix.

All 643 core tests pass.

### DIFF
--- a/core/src/main/java/dev/hardwood/internal/ExceptionContext.java
+++ b/core/src/main/java/dev/hardwood/internal/ExceptionContext.java
@@ -1,0 +1,98 @@
+/*
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Copyright The original authors
+ *
+ *  Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package dev.hardwood.internal;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+
+/// Utility for enriching exception messages with file-name context.
+///
+/// All exception-wrapping for file-name enrichment is centralised here so that
+/// reader classes share the same logic.
+public final class ExceptionContext {
+
+    private ExceptionContext() {
+    }
+
+    /// Amends the exception message with a `[fileName] ` prefix. Preserves the
+    /// original exception type and cause chain. Returns the original exception
+    /// unchanged when the file name is unavailable or the prefix is already present.
+    ///
+    /// @param fileName the originating file name, may be `null`
+    /// @param e        the exception to enrich
+    /// @return the enriched (or original) exception — never `null`
+    public static RuntimeException addFileContext(String fileName, RuntimeException e) {
+        if (fileName == null || fileName.isEmpty()) {
+            return e;
+        }
+        String prefix = "[" + fileName + "] ";
+        String originalMessage = e.getMessage();
+        if (hasFilePrefix(originalMessage)) {
+            return e;
+        }
+        // If the cause already carries file context (e.g. assembly-thread error
+        // propagated through CompletionException), don't add a second layer.
+        Throwable cause = e.getCause();
+        if (cause != null && hasFilePrefix(cause.getMessage())) {
+            return e;
+        }
+        String newMessage = prefix + (originalMessage != null ? originalMessage : e.getClass().getSimpleName());
+
+        if (e instanceof UncheckedIOException uio) {
+            // UncheckedIOException requires an IOException cause, so we can't chain
+            // the original UncheckedIOException as the cause. Preserve it as suppressed.
+            IOException ioCause = uio.getCause();
+            if (ioCause == null) {
+                ioCause = new IOException(originalMessage);
+            }
+            UncheckedIOException wrapped = new UncheckedIOException(newMessage, ioCause);
+            wrapped.addSuppressed(e);
+            return wrapped;
+        }
+        if (e.getClass() == IllegalArgumentException.class) {
+            return new IllegalArgumentException(newMessage, e);
+        }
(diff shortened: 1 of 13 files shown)


brunoborges:
@gunnarmorling and @rionmonster 

I believe this is good for review now!

gunnarmorling:
Hey @brunoborges, thanks a lot for looking into this! This will be very useful. Quick heads-up that I'm currently working on a large scale refactoring of the decoding pipeline and remote IO path (see https://github.com/hardwood-hq/hardwood/pull/245), which is why I'm holding off with merging this one right now. I'm expecting that a fair bit of rework will be needed once that refactoring has landed, but Claude/Co-Pilot should probably be able to get it done quickly. I'll get back to you once I'm done (it's taking a bit longer than I hoped, but I'm starting to see the light at the end of the tunnel).

gunnarmorling:
Hey @brunoborges, so the pipeline rework has been integrated into main. Could you rebase this PR? Note quite a few classes have gone, so having the agent start from scratch might be the most promising approach.

brunoborges:
@gunnarmorling PR updated !

gunnarmorling:
Thanks a lot, @brunoborges! I've reworked this a bit (prefixing exceptions with file names even more generally, changing tests, etc.), the outcome is PR https://github.com/hardwood-hq/hardwood/pull/317/, retaining you as the original author. Gonna close the one here in favor of the new one. Thanks again, very nice improvement!

Issue #251, opened 2026-04-16
Document how inequality filters interact with nulls


mike-d-d:
I looked for some documentation of how FilterPredicates `lt`, `gt`, `ltEq`, and `gtEq` treat nulls and couldn't find anything — did I miss it?

The [Apache docs](https://javadoc.io/static/org.apache.parquet/parquet-column/1.10.0/index.html?org/apache/parquet/filter2/predicate/FilterApi.html) say that its versions always exclude nulls, which seems reasonable except that then `not(gt(x, 5))` is no longer equivalent to `ltEq(x, 5)` (since the former would include columns where x is null, and the latter wouldn't).

What's the expected behavior in this case?

gunnarmorling:
Great catch, thanks for reporting!

I've filed PR #253 to clarify this. In a nutshell, Hardwood applies SQL three-valued logic uniformly: any comparison against a null column value is UNKNOWN and the row is dropped. notEq included, and not(p) preserves it — so `not(gt(x, v))` = `ltEq(x, v)` on every row, nulls included. To keep nulls, ask explicitly: `or(gt("age", 30), isNull("age"))`.

There's one deliberate divergence from parquet-java though: its notEq keeps null rows; Hardwood drops them for the sake of SQL compatibility. A compat mode for opting into parquet-java's behavior is tracked via #252.

mike-d-d:
Thanks, that's all very clear now.

PR #253, opened 2026-04-17
#251 Clarifying null semantics in filters

gunnarmorling:
First and foremost, a docs update. The test just is there to demonstrate the actual behavior in an executable fashion and detect and potential regressions in the future.

### DIFF
--- a/parquet-testing-runner/src/test/java/dev/hardwood/testing/NullFilterSemanticsComparisonTest.java
+++ b/parquet-testing-runner/src/test/java/dev/hardwood/testing/NullFilterSemanticsComparisonTest.java
@@ -1,0 +1,136 @@
+/*
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Copyright The original authors
+ *
+ *  Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package dev.hardwood.testing;
+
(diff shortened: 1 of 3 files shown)


parquet-testing-runner/src/test/java/dev/hardwood/testing/NullFilterSemanticsComparisonTest.java
```diff
@@ -0,0 +1,136 @@
...
+            SimpleGroupFactory factory = new SimpleGroupFactory(schema);
+            writer.write(factory.newGroup().append("age", 25));
+            writer.write(factory.newGroup().append("age", 35));
+            // Third row: age left unset → null.
+            writer.write(factory.newGroup());
+        }
+    }
+
+    @Test
+    void caseA_notGt_bothEnginesDropNulls() throws IOException {
```
rionmonster: Minor nit on the naming of these as the case-based prefix (e.g., "caseA" and "caseB") feels a little funky as I think the later part of the test names cover it. Not a hill worth dying over by any means.

rionmonster (approved):
I think overall the clarification of behavior is useful and important — looks good to me overall! Left a minor nit on the test naming but otherwise approving. 👍

gunnarmorling:
Thanks a lot, @rionmonster. Good call on the naming, updated.

PR #327, opened 2026-04-25
#128 Design doc for modular encryption support

mghildiy:
## Checklist

- [X] The commit message is prefixed with the issue key ("# 128 Design doc for parquet modular encryption support")

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -333,6 +333,24 @@ A from-scratch implementation of Apache Parquet reader/writer in Java with no de
 - [x] Inline `DataPageHeader.statistics` fallback when Column Index is absent (`InlinePageStatsScanner`)
 - [ ] Bloom filter-based filtering
 
+### 9.5 Modular Encryption (AES-GCM / AES-GCM-CTR)
+- [x] `ParquetModuleType` — module type constants (0x00–0x09)
+- [x] `ParquetCryptoHelper` — raw AES-GCM and AES-CTR crypto primitives
+- [x] `ColumnDecryptor` — per-column-chunk decryption context
+- [x] `DecryptionKeyProvider` — public interface for key resolution
+- [x] `AadPrefixProvider` — public interface for AAD prefix supply
+- [x] PARE footer decryption (`FileCryptoMetaData` reading + footer AES-GCM decrypt)
+- [x] PAR1 plaintext footer with encrypted columns
+- [x] Per-column key resolution (footer key vs column key)
+- [x] Dictionary page decryption (`SequentialFetchPlan`, `IndexedFetchPlan`)
+- [x] Data page header + data decryption (`PageDecoder`)
+- [x] Offset index decryption (`ColumnDecryptor.decryptOffsetIndex`)
+- [x] AAD prefix stored in file (`supplyAadPrefix=false`)
+- [x] AAD prefix supplied by caller (`supplyAadPrefix=true`)
+- [ ] Column index decryption
+- [ ] Footer signing verification (PAR1 plaintext footer with signature)
+- [ ] Write path encryption
+
 ---
 
 ## Phase 10: Public API Design
@@ -460,6 +478,15 @@ A from-scratch implementation of Apache Parquet reader/writer in Java with no de
 - [ ] Input validation
 - [ ] Parallel writing support
 - [ ] **Validate**: Full compatibility with parquet-java and PyArrow
+
+### Milestone 8: Modular Encryption (under progress)
+- [x] PARE footer decryption
+- [x] PAR1 plaintext footer with encrypted columns
+- [x] Dictionary and data page decryption
+- [x] Offset index decryption
+- [ ] Column index decryption
+- [ ] Footer signing verification
+- [ ] **Validate**: Read all parquet-testing aes256 encrypted files
 
 ### Interactive CLI (`hardwood dive`)
 TUI for exploring Parquet file structure. See `_designs/INTERACTIVE_DIVE_TUI.md`.
--- a/_designs/MODULAR_ENCRYPTION_SUPPORT.md
+++ b/_designs/MODULAR_ENCRYPTION_SUPPORT.md
@@ -1,0 +1,324 @@
+# Design: Parquet Modular Encryption (read path)
+
+**Status: Under implementation**
+
+**Issues:** #128 (Support Parquet Modular Encryption)
+
+## Scope
+
+This document only takes into account decryption support since Hardwood currently have a read-only implementation. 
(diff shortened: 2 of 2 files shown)


gunnarmorling:
Thanks a lot, @mghildiy! I've asked Claude to do a review pass, sharing it below. Not everything may apply verbatim, but can you check it out and address what makes sense? Thanks!

---

# Review of PR #327 — Design doc for modular encryption support

## Things that should change before merging

**1. Filename convention.** Other design docs use `ALL_CAPS_SNAKE_CASE.md` (e.g. `OFFSET_INDEX_SUPPORT.md`, `COLUMN_INDEX_PUSHDOWN.md`). Rename to `MODULAR_ENCRYPTION_SUPPORT.md`.

**2. Missing standard front matter.** Other recent design docs (`VARIANT_LOGICAL_TYPE.md`, `AVRO_GENERICRECORD_SUPPORT.md`) lead with **Status:** and **Issues:** lines. This one starts with `# Design document:` (with a trailing colon and no project name). Suggest:

```
# Design: Parquet Modular Encryption (read path)

**Status:** Proposed
**Issues:** #128
```

**3. Scope is unclear.** Issue #128 covers both reading and writing. The doc only covers reading but never says so explicitly. Add a one-line scope statement near the top — e.g. *"This document covers the read (decrypt) path only; the write path is deferred to a follow-up."*

**4. `ROADMAP.md` has no encryption entry.** `CLAUDE.md` says *"Update the status in the roadmap after implementing a feature."* Implementation isn't in this PR, but it's worth either adding a "Modular Encryption" section now (so the design and roadmap line up) or noting in the doc that one will be added when implementation lands.

**5. Public-API surface needs more thought.** The proposed builder is presented as a fait accompli, but the doc only shows the happy path:

```java
ParquetFileReader.builder(InputFile.of(path))
    .keyProvider(...)
    .aadPrefixProvider(...)
    .open();
```

Open questions worth resolving in the design, not in code review later:
- The existing static `open(InputFile)` and `open(InputFile, HardwoodContext)` factories already cover the basic + context cases. Is `builder()` an addition or a replacement, and how does it compose with `HardwoodContext`?
- `DecryptionKeyProvider` and `AadPrefixProvider` are mentioned but never defined. Sketch their signatures (return type, exceptions, what `keyMetadata` looks like).
- What happens if a file is encrypted but no `keyProvider` is supplied — `IOException`, dedicated `ParquetCryptoException`, something else? "Fail early" per `CLAUDE.md` argues for a dedicated typed exception.
- Where do the new types live? Per the public-API rules in `CLAUDE.md`, only the user-facing pieces (`DecryptionKeyProvider`, `AadPrefixProvider`, the builder, an exception type) belong in a public package; `ColumnDecryptor`, `ParquetCryptoHelper`, `ParquetModuleType` should be in `internal`. State this explicitly.

**6. `ColumnDecryptor.forColumnChunk()` returning `null` for unencrypted files.** This is the kind of API choice that's worth flagging in the design rather than discovering during review. `Optional<ColumnDecryptor>` or a `NoopColumnDecryptor` is probably more in line with the codebase — check what the rest of the reader does.

**7. AAD prefix handling is under-specified.** Section "Encryption Flows Supported" lists *"AAD prefix stored in file"* (flow 6) and *"AAD prefix supplied by caller"* (flow 7) but the design doesn't say how the reader decides which to use, or what happens if the file has `supply_aad_prefix=true` and the caller provides none. This is the spec's main interop pitfall — please cover it.

**8. Failure modes are absent.** A read-side crypto design should enumerate: GCM tag mismatch, unknown algorithm, missing key for `keyMetadata`, truncated module (length prefix overruns buffer), `PARE` magic with no `FileCryptoMetaData`. One short "Error handling" section will do.

**9. Per `CLAUDE.md` doc style.** JavaDoc rules require `///` Markdown — fine, this is a design doc not JavaDoc, so not applicable. But the snippets in the doc (`{@link}`-free, plain Markdown) are consistent with that, good.

## Smaller nits

- L24: *"I there is any mismatch"* → *"If there is any mismatch"*.
- L96: trailing `---` directly after `**Thread safety:** …` with no blank line; minor rendering glitch.
- Table on L90 has *"GCM or CTR depending on `useCtr`"* but the field is described above as `boolean useCtr` — consider naming it after the algorithm enum (`AES_GCM_V1` / `AES_GCM_CTR_V1`) to keep the design extensible if the spec ever adds a third mode.
- L196–198: closing line *"-Diagrams generated with inputs from Sonnet 4.6."* — leading hyphen looks like a typo, and `CLAUDE.md` (*"Write as if the reader has no context on the conversation that produced the document"*) argues for dropping the attribution entirely. Same for any other "generated with" markers.
- File ends without trailing newline (`\ No newline at end of file`).
- The roadmap reference and most other docs avoid the colon in headings — `# Design document:` with a trailing colon reads oddly; rename to a real title (see point 2).

## Risks / things that aren't covered

- **No mention of `aadFileUnique` provenance.** Where is it read from in the footer? For `PARE` files it lives in `FileCryptoMetaData`; for plaintext-footer encrypted files it's in `FileMetaData.encryption_algorithm`. Worth saying so.
- **Plaintext footer signing (`footerSigningKeyMetadata`).** The diagram mentions it but the doc never says whether signature verification is in scope. The Parquet spec requires verifying the GCM tag over the footer bytes when a plaintext footer is signed; if you don't verify, you've defeated the point of the signed-footer mode. Either implement or explicitly defer with a note.
- **Encrypted column/offset/bloom indexes.** Module types `COLUMN_INDEX`, `OFFSET_INDEX`, `BLOOM_FILTER_HEADER`, `BLOOM_FILTER_BITSET` are listed in `ParquetModuleType` but the integration section only walks through page header + page data. Hardwood has column-index / offset-index / bloom support — call out whether their decryption is in scope for the first cut.
- **Performance.** No discussion of `Cipher` instance reuse vs. per-call allocation. Given Hardwood's strong "avoid object access and boxing" stance in `CLAUDE.md`, a sentence on this is warranted (even if just "we'll measure and tune in the implementation PR").

## Verdict

The doc is a solid first draft and covers the core mechanics correctly. Before merging, I'd want to see: scope statement, sketches of the new public types (`DecryptionKeyProvider`, `AadPrefixProvider`, exception class, internal-vs-public placement), explicit handling of the AAD-prefix and signed-footer cases, an error-modes section, and the small textual cleanups above. The crypto-class shapes themselves look reasonable and don't need to change.

gunnarmorling:
Hey @mghildiy, thank you so much for driving this, and sorry for not getting back on this earlier. My current thinking is to keep encryption/decryption for 1.1 as we're on the homestretch towards 1.0, and in particular this one shouldn't be rushed. It shouldn't be very long until 1.1 once 1.0 is out, while giving us the time to wrap this up.

PR #346, opened 2026-04-26
#140 Supporting Map & List Logical Types

### LINKED ISSUE #140: Support MAP and LIST variants in LogicalType union
## Problem

`LogicalTypeReader` handles most `LogicalType` union variants but silently skips MAP (field 2) and LIST (field 3). Currently, map and list types are recognized via the legacy `ConvertedType` field on `SchemaElement`. If a future writer emits `LogicalType` without the legacy `converted_type`, maps and lists would silently lose their type annotation.

## Proposed Fix

Parse fields 2 (MAP) and 3 (LIST) in `LogicalTypeReader`. Both are empty structs (no inner fields), analogous to the existing STRING and ENUM handling — just read the struct and return the corresponding `LogicalType` instance.

## Context

Found during the audit of skipped Thrift fields for #135. All other skipped fields (`file_path`, `file_offset`, `index_page_offset`, `index_page_header`, `INTERVAL`) are either deprecated, redundant, or have no practical use case and are intentionally left unimplemented.

iifawzi:
Hello, this PR should resolve #140, supporting map & list logical types

## Checklist


iifawzi:
On a side note, each time the data generator runs, it produces diffs on already committed parquet files, and it's quite annoying. it happens because pyarrow embeds its own version string in the file footer's created_by field, so regenerating with a different Pyarrow install changes every parquet file

Claude suggested two options: pin pyarrow in requirements.txt to the exact version used to generate the committed files, or post-process each generated file to overwrite created_by with a fixed constant like "hardwood-test-datagen", making the output always deterministic regardless of which version used

That would make the devx a bit better, if you agree with any of the approahces @gunnarmorling i can commit it in this PR or in a separate one.

core/src/main/java/dev/hardwood/internal/thrift/LogicalTypeReader.java
```diff
@@ -67,7 +75,7 @@ private static LogicalType readInternal(ThriftCompactReader reader) throws IOExc
                         yield new LogicalType.UuidType();
                     }
                     case 16 -> readVariantType(reader);
-                    // Skip unsupported types (MAP, LIST, etc.)
+                    // Skip unsupported types (NullType, Float16Type, etc..)
```
iifawzi, line 82: I'm not sure if it's needed or not? null and float16, but i assume they're the only two missing

iifawzi: Resolved as i saw other tasks already for float16, perhaps null is just missing, and there's also other tasks for gemoetery, and geography

gunnarmorling:
Ah, just seeing your comment on the datagen issue now. Yes, totally agreed, as per our convo around #349.

iifawzi:
Will be rebased after #347

gunnarmorling:
@iifawzi, could you rebase this one to current main now? Thx!

iifawzi:
sure, rebased and squashed @gunnarmorling

gunnarmorling:
@iifawzi, it looks like those new Parquet test files have a customized Created-By header, but I can't find any logic in the generator for doing so, which causes re-generation to produce a diff for me:


```  
  cmp output:
    orig = 747 bytes, new = 758 bytes
    first differing offset = 0x2c4 (708)

  --- ORIGINAL fixture (PR commit daebcc3) ---
  000002b4: 41 41 41 53 41 41 41 41 41 41 41 41 41 41 00 18  AAASAAAAAAAAAA..
  000002c4: 15 68 61 72 64 77 6f 6f 64 2d 74 65 73 74 2d 64  .hardwood-test-d
  000002d4: 61 74 61 67 65 6e 19 2c 1c 00 00 1c 00 00 00 64  atagen.,.......d
  000002e4: 02 00 00 50 41 52 31                             ...PAR1

  +++ REGENERATED fixture (this checkout) +++
  000002b4: 41 41 41 53 41 41 41 41 41 41 41 41 41 41 00 18  AAASAAAAAAAAAA..
  000002c4: 20 70 61 72 71 75 65 74 2d 63 70 70 2d 61 72 72   parquet-cpp-arr
  000002d4: 6f 77 20 76 65 72 73 69 6f 6e 20 32 34 2e 30 2e  ow version 24.0.
  000002e4: 30 19 2c 1c 00 00 1c 00 00 00 6f 02 00 00 50 41  0.,.......o...PA
```

Any idea why that could be the case? I think we should just keep the orginal `parquet-cpp-arrow version...` value.

iifawzi:
> @iifawzi, it looks like those new Parquet test files have a customized Created-By header, but I can't find any logic in the generator for doing so, which causes re-generation to produce a diff for me:

Ops, those must have been generated with an older version while i was testing hard coding it. Will regenerate it
good to know about cmp as well :p

gunnarmorling:
Nice, looking good now. Squashed and applied. Thanks a lot, @iifawzi! In case you're up for more, I've tagged some more issues with "good first issue" and "help wanted" earlier today. Also open for any other suggestions or ideas, of course.

iifawzi:
> Nice, looking good now. Squashed and applied. Thanks a lot, @iifawzi! In case you're up for more, I've tagged some more issues with "good first issue" and "help wanted" earlier today. Also open for any other suggestions or ideas, of course.

Thank you, I really appreciate the support and the effort you're putting making contributions easier, i have strong interest in parsers in general, and recently i've been diving into parquet as well, so i'm excited spending some of my free time on that. 

Next, I plan to look into the side effects introduced by the interval support (#350 and #351), and then move on to reviewing #3 as it feels cool to have

iifawzi:
@gunnarmorling Final thought, to close logical types, I assume the only one missing would be UNKNOWN/NULL? do we have issue to track it? https://parquet.apache.org/docs/file-format/types/logicaltypes/

gunnarmorling:
Good question, I’m not sure tbh. Probably not. Worth double checking,
logging, and fixing, I suppose.

PR #353, opened 2026-04-26
#349 Pinning pyarrow to 24.0.0 and regenerate files with it

### LINKED ISSUE #349: simple-datagen.py: only overwrite fixtures whose bytes actually change
## Problem

Running `simple-datagen.py` currently rewrites every Parquet test fixture under `core/src/test/resources/` (and a few others) on every invocation, even when the regenerated bytes are functionally identical. This makes the script disruptive to use:

- A contributor who only wants to add or refresh **one** fixture ends up with ~30+ fixtures showing as modified in `git status`.
- Differences are typically incidental — e.g. the embedded `created_by` writer-version string changes (`parquet-cpp-arrow 22.0.0` → `23.0.1`) — which then breaks unrelated tests that assert against that exact string (e.g. `InfoCommandContract.displaysFileInfo`).
- Code reviewers have to manually distinguish meaningful fixture changes from drive-by churn, and contributors have to remember to revert the unrelated ones.

We hit this concretely while adding the INTERVAL fixture for #122: regenerating produced a 34-fixture diff with one CLI test failure, none of it related to the actual change.

## Proposal

Make `simple-datagen.py` write each output file only when its bytes actually differ from the on-disk version. Sketch:

1. Write the candidate bytes to a temp file (or buffer) instead of the final path.
2. Compare against the existing file (if any). If identical, drop the temp file and leave the original in place (preserving its `mtime`).
3. Otherwise, atomically replace the target.

For fixtures that go through a post-processing step (the `parquet_*_annotation.py` helpers), the comparison must run **after** annotation, so transient pre-annotation bytes don't trigger spurious rewrites.

A nice-to-have: print a summary at the end listing only the files that were rewritten, so contributors can sanity-check what changed.

## Out of scope

- Stabilizing the `created_by` string (would require pinning PyArrow or post-processing the footer; orthogonal to this issue).
- Refactoring the three `parquet_*_annotation.py` helpers to share their IDL (separate cleanup).

iifawzi:
Hello, this PR resolves #349, pinning pyarrow to 24.0.0, and regenerating the files so they all match the version
future contributors wouldn't suffer now, and existing contributors would need to update their local dependency to avoid committing new tests with older versions
```
source .docker-venv/bin/activate && python -m pip install pyarrow==24.0.0
```
## Checklist


iifawzi:
In theory, this should be enough, testing locally now the files aren't getting modified each run @gunnarmorling, we can pin to 22 if you prefer, 24 is the latest stable

requirements.txt
```diff
@@ -7,6 +7,6 @@
 #
 
 numpy>=2.4.1
-pyarrow>=22.0.0
+pyarrow>=24.0.0
```
gunnarmorling, line 10: Hum, hum, this would still install a newer version, if available. We'd have to pin one specific version.

gunnarmorling (commented):
One comment inline; also I suppose the script will still regenerate everything unconditionally (updating file change timestamps)? If so, we'd still need the the temp dir / buffering solution described in the issue.

requirements.txt
iifawzi: Changed, I believe you had the tab opened in an old diff. 

Related to your comment on timestamps, do you think it's worth it the buffer solution? even if it won't be identified diff? (It's not identified diff by git, so maybe timestamps aren't getting changed?)

iifawzi: Testing again, I confirm no diff detected. I'd suggest avoiding the buffer solution for simplicity given timestamp changes wouldn't be detected anyway 

edit: timestamps apparently change for sure, but git detect only content, so that's why it's not detected

gunnarmorling: Ok, nice. I thought git diff would show the files as tainted also if there's solely a timestamp change; but if that's not the case, we're good. Will merge this one tomorrow.

iifawzi: Yeah seems like so. Perfect, thank you!

gunnarmorling:
All looking good; picked up some more lose ends in https://github.com/hardwood-hq/hardwood/commit/23015cf632beb541e0ef6947b75384f5758bbb09 (some files were actually missing from the data generator) and merged. Thanks a lot, @iifawzi!

PR #354, opened 2026-04-26
#75 Implement multi-release mapping to bypass 2GB chunk limit

### LINKED ISSUE #75: MappedByteBuffer 2GB limit on column chunk size
`MappedByteBuffer.slice(int, int)` takes `int` parameters, but file offsets are `long`. When a single mapped region exceeds ~2GB, the `(int)` cast on the offset calculation silently overflows, producing incorrect slices.

This affects `PageScanner`, `SingleFileRowReader`, `FileManager`, and other code that slices into a file-wide mapping. Typical files stay well under the limit, but large single-row-group files with many columns or wide schemas can exceed it.

Possible approaches:
- Multiple mappings with region splitting
- `MemorySegment` (Foreign Memory API) which supports `long` offsets

bmscomp:
This issue has been resolved by implementing a dual-path architecture at the MappedInputFile level:

1. Java 22+ Path (Primary): Utilizes the new FFM java.lang.foreign.FileChannel.map and MemorySegment APIs, which natively support long offsets and lengths. This completely bypasses the int bounds of legacy byte buffers and enables zero-allocation slicing.

2. Java 21 Path (Fallback): Implements a multi-segment MappedByteBuffer[] array. The file is split into chunks of at most Integer.MAX_VALUE, and readRange(long, int) dynamically routes reads to the correct segment using integer division, safely assembling spanning reads via heap allocation when necessary.

Additionally, ByteBufferInputFile.readRange has been hardened to use Math.toIntExact, replacing the silent (int) cast overflow with a descriptive ArithmeticException on illegal accesses.

A detailed architecture document (_designs/large-file-mmap.md) has been added to outline the memory safety and lifecycle guarantees (e.g., Arena.ofAuto() usage) established by this fix.

Fixes #75
## Checklist

- [x] The commit message is prefixed with the issue key ("#75")


gunnarmorling:
Wow, thanks a l lot, @bmscomp! Will look into this ASAP. We've done some fixes to the data generator, so could you rebase to current main and force-push to this PR; the changes to the (unrelated) test Parquet files should be gone (make sure to update your Python venv as per latest requirements.txt).

bmscomp:
Thanks so much @gunnarmorling  for the review, I will manage to do necessary to resolve the review comments

bmscomp:
@gunnarmorling  Branch rebased on main and  updated Python venv as per latest requirements.txt

gunnarmorling:
_(With help from Claude)_

Thanks for the rebase! Before diving into line-level review, could we split this into separate PRs (and issues)? Right now it bundles four largely independent changes:

1. #75 — 2 GB mmap fix. The Java 21 multi-segment fallback + Java 22 MemorySegment override in MappedInputFile, plus the LargeFileMappedInputFileIT test and the doc/release-notes updates. This is the change the issue asks for.
2. FFM zero-copy decode pipeline. LibdeflateDecompressor.decompressToSegment, NativePlainDecoder, NativeBssDecoder, the java22 PageDecoder override, the nativeFastPath JFR field, and ZeroCopyPipelineIT. This is a substantial perf change that deserves its own issue and review on its own merits.
3. ColumnWorker.close() interrupt-flag fix. Defers re-asserting the interrupt until after allOf().join() drains in-flight
decodes.
4. ColumnWorker.runRetriever() inFlightDecodes race fix. Pre-registers a ticket before runAsync so close() can't observe an empty set while a task is still executing.

(3) and (4) are real bugs, so ideally, we'd get them out of the way first. Then we'd have two follow-up PRs, one for (1) and one for (2).

bmscomp:
@gunnarmorling  I will manage to do that asap

gunnarmorling:
Awesome, thank you so much!

gunnarmorling:
Hey @bmscomp, I've opened PR #502 as we haven't heard back. This provides a a more focused solution which doesn't require the FFM API and works for local files. I've also logged a couple of follow-ups:

* #501: Lifting the 2 GB limitation for remote (S3) files too; we'd probably want to use the FFM API for that (to model the local fetch cache as one consecutive file)
* #503: Make readRange return read-only buffers as proposed here
* #504: Exploring the FFM zero-copy native decode pipeline proposed here, if benchmarks show an advantage

I'm gonna close this PR here for the time being as it's superseded by #502. Contributions towards that (or any other issue) more than welcome. Thanks a lot for your efforts!

PR #368, opened 2026-04-27
#350 CLI Support of Interval logical type

### LINKED ISSUE #350: CLI: render INTERVAL columns as a human-readable duration instead of raw bytes
## Problem

After #122 added typed `getInterval` support to the core reader, the CLI value formatters were not updated. Files containing `INTERVAL` columns still surface in `hardwood dive`, `hardwood cat`, etc. as raw 12-byte hex strings (the `formatRawBytes` fallback in `RowValueFormatter` / `IndexValueFormatter`), which is unreadable.

## Where

- `cli/src/main/java/dev/hardwood/cli/internal/RowValueFormatter.java` — both the single-line `format` path (~L80–L100) and the expanded path (~L342–L349) handle every other logical type (`UuidType`, `JsonType`, `BsonType`, `IntType`, …) but not `IntervalType`.
- `cli/src/main/java/dev/hardwood/cli/internal/IndexValueFormatter.java` — same gap (~L132, L167).
- `cli/src/main/java/dev/hardwood/cli/internal/table/RowTable.java` — same gap (~L79, L150).

## Proposal

When `lt instanceof LogicalType.IntervalType`, call `reader.getInterval(fieldIndex)` (or the equivalent path for stats) and format the resulting `PqInterval` to a single readable string. Two reasonable rendering options:

- ISO-8601-ish: `P1M15DT3600S` for `(months=1, days=15, milliseconds=3_600_000)`.
- Plain components: `1mo 15d 3600000ms`, omitting zero components.

I lean toward the plain-components form — INTERVAL months/days are not the same as ISO `P` months/days (independent components, no normalization), and the plain form makes that clearer at a glance.

Don't forget the unsigned-int caveat from `PqInterval`'s Javadoc: when rendering, use `Integer.toUnsignedLong` so values ≥ 2³¹ don't display as negative.

## Tests

Extend `IntervalLogicalTypeTest` (or a sibling CLI test) to assert the rendered string for a row of the existing `interval_logical_type_test.parquet` fixture — both via `format` (single-line) and `formatExpanded`.

## Out of scope

- Localized / user-configurable formatting.
- Rendering INTERVAL inside Variant payloads (Variant doesn't carry INTERVAL).

iifawzi:
Hello, this PR resolves #350 to render Interval columns correctly 

## Checklist

### DIFF
--- a/cli/src/main/java/dev/hardwood/cli/internal/IndexValueFormatter.java
+++ b/cli/src/main/java/dev/hardwood/cli/internal/IndexValueFormatter.java
@@ -10,6 +10,7 @@ package dev.hardwood.cli.internal;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
 import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.time.LocalDate;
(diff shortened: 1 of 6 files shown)


iifawzi:
To be rebased after #372 lands

gunnarmorling:
Rebased, added a commit for some clean-up, and merged. Thanks a lot, @iifawzi!

Issue #277, opened 2026-04-21
Per-column in-page row skipping in PageDecoder


gunnarmorling:
Two problems share a root cause:

1. **Silent data corruption in filter page-pushdown when column page boundaries disagree within a row group.** Real-world parquet-mr files routinely produce this — e.g. `performance-testing/test-data-setup/target/tlc-trip-record-data/yellow_tripdata_2025-12.parquet`, RG 0:

   ```
   VendorID               (1 data page):   [0, 1048576)
   tpep_pickup_datetime   (7 data pages):  [0, 263168)  [263168, 394240)
                                           [394240, 525312)  [525312, 656384)
                                           [656384, 787456)  [787456, 918528)
                                           [918528, 1048576)
   ```

   parquet-mr picks page size by uncompressed byte count per column, so a narrow INT32 gets one giant page while a wider TIMESTAMP gets many. For a filter like `tpep_pickup_datetime >= T` producing `matchingRows = [500 000, 800 000)`:

   - `tpep_pickup_datetime`: keeps pages `[394240, 918528)` → 524 288 rows, first emitted at row 394 240.
   - `VendorID`: keeps its only page in full → 1 048 576 rows, first emitted at row 0.

   `FlatRowReader.loadNextBatch` pairs `VendorID[0]` with `tpep_pickup_datetime[394240]` for output row 0; `FilteredRowReader` re-applies the predicate on the datetime (value passes) and emits the row with the wrong `VendorID`. Silent corruption.

   Existing predicate-pushdown tests don't catch this because every fixture (`filter_pushdown_int.parquet`, `column_index_pushdown.parquet`, …) happens to have matched page boundaries across columns.

2. **Tail reads decode and discard the leading rows of the first kept row group.** #214 drops earlier row groups entirely, but within the straddling row group every page is fetched and every row decoded; the `ParquetFileReader.createTailRowReader` caller then advances past the prefix at the `RowReader` level. A prior attempt pushed this to the fetch plan via `OffsetIndex`, but it required a shared page boundary across all projected columns at the tail cut point (same alignment story as #1), with a fallback when it didn't hold.

## Proposed fix

Push per-row skipping down to `PageDecoder`. Each column independently:

- In `computeFetchPlans`, drops pages whose rows end before its column-specific `matchingRows` start (or, for tail mode, before `skip`). No cross-column alignment check — each column uses its own `OffsetIndex`.
[...]

PR #369, opened 2026-04-27
#277 Address Data Corruption During Pushdown for Unaligned Page Boundaries

rionmonster:
## Description

This pull request addresses #277, which tracked a silent data corruption bug in filter pushdown when sibling columns within a row group have differently-sized pages. When a writer flushes pages by per-column uncompressed byte count, the per-column page filter ends up keeping a different first kept page on each column under any non-trivial filter and the row reader, which pairs columns by stream position, emits values from misaligned global rows.

This proposed fix attaches a new per-page row mask `PageRowMask` to each kept page and applies it during batch assembly so every projected column emits values for the same set of global rows. The same mechanism also fixes the related misalignment under OR predicates with non-contiguous matching ranges. Additionally it covers a performance-related tail-read optimization case. 

## Key Changes
- **Introduced `PageRowMask` Class**  which functions as a sorted, non-overlapping list of offsets (e.g. start-end) relative to a page's first row. Additionally a new sentinel `PageRowMask.ALL` is shared across the unfiltered path so the existing fast path stays allocation-free.
- **Added a RowRanges.maskForPage(pageFirstRow, pageLastRow) Function** responsible for producing the mask for a page given the matching row ranges. 
- **Workflow Changes**
  - **PageInfo carries the mask** and is defaulted to `PageRowMask.ALL` so existing constructors are unaffected. The mask travels with the page from the fetch plan to the assembler.
  - **ColumnWorker reorder buffer carries** via a small `DecodedPage` record so the drain receives both the decoded values and the per-page row selection in a single atomic read.
  - **Column workers (e.g. `FlatColumnWorker` and `NestedColumnWorker`) apply the mask during assembly**  applying appropriate short-circuits to the existing walk. Otherwise flat assembly iterates intervals and copies kept subranges; nested assembly tracks the page-relative top-level record index and skips records whose index falls outside the current interval.
- **Introduced `createRowReader(...)` tail-mode fast path** used when every projected column in every subset row group has an `OffsetIndex` (the reader synthesizes a [skip, numRows) matching range on the first row group and lets the per-page mask machinery drop the leading pages and trim the straddling page).

## Testing and Verification
Several specific tests to verify the underlying corruption issue and the related fix were introduced as well as adjustments to other classes within the affected files, namely:
- Added `MisalignedPageBoundariesTest.java` to verify behavior against a new `misaligned_pages.parquet` fixture whose narrow and wide columns have coprime per-column page periods so no narrow boundary coincides with a wide boundary. 
- Updated `RowRangesTest` with several new unit tests surrounding `maskForPage()` behavior covering ALL sentinel, no-overlap → null, page-inside-range, range-starts/ends-inside-page, range-straddles-both-ends, multi-interval, drop-out-of-range, exclusive boundaries.
- Updated `simple-datagen.py` with a new fixture-generation block responsible for generating misaligned parquet content.
- Added a new `TailReadBenchmarkTest` performance test, primarily for posterity, to track tail-based read performance.

### Performance and Regressions
Classic throughput-based performance tests showed no regressions from the `main` branch as demonstrated below:
```
PERFORMANCE TEST RESULTS

Environment:
  CPU cores:       14
  Java version:    25.0.2

Data:
  Files processed: 70
  Total rows:      480,968,045
  Total size:      6,568.1 MB
  Runs per contender: 3

Performance (all runs):
  Hardwood (row reader indexed) [1]         1.53     315,388,882         22,527,777       4306.9
  Hardwood (row reader indexed) [2]         1.52     317,470,657         22,676,475       4335.4
  Hardwood (row reader indexed) [3]         1.52     317,261,243         22,661,517       4332.5
  Hardwood (row reader indexed) [AVG]         1.52     316,843,244         22,631,660       4326.8
                                   min: 1.52s, max: 1.53s, spread: 0.01s
```

Additionally, the tail-read path optimizations demonstrated a **3.8x** improvement in comparison to the existing `main` branch for a 50M row file:

 | Run | Current | Proposed |
  |---:|---:|---:|
  | 1 | 12.3 ms | 2.6 ms |
  | 2 | 7.3 ms | 2.2 ms |
  | 3 | 6.6 ms | 2.2 ms |
  | 4 | 9.2 ms | 2.0 ms |
  | 5 | 5.8 ms | 2.0 ms |
  | **Average** | **8.3 ms** | **2.2 ms** |
  
## Follow Up
While these changes address the underlying data corruption potential and the fast-path tail-read optimization in some cases, it doesn't address the latter in cases where no OffsetIndex is available. A separate issue is being created to look into that optimization.

## Checklist


rionmonster:
@gunnarmorling 

I think I have both the data-corruption bug verified and fixed within the current changes. The bug was both confirmed to exist and subsequently fixed and the initial pass at tail-read performance yielded some solid performance benefits (in the cases where they were supported). 

It's worth noting that the tail-read optimizations only the `OffsetIndex` cases, not the ones without it. I'm going to spin off #371 to work on that effort separately and close that potential performance gap.

Otherwise, feel free to take a look when you have some time.

gunnarmorling:
Excellent work @rionmonster! I've rebased and added one more commit with a few minuscule adjustments. Opened PR #384 for it. I think we can merge it? Gonna close the one here in favor of the rebase. Thanks a lot, @rionmonster, really great to have this nasty bug caught and fixed.

rionmonster:
@gunnarmorling 

Yup, merging it should be fine IMO. All of the CI checks pass.

Issue #375, opened 2026-04-27
ColumnWorker.close() returns without joining its threads when the closing thread is interrupted


bmscomp:
### Summary

`ColumnWorker.close()` returns without waiting for its retriever and drain threads when the thread calling it is interrupted. The retriever can then still be reading the `InputFile` after `close()` has returned, and after the caller has closed that file.

### Cause

`close()` joins both threads in one `try` block and gives up on the first `InterruptedException`:

```java
try {
    retrieverThread.join();
    drainThread.join();
}
catch (InterruptedException e) {
    Thread.currentThread().interrupt();
}
```

If the closing thread is interrupted during the join, or already carries the interrupt flag when it calls `close()`, `retrieverThread.join()` throws and neither thread is joined. `close()` then snapshots `inFlightDecodes` while the retriever may still be running:

- The retriever may be inside `PageSource.next()`, reading the `InputFile`. It keeps going after `close()` returns.
- A decode task passed to `CompletableFuture.runAsync` but not yet added to `inFlightDecodes` can be running with `done` still false, and is missing from the set `close()` waits on.

The already-interrupted case is the common one: a task cancelled via `Future.cancel(true)`, `ExecutorService.shutdownNow()` or a `StructuredTaskScope` shutdown carries the interrupt flag into the try-with-resources block that closes the reader. `join()` then throws on entry, every time.

`CompletableFuture.allOf(pending).join()`, which follows, is not affected by the interrupt flag: `join()` is not interruptible and waits for every future in the snapshot.

### Impact

`close()` documents that once it returns, the caller may release the `InputFile`'s resources. That guarantee does not hold for an interrupted caller. None of the current `InputFile` implementations frees memory explicitly (mapped, direct and heap buffers are all released by GC), so there is no crash or data corruption. The observable effect is work continuing after the reader is closed: a page read or S3 range fetch against a closed channel or client, whose failure is signalled to an exchange nobody reads any more.

### Fix

Join both threads even when interrupted, remember the interrupt, and restore the flag after the in-flight decodes have been awaited. Once the retriever is always joined, every decode future it submitted is in `inFlightDecodes` before the snapshot is taken.

gunnarmorling:
Closed via #1255.

PR #379, opened 2026-04-28
#375 ColumnWorker.close() returns early when the calling thread is interrupted, leaving decode tasks running

bmscomp:
A decode task submitted via CompletableFuture.runAsync() could complete and
    trigger its whenComplete removal before inFlightDecodes.add() ran on the
    retriever thread. close() then missed the future in toArray() and returned
    before the task had actually finished — a use-after-free hazard on mapped
    or direct-allocated byte buffers released by InputFile.close().
    
    Fix: allocate a CompletableFuture<Void> ticket and add it to inFlightDecodes
    *before* calling runAsync. The whenComplete callback completes and removes
    the ticket, so close()'s allOf() always waits on every submitted decode.
    
    Also tighten closeJoinsThreadsAndAwaitsInFlightDecodes: track the virtual
    threads started by the stalled executor and join them before asserting
    decodesFinished, giving them time to run the post-command bookkeeping
    that falls outside the CompletableFuture scope.

test: add regression coverage for ColumnWorker.close() interrupt-flag bug
    
    closeCompletesDecodesDespiteCallerInterrupt exercises the specific failure
    mode fixed by 'defer interrupt re-assertion in ColumnWorker.close()':
    
      1. A decode executor is stalled so tasks are in flight when close() is called.
      2. The thread calling close() is interrupted while blocked in join().
      3. The test asserts close() does NOT return early (tasks are still running).
      4. After the tasks are released close() completes, and:
           - all decode tasks have finished (decodesFinished == decodesEntered)
           - the interrupt flag is restored on the caller thread
    
    Without the fix, CompletableFuture.allOf().join() sees the interrupt flag set
    by the immediate Thread.currentThread().interrupt() call and returns at once,
    leaving tasks executing against memory that may already have been freed.

Solves #375 

## Checklist

- [x] The commit message is prefixed with the issue key ("#375")


gunnarmorling:
Hey @bmscomp, did you run into this issue actually? Or could you create a reproducer? I asked Claude to review this PR and suggests that the described issue can't actually occur:

> Issue #375 and the PR description claim that CompletableFuture.allOf(pending).join() returns early (throws CancellationException) when the calling thread's interrupt flag is set on entry, so the immediate Thread.currentThread().interrupt() inside the catch block in close() causes the subsequent allOf().join() to skip waiting for in-flight decodes.
>
> That is not how CompletableFuture.join() behaves. join() calls waitingGet(false), which uses ForkJoinPool.managedBlock(). With interruptible=false, the only loop-exit condition is result != null. The interrupt flag is preserved across the call, never honoured to abort it. CancellationException is only thrown if a future is cancelled — not because the calling thread is interrupted.
>
> I verified this on JDK 25 with the exact pattern from the issue:
>
> After 500ms:
>   joiner alive (flag set on entry): true     ← still blocking, NOT short-circuited
>   taskFinished: false
> After release:
>   join() returned after 507 ms                ← returned only after task completed
>   CancellationException thrown: false         ← never thrown
>   interrupt flag preserved after join(): true
>
> So the SIGSEGV / use-after-free scenario described in the issue cannot occur via the path described.

I've yet to take a closer look myself but I wanna make sure we're not hunting phantoms here. Thanks a lot!

bmscomp:
@gunnarmorling  I think the bug occurs in some tests after the introduction of memory mapped files to be able to run larger files then the limit of 2GB, I will take time to make more test and experiment this in the other pull request that I am splitting, but I think this pull request must come the latest one, I need just some time to deep dive on it and I will come with clear explanation

ZhaoMJ:
Some findings from my Claude:

> The mechanism described in issue #375 contains an incorrect premise, which may explain why it is hard to reproduce: `CompletableFuture.join()` is **not** interruptible. It calls `waitingGet(false)` — the `false` means "not interruptible" — so it blocks until the future completes regardless of whether the calling thread's interrupt flag is set. `CancellationException` is only thrown if the future itself was `cancel()`-ed, not due to the joining thread's interrupt state. I verified this against the OpenJDK 25 source and by running the join under a pre-interrupted thread: the joiner stays blocked until the task finishes.
> 
> The **real** (narrower) race is:
> 
> 1. The calling thread is interrupted during `retrieverThread.join()`, which then throws `InterruptedException` — leaving the retriever still running.
> 2. `close()` falls through to `inFlightDecodes.toArray()` while the retriever is mid-submission.
> 3. A task submitted *after* the snapshot is not in the `allOf()` wait-set.
> 4. `allOf(snapshot).join()` correctly drains the captured tasks, then `close()` returns with the post-snapshot task still reading from buffers that `InputFile.close()` is about to release.
> 
> The ticket pre-registration fix is the correct remedy for this: every potential task has a ticket in `inFlightDecodes` *before* `runAsync()` is called, so even a snapshot taken mid-retriever accounts for every task that will ever be submitted. The existing test (`closeCompletesDecodesDespiteCallerInterrupt`) validates that the fix works in the interrupted-caller scenario, but it does not demonstrate a pre-fix failure — we were unable to write a deterministic failing test without instrumenting `add()` with a sleep, because the window requires the interrupt to land at a precise instruction boundary within `join()`.

gunnarmorling:
Took a new stab at this, via #1255, so I'm gonna close this one. Thanks for your efforts nevertheless, @bmscomp!

Issue #373, opened 2026-04-27
S3InputFile: bounded byte-range LRU cache


gunnarmorling:
## Summary

Add an LRU cache of fetched byte ranges to \`S3InputFile\`, bounded by a configurable byte budget (e.g. 256 MB by default). Today \`S3InputFile\` only caches the 64 KB tail prefetched at \`open()\`; every other \`readRange\` always issues a network GET — even when the same range was just fetched seconds earlier.

## Motivation

Spotted while finalising #361. Once the dive Data preview's row-cap fix lands, every cursor rebuild within a row group re-issues identical ranged GETs to S3. Backward-pagination at the tail of a file is the worst-affected workflow:

- User presses \`End\`. Cursor builds at \`fromRowGroup(lastRg)\`, fetches all column chunks of the last RG (~one RG worth of bytes).
- User presses \`PgUp\`. \`firstRow\` moves backward by \`pageSize\`, the cursor-reuse check fails (\`previewCursorPosition > firstRow\`), and the cursor is rebuilt at \`fromRowGroup(lastRg)\` again — re-fetching the same column chunks.
- Each subsequent \`PgUp\` repeats the re-fetch.

For Overture-shaped files this is ~13 MB per PgUp; for big-RG files it's ~52 MB+ per PgUp. The decode work is unavoidable, but the *bytes-over-the-wire* are identical to what a previous fetch already retrieved.

The same pattern applies to anything that builds multiple readers over a hot region — Inspect commands run in sequence, performance benchmarks that re-open the file, future replay/scrub workflows, etc.

## Proposed shape

A range cache at the \`S3InputFile\` layer:

- **Key:** \`[offset, offset+length)\` byte range.
- **Storage:** the \`ByteBuffer\` returned by \`readRange\`, held by reference. Direct buffers are fine; we already use them for tail.
- **Eviction:** LRU by total bytes, default capacity 256 MB. Configurable via system property (\`hardwood.s3.rangeCacheBytes\`) or programmatic API on \`S3InputFile\`.
- **Lookup behaviour:** exact-match first; for partial-overlap (a request that spans across cached and non-cached ranges), fall back to the network — splitting and merging adds complexity not justified at this stage. Worth revisiting if profiling shows it matters.
- **Coexistence with the tail cache:** the existing 64 KB tail cache stays as-is; the LRU sits in front of the network path only.

## Out of scope here

[...]

PR #385, opened 2026-04-28
#373 Whole-file range backing for remote InputFiles

gunnarmorling:


gunnarmorling:
@rionmonster, in case you're feeling adventurous: this adds a key improvement to the S3 story for the dive TUI, avoiding repeated retrieval of the same segments of a remote file. If you wanna test it, then just follow the instructions in [TESTING.md](TESTING.md). There's a request/bytes counter in the top bar which shows how much data has been fetched.

PR #387, opened 2026-04-28
#351 Adding UUID Filter Predicate and tighten SignedBinaryColumnPredicate

### LINKED ISSUE #351: SignedBinaryColumnPredicate accepts any FIXED_LEN_BYTE_ARRAY column, not just DECIMAL
## Problem

`SignedBinaryColumnPredicate` is intended for filtering **DECIMAL columns physically encoded as `FIXED_LEN_BYTE_ARRAY`**, where Parquet stores values in big-endian two's-complement and a *signed* byte comparison reproduces the correct numeric ordering. The resolver, however, only validates the **physical** type and ignores the **logical** type:

```java
// core/src/main/java/dev/hardwood/internal/predicate/FilterPredicateResolver.java
case FilterPredicate.SignedBinaryColumnPredicate p -> {
    ColumnSchema cs = resolveColumn(p.column(), schema);
    rejectRepeated(p.column(), cs);
    validateType(p.column(), PhysicalType.FIXED_LEN_BYTE_ARRAY, cs);   // ← physical-only
    yield new ResolvedPredicate.BinaryPredicate(cs.columnIndex(), p.op(), p.value(), true);
}
```

So the predicate is happily accepted against any `FIXED_LEN_BYTE_ARRAY` column. The signed byte comparator (`BinaryComparator.compareSigned`) then runs over min/max stats and per-row values whose bytes have nothing to do with a signed integer:

| Column logical type | Physical | Signed-integer comparison meaningful? |
|---|---|---|
| `DECIMAL` (FLBA encoding) | FLBA(N) | ✅ yes — intended use |
| `UUID` | FLBA(16) | ❌ UUIDs aren't ordered as signed 128-bit integers |
| `FLOAT16` | FLBA(2) | ❌ IEEE-754 half-float ordering ≠ signed integer ordering |
| `INTERVAL` (after #122) | FLBA(12) | ❌ three independent little-endian uint32s, not a single number |
| Bare FLBA (no logical type — WKB, custom encodings) | FLBA(N) | ❌ depends on the app's encoding |

## Failure modes

The same comparator is used both for row-group/page stats pruning (`StatisticsFilterSupport`) and for row-level evaluation (`RecordFilterEvaluator`). When applied to a non-DECIMAL FLBA column the predicate produces:

1. **False negatives in pruning** — row groups silently dropped because their (meaningless) min/max comparison says \"can't contain a match.\" The user gets fewer rows than they should and there is no signal that anything is wrong.
2. **Incorrect inclusions/exclusions per row** — the row-level evaluator reproduces the same wrongness for individual rows.

The wrongness is only triggered when a caller manually constructs a `SignedBinaryColumnPredicate` against a non-DECIMAL column, so this is not a remote-exploit class of bug — but it is a quiet correctness hole.

## Proposal

Two options, not mutually exclusive. Pick one for this issue and split the other if needed.

**Option A — gate by logical type (smaller fix, conservative).** Tighten the `SignedBinaryColumnPredicate` branch in `FilterPredicateResolver` so it accepts only columns whose `logicalType()` is `LogicalType.DecimalType`. Reject everything else with a clear message, e.g. *\"signed binary comparison is only valid for DECIMAL columns; got UUID\"*. This is the safe default and matches the predicate's documented intent.

**Option B — introduce typed predicates per logical type.** Add `UuidColumnPredicate`, `IntervalColumnPredicate`, `Float16ColumnPredicate`, …, each routed to a comparator that matches the column's actual semantics:
- UUID → lexicographic compare (already correct for `compareUnsigned`).
- FLOAT16 → IEEE-754 ordering (sign-bit-aware, NaN handling).
- INTERVAL → reject; the spec defines no sort order.

Option A is the right minimum bar; Option B is the proper public-API direction.

## Tests

For Option A:

- Construct a `SignedBinaryColumnPredicate` against a UUID, FLOAT16, INTERVAL, and bare FLBA column. Each should throw at resolve time.
- Construct it against a DECIMAL FLBA column — should still resolve and behave as today.

## Why this is filed now

Surfaced during review of #122 (INTERVAL support). Adding INTERVAL doesn't introduce the bug — it just adds one more logical type to the list of FLBA columns that can be silently filtered with a wrong comparator. Worth tracking on its own.

iifawzi:
Hello, this PR resolves #351, adding a UUID Filter Predicate and tightening SignedBinaryColumnPredicate to logical type decimal, and a very slight performance improvement of the binary comparator 

## Checklist

- [ x The commit message is prefixed with the issue key ("#123 Adding support for...")

### DIFF
--- a/core/src/main/java/dev/hardwood/internal/predicate/BinaryComparator.java
+++ b/core/src/main/java/dev/hardwood/internal/predicate/BinaryComparator.java
@@ -7,6 +7,8 @@
  */
 package dev.hardwood.internal.predicate;
 
+import java.util.Arrays;
+
 /// Byte array comparison utilities for filter predicate evaluation.
 ///
 /// Provides unsigned lexicographic comparison (for BYTE_ARRAY) and signed
@@ -21,14 +23,7 @@ public final class BinaryComparator {
     ///
     /// @return negative if a < b, zero if equal, positive if a > b
     public static int compareUnsigned(byte[] a, byte[] b) {
-        int minLen = Math.min(a.length, b.length);
-        for (int i = 0; i < minLen; i++) {
-            int cmp = (a[i] & 0xFF) - (b[i] & 0xFF);
-            if (cmp != 0) {
-                return cmp;
-            }
-        }
-        return a.length - b.length;
+        return Arrays.compareUnsigned(a, b);
     }
 
     /// Compare two same-length byte arrays as big-endian two's complement signed values.
@@ -36,6 +31,7 @@ public final class BinaryComparator {
     ///
     /// @return negative if a < b, zero if equal, positive if a > b
     public static int compareSigned(byte[] a, byte[] b) {
+        int len = a.length;
         if (a.length != b.length) {
             throw new IllegalArgumentException(
                     "Signed binary comparison requires same-length arrays: " + a.length + " vs " + b.length);
@@ -49,12 +45,6 @@ public final class BinaryComparator {
             return cmp;
         }
         // Remaining bytes: compare as unsigned
-        for (int i = 1; i < a.length; i++) {
-            cmp = (a[i] & 0xFF) - (b[i] & 0xFF);
-            if (cmp != 0) {
-                return cmp;
-            }
-        }
-        return 0;
+        return Arrays.compareUnsigned(a, 1, len, b, 1, len);
     }
 }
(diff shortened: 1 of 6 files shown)


iifawzi:
I will add the Float16 Filter in a separate PR, as the binary comparator would need to have an additional method to compare IEEE-754 bytes correctly

gunnarmorling:
SGTM!

gunnarmorling:
Thanks a lot! LGTM, pushed one more commit with some clean-up.

gunnarmorling:
And applied; leaving the issue open for the FLOAT fix, too. Thx, @iifawzi!

Issue #193, opened 2026-03-30
Explore vectorized record level filtering


gunnarmorling:
Currently, record-level filters are evaluated value by value (see `RecordFilterEvaluator`); there might be an upside to be had by doing this in batches instead.

iifawzi:
I started exploring that with claude couple of hours ago out of curiosity, and numbers are impressive, although it's not yet 100% aligning with your proposed approach yet.  

Sharing some numbers for now, i would try to refine and share an initial PR tomorrow's night 

Legacy is today's code, stage one is a first iteration where it was looping per-row compiled matcher:

```
return row -> {
    for (int i = 0; i < matchers.length; i++) {
        if (!matcher[i].test(row)) return false;
    }
    return true;
};
```

stage two is a single compiled matcher, which's ended way way way faster. 
```
    private static final class And2Matcher implements RowMatcher {
        private final RowMatcher a;
        private final RowMatcher b;
        And2Matcher(RowMatcher a, RowMatcher b) {
            this.a = a;
            this.b = b;
        }
        @Override
        public boolean test(StructAccessor row) {
            return a.test(row) && b.test(row);
        }
    }
```

<img width="755" height="506" alt="Image" src="https://github.com/user-attachments/assets/249149de-54fa-4738-be6c-c5191809da1d" />

However end to end still showing a very slight to negligible improvement: 

<img width="493" height="219" alt="Image" src="https://github.com/user-attachments/assets/be696fee-7133-4859-983a-1593d92635d2" />

and discussing that with claude it suspects the struct accessor hash lookups, so tomorrow i will give it an additional push to improve that and see how it would affect e2e numbers
```
    private static RowMatcher intLeaf(String[] path, String name, Operator op, int v) {
        return switch (op) {
            case EQ -> row -> { StructAccessor a = resolve(row, path); return a != null && !a.isNull(name) && a.getInt(name) == v; };
            case NOT_EQ -> row -> { StructAccessor a = resolve(row, path); return a != null && !a.isNull(name) && a.getInt(name) != v; };
            case LT -> row -> { StructAccessor a = resolve(row, path); return a != null && !a.isNull(name) && a.getInt(name) < v; };
            case LT_EQ -> row -> { StructAccessor a = resolve(row, path); return a != null && !a.isNull(name) && a.getInt(name) <= v; };
            case GT -> row -> { StructAccessor a = resolve(row, path); return a != null && !a.isNull(name) && a.getInt(name) > v; };
[...]

PR #399, opened 2026-04-29
#193 Optimize record filtering by removing per-row dispatch overhead

iifawzi:
Hello, this PR introduces the first iteration of record filter compilation to eliminate per-row predicate dispatch

I explored several variants, and while the micro-benchmark improvements are significant, the end-to-end gains are not as high. Overall, the current result feels like a good trade-off between performance and complexity.

### End-to-end (`RecordFilterBenchmarkTest -Dperf.runs=10`)

before: 

| Contender | Avg time | Records/sec |
|---|---:|---:|
| No filter (baseline) | 17.2 ms | 580,396,520 |
| Match-all filter (worst case) | 85.6 ms | 116,809,239 |
| Selective filter (id < 1%) | 3.4 ms | 29,745,674 |
| Compound match-all (id>=0 AND value<+inf) | 251.4 ms | 39,778,639 |
| Page+record (id range AND value<500) | 7.2 ms | 6,938,212 |

After: 

| Scenario | Avg time | Records/sec |
|---|---:|---:|
| No filter (baseline) | 16.1 ms | 622,640,291 |
| Match-all (1 leaf, `id >= 0`) | 26.6 ms | 375,664,925 |
| Selective (`id < 1%`) | 1.8 ms | 54,317,317 |
| Compound match-all (`id >= 0 AND value < +inf`) | 47.0 ms | 212,980,449 |
| Page+record combined (`id BETWEEN ... AND value < 500`) | 2.8 ms | 17,858,593 |

### JMH micro-benchmark (`RecordFilterMicroBenchmark`)

2 forks × 5 warmup × 5 measurement iterations. `ns/op` is per-row cost over a 4096-row in-memory batch (`@OperationsPerInvocation(BATCH_SIZE)`). Match-all predicates only — every row evaluated, so dispatch overhead is the entire cost.

| Shape | Legacy ns/op | Compiled ns/op | Speedup |
|---|---:|---:|---:|
| `single` (`id >= 0`) | 2.693 ± 0.014 | **0.509 ± 0.008** | **5.29×** |
| `and2` | 10.357 ± 0.435 | **0.576 ± 0.002** | **17.98×** |
| `and3` | 21.999 ± 0.133 | **0.550 ± 0.009** | **40.00×** |
| `and4` | 32.666 ± 0.219 | **0.633 ± 0.005** | **51.60×** |
| `or2` | 4.755 ± 0.195 | **0.513 ± 0.011** | **9.27×** |
| `nested` (and-of-or) | 33.872 ± 0.266 | **0.981 ± 0.007** | **34.53×** |
| `intIn5` | 3.769 ± 0.010 | **1.441 ± 0.022** | **2.62×** |
| `intIn32` | 7.040 ± 0.046 | **3.196 ± 0.011** | **2.20×** |

I also experimented with batch-oriented RowReader changes, but they added complexity without measurable gains, so they were reverted.

I explored an additional optimization layer: manual predicate fusion for common binary cases (e.g. int AND int, long AND long), generating specialized fused lambdas that remove inner virtual calls entirely and rely on direct primitive comparisons

| Method | Scenario | Legacy | Stage 1 | Stage 2 | Stage 2.5 | **Stage 3** | **Total speedup** |
|---|---|---:|---:|---:|---:|---:|---:|
| End-to-end | No filter (sanity) | 17.2 ms | 18.2 ms | 15.1 ms | 19.3 ms | 19.3 ms | flat |
| End-to-end | Match-all (1 leaf) | 85.6 ms | 68.6 ms | 67.7 ms | 74.1 ms | **31.5 ms** | **2.72×** |
| End-to-end | Selective | 3.4 ms | 3.5 ms | 3.4 ms | 3.7 ms | **2.8 ms** | **1.21×** |
| End-to-end | Compound (2-leaf AND) | 251.4 ms | 115.9 ms | 105.7 ms | 97.6 ms | **46.6 ms** | **5.39×** |
| End-to-end | Page + record combined | 7.2 ms | 5.1 ms | 4.8 ms | 5.1 ms | **3.9 ms** | **1.85×** |
| JMH (no I/O, monomorphic) | single | 2.696 | 0.530 | 0.501 | 0.501 | **0.501 ns/op** | **5.4×** |
| JMH (no I/O, monomorphic) | and2 | 10.186 | 2.192 | 0.568 | 0.601 | **0.601 ns/op** | **17.0×** |
| JMH (no I/O, monomorphic) | and3 | 21.595 | 6.429 | 0.566 | 0.572 | **0.572 ns/op** | **37.8×** |
| JMH (no I/O, monomorphic) | and4 | 31.906 | 11.445 | 0.652 | 0.657 | **0.657 ns/op** | **48.6×** |
| JMH (no I/O, monomorphic) | or2 | 4.515 | 0.912 | 0.506 | 0.506 | **0.506 ns/op** | **8.9×** |
| JMH (no I/O, monomorphic) | nested | 32.447 | 10.553 | 0.924 | 1.047 | **1.047 ns/op** | **31.0×** |
| JMH (no I/O, monomorphic) | intIn5 | 3.690 | 1.378 | 1.385 | 1.385 | **1.385 ns/op** | 2.7× |
| JMH (no I/O, monomorphic) | intIn32 | 6.845 | 3.189 | 3.113 | 3.113 | **3.113 ns/op** | 2.2× |
| **JMH (no I/O, megamorphic)** | **8 mixed AND2 shapes** | n/a | n/a | 4.604 | **2.127** | 2.127 ns/op | **2.16×** (Stage 2.5 over Stage 2) |

While it improves performance for tests that include megamorphic virtual calls, it comes with the cost of code complexity (only supporting AND), so given that without it, the improvements are so great already, I decided not to include it in this PR at least 

However, the full experiment (code + benchmarks + design notes) is [available here](https://github.com/iifawzi/hardwood/commit/e3f4c601c67e1e78d25015654e353a171886b177#diff-a4e3d547a35a79823b351acd39ca606cea246ec4a3ce27a91e22a190039d9591). Perhaps you can find it useful to build on top of it @gunnarmorling. There are also detailed notes that I asked Claude to document in RECORD_FILTER_COMPILATION_RESULTS.md and RECORD_FILTER_COMPILATION.md throughout the experimentation. 

## Checklist


iifawzi:
I left RecordFilterEvaluator as it's used for comparisons in the tests, but we can remove it if we finally proceed with the compiler

iifawzi:
This weekend I will explore again adding the fused matchers improvement, branching on top of this PR, I will push claude to do the full implementation of all the most useful 2-arity combinations (int-int, long-long,boolean-boolean, int-long, etc..), and write more realistic tests with usage of different matchers to simulate real scenarios where JVM wouldn't inline existing matchers

gunnarmorling:
That sounds great! I suppose we should still try to get in this part here, or do you think it'd be better to land all at once? Great stuff overall, really excited to see this effort!

iifawzi:
> That sounds great! I suppose we should still try to get in this part here, or do you think it'd be better to land all at once? Great stuff overall, really excited to see this effort!

Yes, the improvement is truly impressive, I got super excited working on that as well
What part you mean that we should have it here? the fused matchers? I believe I can have an additional commit by end of today, and then we can decide if we would like to have it or not. 

My only concern is maintaining that part of fused matchers, because it would be a lot, but if the gain is high, we can accept that as a tradeoff, let's see, I will commit that later today when it's ready, and we can iterate, improve and take a decision

gunnarmorling:
Thinking about it, could we perhaps break this further down; I think optimizing indexed-based access (stages 1 and 3) will be a significant win on its on, and probably will be a much smaller change. Could we get this in as a first PR? Kinda annoying that this current implementation slipped through, it's pretty bad, LOL.

iifawzi:
> Thinking about it, could we perhaps break this further down; I think optimizing indexed-based access (stages 1 and 3) will be a significant win on its on, and probably will be a much smaller change. Could we get this in as a first PR? Kinda annoying that this current implementation slipped through, it's pretty bad, LOL.

haha, at least now you can confidently say it's an "ULTRA" fast minimal dependency implementation of Apache Parquet" :d 

btw, I've just seen your comment, and I've already pushed stage 4, numbers are beyond imagination, that I had to run it twice lol. 

so finally, it's just two fusion files, the normal one, and the indexed one, and there's a python script to generate lambdas, so maintainability isn't a big concern as long as I have my assistant mr claude in a good mode. 

Numbers as expected not showing huge difference for monomorphic sites, given the JVM is already doing its job 

| Shape | Legacy ns/op | Compiled ns/op | Fused ns/op | Speedup |
|---|---:|---:|---:|---:|
| `single` (`id >= 0`) | 2.703 ± 0.033 | 0.420 ± 0.004 | **0.523 ± 0.003** | **5.17×** |
| `and2` | 10.299 ± 0.053 | 0.488 ± 0.003 | **0.584 ± 0.006** | **17.63×** |
| `and3` | 22.141 ± 0.104 | 0.596 ± 0.023 | **0.551 ± 0.006** | **40.18×** |
| `and4` | 32.264 ± 0.874 | 0.672 ± 0.011 | **0.649 ± 0.009** | **49.71×** |
| `or2` | 4.628 ± 0.044 | 0.418 ± 0.006 | **0.517 ± 0.003** | **8.95×** |
| `nested` (and-of-or) | 33.766 ± 0.484 | 1.010 ± 0.009 | **1.002 ± 0.021** | **33.70×** |
| `intIn5` | 3.801 ± 0.052 | 1.433 ± 0.066 | **1.404 ± 0.025** | **2.71×** |
| `intIn32` | 7.016 ± 0.026 | 3.176 ± 0.014 | **3.186 ± 0.030** | **2.20×** |

where fused showing a slight increase. However for megamorphic sites, the speedup is significant as expected as well 

| Arm                  | ns/op (per row × per shape) | Speedup vs legacy | Speedup vs generic |
|----------------------|----------------------------:|------------------:|-------------------:|
| `legacyMegamorphic`  | 19.150 ± 0.239              | 1.00×             | —                  |
| `genericMegamorphic` |  5.407 ± 0.074              | 3.54×             | 1.00×              |
| `fusedMegamorphic`   | **3.031 ± 0.035**           | **6.32×**         | **1.78×**          |

also end-to-end showing a good speed up 

| Shape                                                 | Generic ms | Fused ms | Speedup |
|-------------------------------------------------------|-----------:|---------:|--------:|
| `id BETWEEN 1M and 4M` (long+long AND)                |       30.7 |     14.1 |   2.18× |
| `id < 500K OR id > 9.5M` (long+long OR)               |       10.6 |      9.2 |   1.15× |
| `tag BETWEEN 0 and 50` (int+int AND)                  |      120.8 |     69.9 |   1.73× |
| `tag = 5 OR tag = 47` (int+int OR)                    |       58.0 |     42.3 |   1.37× |
| `value BETWEEN 0 and 500` (double+double AND)         |      126.7 |     83.7 |   1.51× |
| `id < 5M AND value < 500` (long+double AND)           |       61.4 |     48.7 |   1.26× |
| `id < 5M OR value > 500` (long+double OR)             |       72.7 |     69.4 |   1.05× |
| `tag < 50 AND id > 5M` (int+long AND)                 |       51.3 |     39.6 |   1.30× |
| `tag < 50 AND value < 500` (int+double AND)           |      120.0 |     98.0 |   1.22× |
| `flag = true AND flag != false` (boolean+boolean AND) |       99.4 |     75.7 |   1.31× |
| `value > 0 AND id < 9999` (double+long AND, swap)     |        1.0 |      0.9 |   1.11× |
| `bin BETWEEN k200 and k800` (binary+binary AND)       |      178.5 |    103.6 |   1.72× |
| **Aggregate (60 shape×run combinations)**             | **4655.7** | **3276.0** | **1.42×** |


I'd say we can go with all that in general, but as you prefer, we can ship them in parts. Why you were thinking about omitting stage 2 from the first go?

gunnarmorling:
> I'd say we can go with all that in general, but as you prefer, we can ship them in parts. Why you were thinking about omitting stage 2 from the first go?

Going step by step will help a bit in terms of reviewing and understanding the change. Arguably, the current repeated--and for top-level fields unnecessary--name retrieval is borderline a bug; it should never have found its way into the code. Whereas the predicate fusing is a clever optimization, independent from that. So if we can untangle this into two PRs, that would be great.

Btw. massive thank you for working on this, it's really amazing to see this!

iifawzi:
I totally agree with you, then let me refine it as you proposed, we can switch this PR to only step 1, and 3 (or 2), and a next PR having this as a base (until getting merged) for stage 2(or 3) and 4, does it sounds good to you?

gunnarmorling:
Yes, that would be perfect!

iifawzi:
Done, now this branch includes stages 1,3, results are as following:

| Shape | Legacy ns/op | Compiled ns/op | Speedup |
|---|---:|---:|---:|
| `single` (`id >= 0`) | 2.691 ± 0.020 | **0.516 ± 0.005** | **5.21×** |
| `and2` | 9.069 ± 0.086 | **2.040 ± 0.034** | **4.45×** |
| `and3` | 22.146 ± 0.186 | **8.085 ± 0.086** | **2.74×** |
| `and4` | 32.022 ± 0.519 | **13.446 ± 0.068** | **2.38×** |
| `or2` | 4.607 ± 0.024 | **0.839 ± 0.007** | **5.49×** |
| `nested` (and-of-or) | 33.213 ± 0.404 | **10.702 ± 0.221** | **3.10×** |
| `intIn5` | 3.796 ± 0.025 | **1.387 ± 0.027** | **2.74×** |
| `intIn32` | 7.035 ± 0.045 | **3.180 ± 0.029** | **2.21×** |

(Errors are 99.9 % confidence intervals from JMH.)

Adding the next PR on #412, can be rebased to main after this one

iifawzi:
Old implementation isn't removed as it's used for testing, we can remove it I believe before merging next PR

gunnarmorling:
Great stuff. I've pushed one more commit which a) removes `IndexedAccessor` (everything needed is exposed via `RowReader` already) and b) applies the optimization also to predicates targeting top-level fields when using the nested row reader.

gunnarmorling:
Rebased an applied. Thanks a lot, @iifawzi!

Issue #250, opened 2026-04-15
Push record-level filter evaluation to column worker drain threads


gunnarmorling:
## Summary

Record-level filter predicates can be evaluated in the column worker drain threads, where decoded values are already in L1 cache, rather than post-aggregation on the consumer side.

## Design

1. **Predicate decomposition**: At pipeline creation, decompose the `ResolvedPredicate` into per-column fragments. Each leaf predicate already references a single column index.
2. **Drain-side evaluation**: After the drain copies a page's values into the batch, evaluate the column's predicate fragment and produce a per-batch `BitSet` of matching record indices.
3. **Aggregation**: In `loadNextBatch()`, intersect all columns' `BitSet`s into a combined result. Columns without a predicate fragment are skipped (implicitly all-ones).
4. **Iteration**: `hasNext()`/`next()` use `nextSetBit` on the combined `BitSet` to skip non-matching rows — no wrapper class needed.

## OR predicates across columns

Top-level AND (the common case) partitions cleanly: each column evaluates its part, results are intersected. For OR predicates spanning multiple columns, fall back to post-aggregation evaluation on the combined batch. Mixed trees (AND of ORs) can push down the pushable parts and evaluate the rest post-aggregation.

## Advantages

- Filter evaluation parallelized across column worker threads
- Evaluated on hot cache lines (data just decoded by the drain)
- No wrapper class, no interface dispatch, no delegate overhead
- Reader iteration hot path stays monomorphic

Issue #238, opened 2026-04-14
Explore producer-side record-level filter evaluation for FlatRowReader


gunnarmorling:
## Context

Record-level filtering in `FlatRowReader` currently runs on the consumer side: after all columns' batch arrays are received, `RecordFilterEvaluator.matchBatch()` evaluates the predicate and non-matching rows are skipped (via `FilteredFlatRowReader`). This works but adds per-batch overhead on the consumer thread.

## Idea

Evaluate the record filter on the producer side, before batch arrays reach the consumer. The challenge: each column's drain only sees its own column's data, but the filter predicate may reference multiple columns (e.g., `fare_amount > 10 AND trip_distance < 5`).

## Possible approaches

1. **Cross-column synchronization barrier.** All column drains produce their batch arrays into a shared staging area. A coordinator thread (or one of the drains) evaluates the filter across all columns, compacts the arrays, then publishes filtered batches to the consumer queues. Adds a sync point but keeps the consumer completely filter-free.

2. **Two-pass assembly.** First pass: all drains assemble their batch arrays normally. Second pass (on a dedicated thread): read all columns' arrays, evaluate the filter, compact in-place, adjust record counts. Similar to approach 1 but without a barrier — uses the existing ready queues with a filter stage between them and the consumer.

3. **Predicate decomposition.** Split the predicate into per-column predicates where possible (e.g., `fare_amount > 10` only needs column 2). Each drain evaluates its column's sub-predicate and produces a `BitSet`. A coordinator ANDs the per-column BitSets to get the final match set, then signals all drains to compact. Requires predicate analysis and cross-column BitSet coordination.

## Considerations

- The current consumer-side approach via `FilteredFlatRowReader` has minimal overhead (~one `matchBatch` call per batch, ~2500 batches per run). It may not be worth optimizing further unless filtering is extremely selective (few matches per batch).
- Producer-side filtering could reduce the batch size delivered to the consumer, meaning fewer rows to iterate — but the savings depend on filter selectivity.
- Any cross-column coordination adds complexity and synchronization overhead that may negate the benefits for low-selectivity filters.

## Related

- `FilteredFlatRowReader` — current consumer-side implementation
- `RecordFilterEvaluator.matchBatch()` — the filter evaluation logic
- `ColumnWorker` drain thread — the per-column assembly stage

PR #412, opened 2026-04-30
#193 Optimize record filtering using fixed arity matchers 

iifawzi:
This PR is the follow up on #399, however as the branch isn't in this repository, i can't have it as a base, so it includes all the commits, should be rebased and merged after #399
This PR adds stage 3, and 4 for the fusion, results are as following: 

### JMH micro-benchmark (`RecordFilterMicroBenchmark`)

| Shape | Legacy ns/op | Compiled ns/op | Fused ns/op | Speedup |
|---|---:|---:|---:|---:|
| `single` (`id >= 0`) | 2.619 ± 0.025 | 0.507 ± 0.002 | **0.506 ± 0.006** | **5.18×** |
| `and2` | 10.005 ± 0.018 | 0.579 ± 0.015 | **0.583 ± 0.006** | **17.16×** |
| `and3` | 21.396 ± 0.179 | 0.563 ± 0.012 | **0.558 ± 0.013** | **38.34×** |
| `and4` | 30.802 ± 0.578 | 0.636 ± 0.011 | **0.638 ± 0.012** | **48.28×** |
| `or2` | 4.466 ± 0.007 | 0.507 ± 0.003 | **0.504 ± 0.024** | **8.86×** |
| `nested` (and-of-or) | 31.821 ± 0.191 | 0.934 ± 0.020 | **0.932 ± 0.069** | **34.14×** |
| `intIn5` | 3.696 ± 0.017 | 1.351 ± 0.021 | **1.390 ± 0.077** | **2.66×** |
| `intIn32` | 6.724 ± 0.091 | 3.085 ± 0.013 | **3.092 ± 0.015** | **2.17×** |

### JMH micro — `RecordFilterMegamorphicBenchmark` (megamorphic three-way)

| Arm                  | ns/op (per row × per shape) | Speedup vs legacy | Speedup vs generic |
|----------------------|----------------------------:|------------------:|-------------------:|
| `legacyMegamorphic`  | 18.145 ± 0.168              | 1.00×             | —                  |
| `genericMegamorphic` |  5.120 ± 0.009              | 3.54×             | 1.00×              |
| `fusedMegamorphic`   | **3.068 ± 0.077**           | **5.91×**         | **1.67×**          |

### End-to-end — `RecordFilterMegamorphicEndToEndTest`

| Shape                                                 | Generic ms | Fused ms | Speedup |
|-------------------------------------------------------|-----------:|---------:|--------:|
| `id BETWEEN 1M and 4M` (long+long AND)                |       23.7 |     13.9 |   1.71× |
| `id < 500K OR id > 9.5M` (long+long OR)               |        6.0 |      6.5 |   0.92× |
| `tag BETWEEN 0 and 50` (int+int AND)                  |      107.8 |     72.3 |   1.49× |
| `tag = 5 OR tag = 47` (int+int OR)                    |       55.3 |     41.7 |   1.33× |
| `value BETWEEN 0 and 500` (double+double AND)         |      110.2 |     78.9 |   1.40× |
| `id < 5M AND value < 500` (long+double AND)           |       55.4 |     42.9 |   1.29× |
| `id < 5M OR value > 500` (long+double OR)             |       63.6 |     65.4 |   0.97× |
| `tag < 50 AND id > 5M` (int+long AND)                 |       47.7 |     40.2 |   1.19× |
| `tag < 50 AND value < 500` (int+double AND)           |      116.7 |     94.5 |   1.23× |
| `flag = true AND flag != false` (boolean+boolean AND) |       91.3 |     74.6 |   1.22× |
| `value > 0 AND id < 9999` (double+long AND, swap)     |        1.1 |      0.9 |   1.22× |
| `bin BETWEEN k200 and k800` (binary+binary AND)       |      173.1 |    101.5 |   1.71× |
| **Aggregate (60 shape×run combinations)**             | **4259.1** | **3166.6** | **1.35×** |

## Checklist


gunnarmorling:
@iifawzi, could you rebase this one to current main?

core/src/main/java/dev/hardwood/internal/predicate/RecordFilterCompiler.java
```diff
@@ -24,9 +24,16 @@
...
-    private static final String[] EMPTY_PATH = new String[0];
+    static final String[] EMPTY_PATH = new String[0];
     private static final int IN_LIST_BINARY_SEARCH_THRESHOLD = 16;
 
+    /// Compile-time fusion of arity-2 AND/OR leaves into per-tuple synthetic
+    /// classes. Default true; set `-Dhardwood.recordfilter.fusion=false` to
+    /// route through the generic [And2Matcher] / [Or2Matcher] for benchmark
+    /// A/B comparison. Read once at class init.
+    static final boolean FUSION_ENABLED =
+            !"false".equalsIgnoreCase(System.getProperty("hardwood.recordfilter.fusion"));
```
iifawzi: I believe we can remove it if we wouldn't need to run benchmarks with fusion disabled again

iifawzi:
@gunnarmorling, now is ready, however I understand this might be a lot as well, if you want we can split even further so here's only the fixed arity classes, and the next final one to have the fusion

gunnarmorling:
@iifawzi, so why are there no significant differences in the first table between "compiled" and "fused"?

gunnarmorling:
Btw. you might be interested in these potential follow-ups: #238, #250 (probably duplicates). The essential idea being to push row-level filter evaluation to decoder threads. Some workloads (e.g. `FlatPerformanceTest` with three projected columns) are consumer bound, so moving filtering logic from the consumer thread should be beneficial in terms of end-to-end latency.

gunnarmorling:
Hey @iifawzi, the following is from Claude. I think the Method Handle idea is intriguing, curious about your thoughts. I tend to agree about the maintainability/scalability concern of the current proposal.

--

This is feedback on the architectural direction of arity-2 record-filter fusion, separate from the line-level review in `PR-412-review.md`.

The arity-2 fusion direction is the right next step after Stage 1–3. The reasoning chain — outer megamorphism is residual, inner megamorphism is the real cost, fix it by emitting one synthetic class per `(typeA, opA, typeB, opB, connective, accessMode)` tuple — is correct and is what the JMH `genericMegamorphic → fusedMegamorphic` 1.67× delta actually measures. The architecture is sound. The feedback below is on **scope, mechanism, and where the design effort is being spent**, not on whether to do it.

## 1. The matrix is the wrong unit of optimization

The PR builds out 7 type combos × {AND, OR} × {same-column, different-column} × 36 op pairs × {name, indexed} ≈ 1500 lambdas. That's a lot of bytecode for a static decision tree the user can't influence. Three observations:

- **The wins concentrate.** Look at the end-to-end table: the meaningful gains (>1.4×) are on same-column AND-BETWEEN shapes — `id BETWEEN`, `tag BETWEEN`, `value BETWEEN`, `bin BETWEEN`. Those are 4 shapes out of 12. Cross-type combos contribute 1.19×–1.29×, OR shapes regress or are flat, and one shape moves 0.1ms. The 1.35× aggregate is mostly the BETWEENs.
- **The matrix scales with combinatorics, the wins don't.** Adding `float+float`, `IsNull` paired with leaves, etc. (Option B in the doc) doubles the matrix without doubling the win — most workloads don't hit the new combos often enough to matter.
- **The body of every fused class is the same shape.** Load value(s), do two primitive comparisons, AND/OR. The 36-lambda explosion encodes one sentence of logic.

If I were doing this fresh, I would land just **same-column same-type AND** as the first fusion target (handles BETWEEN, the biggest win) and not build the cross-type or OR matrix until there's a workload that demonstrates the gain. The OR shapes especially: the doc shows two regressions and the JMH micro shows fusion is essentially flat for `or2`. There's no evidence OR fusion pays for itself.

## 2. Hand-typed lambda cases are the wrong mechanism for the matrix

The PR commits to "one source file with 1500 lambda cases as the JIT-friendly artifact." This works but creates ongoing pain:

- **Stack traces from a lambda body point to one giant line.** Already noted as a smaller concern in the review, but it's bigger than that — when something goes wrong (`NullPointerException` on a malformed predicate), debugging requires reading through `case GT_EQ -> row -> { ... 200 chars ... }` lines.
- **The generator is the source of truth but the build doesn't enforce it.** Drift is a question of when, not if.
- **The pattern is fundamentally a code-generation pattern fighting Java's lack of generics over primitives + value templates.** The right tool for this in 2026 is a build-time annotation processor or a real bytecode generator (ASM / `java.lang.classfile`). The PR's design doc explicitly considers this and rejects it ("stack traces point at real source") — but stack traces from a one-line lambda already don't point at real anything, so that argument is weak.

A direction I'd consider: **do the fusion at runtime via `MethodHandle`** instead of compile-time matrix expansion. `MethodHandles.guardWithTest` + primitive method handles let you compose `(load A) → (compare A) → (and|or) → (load B) → (compare B)` from a small number of building blocks, and `MethodHandle.invokeExact` through a constant `MethodHandle` field is monomorphic by JIT contract — same inlining behavior as the synthetic-lambda approach, without the matrix. Same JIT effect; one helper file instead of 1500 lambdas; new combinations cost zero source.

This is a bigger change than what's in flight, but it's the natural endpoint if the matrix keeps growing.

## 3. The "Future Work / Stage 5" reader specialization is the bigger lever

The doc lists three options for the residual outer-site megamorphism. Option 2 — generate a per-query `FilteredRowReader` subclass with the matcher inlined into `hasNext` — is the architecturally clean path and would subsume this PR. If you write that compiler, you don't need the fusion matrix because there's no megamorphic site left to defeat.

Doing reader specialization is a real project (bytecode generation, debuggability story, hot-load class management). But it's the design that *ends* the optimization curve, whereas Stage 4 fusion is a fixed % win that doesn't compound. If this team's next sprint will be Stage 5 anyway, Stage 4 may not be worth landing — it's a 1.35× that becomes 1.0× the moment Stage 5 ships.

If reader specialization is more than a quarter away, Stage 4 stands. If it's the next thing on the roadmap, I'd consider skipping fusion and going straight to specialization.

## 4. The documentation-as-process problem

`RECORD_FILTER_FUSION_PLAN.md` (1143 lines of agentic worker checklist) shouldn't be in `_designs/`. More broadly: this work has produced three design docs (compilation, fusion, fusion-plan) over a few PRs. Each PR adds a doc; the corpus grows monotonically. The `CLAUDE.md` rule about end-state-only is correct precisely because of this — without the rule, the design directory becomes a project diary. Recommend folding all four stages into one `RECORD_FILTER_COMPILATION.md` once the dust settles, with sections per stage for what each layer does and why, no per-stage history.

## 5. The `FUSION_ENABLED` flag is a code smell worth a design conversation

A static-final-at-class-init flag exists for one reason: benchmarks. It's a backdoor that future code will trip on (someone will write a test that depends on a specific compile path and break it under the other flag). Better options:

- **Make the compiler entry point take a `CompileOptions` record** that includes `boolean fusion`. Production passes the default (`true`); benchmarks pass `false`. No global state, no system property, no class-init order dependency.
- Or **delete the flag** and rely on Stage 1–3's existing JMH coverage as the "without fusion" baseline. If the team really wants A/B in one JVM, the options-record path is cleaner.

The system-property pattern is borrowed from JVM ergonomics, where it makes sense for runtime tuning. This isn't tuning; it's a benchmark switch.

## Bottom line on direction

The work is correct and the JMH evidence is strong. The questions worth thinking about before merging are:

1. Is the full 7×36×2×2 matrix worth the surface area, or would just same-column same-type AND capture the wins?
2. Is Stage 5 (reader specialization) on the near roadmap? If so, is Stage 4 worth the maintenance burden it adds before being subsumed?
3. Is hand-written-via-generator the right long-term shape, or is `MethodHandle`-based runtime composition (or full bytecode generation) the natural next move once the matrix grows?

These are scope/strategy calls, not technical objections. The implementation is solid. I'd land it if the answers are "yes, the matrix is worth it" and "Stage 5 isn't imminent" — and address the three documentation blockers either way.

core/src/main/java/dev/hardwood/internal/predicate/RecordFilterCompiler.java
gunnarmorling: Yes, agreed (discard the comment on this in the CLAUDE post below).

iifawzi:
> @iifawzi, so why are there no significant differences in the first table between "compiled" and "fused"?

It's because the JVM is already doing its job very well for all the cases in that test, in that file the compiler sees every shape exactly once, so they're all monomorphic sites where the compiled version is already inlining them, fusion finally is just the same more or less, not better

```
@Setup
public void setup() {
    schema = buildSchema();
    rows = buildRows(BATCH_SIZE, 42L);
    predicate = buildPredicate(shape);
    compiled = RecordFilterCompiler.compile(predicate, schema);
}
```

so there's only single predicate each time the JVM run 

On the other hand, for the megamorphic tests, each JVM run goes through 12 different shapes, simulating probably a real scenario where hardwood runs once, and there's a lot of different read operations being performed, in that case, the JVM would see a lot of different matchers being called at `a.test()` & `b.test()` so it will give up inlining when there's more then `3` receivers (a, b), here's fused methods benefits comes up given lambdas are already compiled, showing the difference shared in the second benchmark (1.67×). 

```
    @Setup
    public void setup() {
        schema = buildSchema();
        rows = buildRows(BATCH_SIZE, 42L);
        predicates = buildPredicates();
        fused = new RowMatcher[predicates.length];
        generic = new RowMatcher[predicates.length];
        for (int i = 0; i < predicates.length; i++) {
            fused[i] = RecordFilterCompiler.compile(predicates[i], schema);
            generic[i] = compileGeneric2(predicates[i], schema);
        }
    }
```

gunnarmorling:
Ah yes, that makes a lot of sense. Thanks for the explanation!

iifawzi:
> The matrix is the wrong unit of optimization

I agree here, and especially that it's just our opinion of what are the important operations that could be performed, honestly, I agree with you that's a lot of code to support only a couple of few operations that we're not sure if will ever be used. The point behind pushing it to implement this matrix of operations is that we have no clue what's the most used operations or behaviors of people that we should fusion, so I pushed for implementing or/and for all of the types, and excluded null. I simpler approach is just to go with AND for the main types if we want, and extend only when asked for/ needed).. or keep going with all that, and pay the maintainability cost. I'd choose the first option, instead of getting rid of the entire fused methods 

 > Hand-typed lambda cases are the wrong mechanism for the matrix
 
 That's interesting, I agree here as well, haven't played with method handling, but that would definitely be better, I initially rejected the option of the bytecode generation as I thought it might be overengineering if it can be written inline, but let's try the method handling approach, sounds interesting, i'd love to explore in that direction 
 
 >  The "Future Work / Stage 5" reader specialization is the bigger lever
 
 Interesting take, i don't know why i had a feeling that this stage 5 wouldn't optimize a lot we can try with claude and see after we take a decision related to stage 4, it should be simpler I think. 
 
 
 Finally the flag should be removed yes, and documentation need to follow the practices, it went messay because it was inspired from the playground that i had, so claude kept editing on it, and never followed the common usage of _design documents 
 
 so in my opinion: 
 1. We need to check the method handling approach, that could be a very good win when it comes to maintainability especially if it held same speedups 
 2. After that exploring optimizing the outer call site in stage 5 
 
 If method handling isn't optimizing a lot (i think it should be the same, but just in case) we can decrease the matrix to a minimum of AND only operations, will be around 36 lambdas only, or entirly drop the fusion optimization and stick with the compiled approach 
 
 What do you think? @gunnarmorling, if you agree with that, we can remove stage 4 from here, and stick to only the compiler optimization, until I can experiment with the previous two points/ other two issues you shared, if they will help in increasing e2e which's something i see important as well

gunnarmorling:
> We need to check the method handling approach, that could be a very good win when it comes to maintainability especially if it held same speedups

Yes, exactly. If you're for exploring the bytecode generation approach via method handles, that would be awesome. Ultimately, it would allow us to fuse any kind of query, so we'd completely sidestep the speculation about what's common and what's not. Two caveats worth mentioning:

* the "no mandatory dependencies" philosophy of Hardwood (but I think we can do it with method handles?)
* we may want to reduce the minimal Java version required to run Hardwood from 21 to 17, should it this prove to be beneficial to adoption; I'm not sure whether there capabilities only available in 21 which we'd need?

> If method handling isn't optimizing a lot (i think it should be the same, but just in case) we can decrease the matrix to a minimum of AND only operations, will be around 36 lambdas only, or entirly drop the fusion optimization and stick with the compiled approach

I don't think we should drop it entirely, it is beneficial after all; but we should narrow down the Lambdas we want to keep in that case, and I'd lean towards being conservative initially.

> if you agree with that, we can remove stage 4 from here, and stick to only the compiler optimization, until I can experiment with the previous two points/ other two issues you shared, if they will help in increasing e2e which's something i see important as well

I'm not quite following on this one; wouldn't removing stage 4 imply closing this PR? Or do you mean to not merge it for now and you keep exploring the issue?

iifawzi:
> the "no mandatory dependencies" philosophy of Hardwood (but I think we can do it with method handles?)
we may want to reduce the minimal Java version required to run Hardwood from 21 to 17, should it this prove to be beneficial to adoption; I'm not sure whether there capabilities only available in 21 which we'd need?

no clue, I need to explore that. if it's the case, then we can narrow down to only few lambdas that we can start with. 

> I'm not quite following on this one; wouldn't removing stage 4 imply closing this PR? Or do you mean to not merge it for now and you keep exploring the issue?

No, there's two stages applied here, the compiler itself, and then fused methods, and the compiled version is the one showing the biggest improvement. 

After stage 1,2 introduced here https://github.com/hardwood-hq/hardwood/pull/399, the matcher is still doing an internal loop, paying a dispatch overhead per matcher, so a 2-and operation would be compiled to the following today: 

```JAVA
    private static final class AndNMatcher implements RowMatcher {
        private final RowMatcher[] children;
        AndNMatcher(RowMatcher[] children) {
            this.children = children;
        }
        @Override
        public boolean test(StructAccessor row) {
            for (int i = 0; i < children.length; i++) {
                if (!children[i].test(row)) {
                    return false;
                }
            }
            return true;
        }
    }
```

so if we're doing `col1 > 50` AND `col2 > 20` there's still two virtual calls here to the Int predicate lambda method, and due to the loop this can't be optimized by the JVM as per my understanding. 

Stage 3 here gets rid of that loop with the fixed arity matchers (2, 3,4 Inputs), which's the reason behind the significant improvement shown in the able above under compiled column, given in that case the JVM can do inlining 


| Shape | Legacy ns/op | Compiled ns/op | Fused ns/op | Speedup |
|---|---:|---:|---:|---:|
| `single` (`id >= 0`) | 2.619 ± 0.025 | 0.507 ± 0.002 | **0.506 ± 0.006** | **5.18×** |
| `and2` | 10.005 ± 0.018 | 0.579 ± 0.015 | **0.583 ± 0.006** | **17.16×** |
| `and3` | 21.396 ± 0.179 | 0.563 ± 0.012 | **0.558 ± 0.013** | **38.34×** |
| `and4` | 30.802 ± 0.578 | 0.636 ± 0.011 | **0.638 ± 0.012** | **48.28×** |
| `or2` | 4.466 ± 0.007 | 0.507 ± 0.003 | **0.504 ± 0.024** | **8.86×** |
| `nested` (and-of-or) | 31.821 ± 0.191 | 0.934 ± 0.020 | **0.932 ± 0.069** | **34.14×** |
| `intIn5` | 3.696 ± 0.017 | 1.351 ± 0.021 | **1.390 ± 0.077** | **2.66×** |
| `intIn32` | 6.724 ± 0.091 | 3.085 ± 0.013 | **3.092 ± 0.015** | **2.17×** |

(Stage 1,2 numbers here: https://github.com/hardwood-hq/hardwood/pull/399#issuecomment-4356376485

so without stage 4, we are still keeping these speedups, thanks to this compiled version of the matchers, taking same example of above, the compiled matcher will be: 

```JAVA
    private static final class And2Matcher implements RowMatcher {
        private final RowMatcher a;
        private final RowMatcher b;
        And2Matcher(RowMatcher a, RowMatcher b) {
            this.a = a;
            this.b = b;
        }
        @Override
        public boolean test(StructAccessor row) {
            return a.test(row) && b.test(row);
        }
    }

```

this helps the JVM actually doing the inline-ing of the lambdas, giving us the maximum benefit compared to stage 1,2 where the loop is preventing it from doing that optimization. 

My suggestion, is to remove fused methods from this PR, and keep this compiler optimization, as this's a real huge win, and we can continue discovering fusion optimization with the method handling, or narrowing down the matrix as we discussed

gunnarmorling:
> My suggestion, is to remove fused methods from this PR, and keep this compiler optimization, as this's a real huge win, and we can continue discovering fusion optimization with the method handling, or narrowing down the matrix as we discussed

Gotcha, yes, that sounds good 👍!

iifawzi:
okies, now stage 4 is removed, and the results are as following: 


| Shape | Legacy ns/op | Main branch ns/op | Fixed-arity ns/op | Speedup vs legacy | Speedup vs main |
|---|---:|---:|---:|---:|---:|
| `single` (`id >= 0`) | 2.631 ± 0.019 | 0.513 ± 0.007 | **0.502 ± 0.018** | **5.24×** | **1.02×** |
| `and2` | 10.183 ± 0.259 | 2.039 ± 0.019 | **0.581 ± 0.007** | **17.53×** | **3.51×** |
| `and3` | 21.304 ± 0.130 | 8.223 ± 0.112 | **0.583 ± 0.045** | **36.54×** | **14.10×** |
| `and4` | 30.989 ± 0.654 | 13.424 ± 0.184 | **0.643 ± 0.015** | **48.19×** | **20.88×** |
| `or2` | 4.493 ± 0.044 | 0.955 ± 0.005 | **0.510 ± 0.005** | **8.81×** | **1.87×** |
| `nested` (and-of-or) | 33.574 ± 1.927 | 10.747 ± 0.286 | **0.948 ± 0.013** | **35.42×** | **11.34×** |
| `intIn5` | 3.714 ± 0.040 | 1.401 ± 0.025 | **1.354 ± 0.047** | **2.74×** | **1.03×** |
| `intIn32` | 6.827 ± 0.042 | 3.116 ± 0.036 | **3.071 ± 0.023** | **2.22×** | **1.01×** |

Legacy is main branch before this PR and the previous PR, then main branch is the main branch as of now (including previous PR), finaly the fixed arity is this branch.. let me know if you have any doubts, I know i might have overwhelmed you, thanks for your care following up on that

core/src/main/java/dev/hardwood/internal/predicate/RecordFilterCompiler.java
```diff
@@ -133,19 +130,78 @@ private static int indexedTopLevel(FileSchema schema, int columnIndex,
...
             IntUnaryOperator topLevelFieldIndex) {
         RowMatcher[] compiled = compileAll(children, schema, topLevelFieldIndex);
-        if (compiled.length == 1) {
-            return compiled[0];
+        return switch (compiled.length) {
+            case 1 -> compiled[0];
+            case 2 -> new Or2Matcher(compiled[0], compiled[1]);
+            case 3 -> new Or3Matcher(compiled[0], compiled[1], compiled[2]);
+            case 4 -> new Or4Matcher(compiled[0], compiled[1], compiled[2], compiled[3]);
+            default -> new OrNMatcher(compiled);
```
iifawzi, line 150: so for the cases of 1,2,3 or 4 matchers we will prefer the compiled fixed matchers, else, we fallback to the normal `AndNMatcher`, i think supporting up to 4 is enough, but it's easy to be extended anyway

gunnarmorling:
Squashed and applied. Thanks a lot for this work and also the fast iteration, @iifawzi! Looking forward to seeing the method handle experiments, whenever you get a chance.

iifawzi:
> Squashed and applied. Thanks a lot for this work and also the fast iteration, @iifawzi! Looking forward to seeing the method handle experiments, whenever you get a chance.

Thanks to you too for the care, the insightful discussions and quick resolutions, I'm willing to explore the method handle this weekend as well, I just took a quick look and it seems like it can still be used even if we lowered minimal version, let's see

gunnarmorling:
>  it can still be used even if we lowered minimal version

Sounds good! For transparency, I'm not sure whether we really want to go back to 17, but it would nice to leave that door open. Worst case, we always could push certain optimizations into a multi-release JAR path (we already do that for some other stuff).

PR #413, opened 2026-05-01
#78 Support geospatial logical types, statistics, and intersects push…

gunnarmorling:
…down

Add the metadata layer for the Parquet geospatial spec (GEOMETRY/GEOGRAPHY logical types, GeospatialStatistics with BoundingBox, page-level stats on ColumnIndex) plus an intersects() filter predicate that performs row-group and page-level pushdown on the stored bounding box. Per-row evaluation is intentionally coarse-grained — callers decode the WKB payload with their own geometry library on the surviving rows.

Without this, hardwood callers had no way to identify spatial columns or skip non-overlapping row groups for spatial queries. The pushdown is zero-runtime-dependency: bbox-vs-bbox arithmetic on doubles already present in the footer / column index. JTS is pulled in only at test scope to verify end-to-end WKB decoding.

Antimeridian wrapping (xmin > xmax for GEOGRAPHY) is handled in the overlap check. Negation of intersects throws UnsupportedOperationException — a correct dual would need different geometric criteria at the chunk level and WKB decoding at the row level, both out of scope for this change.

Refs the design document at _designs/GEOSPATIAL_SUPPORT.md.

## Checklist


gunnarmorling:
Applied; thanks a lot, @mghildiy! Also logged #414 as a follow-up to implement row-level filtering, via an optional dependency for parsing WKB bytes.

Issue #371, opened 2026-04-27
Support Non-OffsetIndex Cases for Per-Page Row Masking


rionmonster:
## Description
As a follow-up to #277 and the work proposed in #369. The per-page row mask machinery introduced there is only honored by `IndexedFetchPlan` (columns with an OffsetIndex). Files without Page Index (e.g. `parquet-cpp-arrow` defaults, older parquet-mr output) fall through `SequentialFetchPlan`, which doesn't apply masks. This leaves two gaps:

  1. **Latent correctness gap in mixed-OffsetIndex configurations.** When the filter column has an OffsetIndex but a projected non-filter column doesn't, the IndexedFetchPlan column applies the mask while the SequentialFetchPlan column
  emits unmasked rows from offset 0 — the same misalignment story as #277, just rarer. Today this is masked by writers being all-or-nothing on Page Index in practice, but it's a real correctness gap.
  2. **Tail-mode fast path is gated off** for any file lacking Page Index. `ParquetFileReader.canFastSkipTail` falls back to the existing decode-and-discard loop when any projected column lacks an OffsetIndex, leaving substantial performance on the table for parquet-cpp / older parquet-mr files.

## Current behavior
`SequentialFetchPlan` emits every page's full content; `RowRanges` are not consumed. Sibling `IndexedFetchPlan` columns under the same filter / tail-skip apply per-page masks, leading to misalignment in mixed configurations. Tail-mode degrades to decode-and-discard on any file without Page Index.

## Expected behavior
`SequentialFetchPlan` accepts `RowRanges` and applies per-page masks the same way `IndexedFetchPlan` does:

  - For each page, compute `pageFirstRow` from a cumulative cursor and `pageLastRow = pageFirstRow + recordsInPage`.
  - `mask = matchingRows.maskForPage(pageFirstRow, pageLastRow)`.
  - `mask == null` → skip the page entirely (advance past `compressedPageSize` without decompressing — the actual perf win on remote backends and the CPU win locally).
  - Otherwise → attach mask to `PageInfo`; the existing assembly machinery handles it.

### Implementation cost depends on column type and page format

  | Column type | Page format | Record-count source | Skip-without-decompress |
  |---|---|---|---|
  | Flat | v1 or v2 | `header.num_values` | yes |
  | Nested | v2 | uncompressed rep-level RLE prefix | yes |
  | Nested | v1 | rep-levels inside compressed area — must decompress to count records | no (perf win is gone) |

[...]

PR #415, opened 2026-05-01
#371 Extend Per-Page Row Masking to SequentialFetchPlan

rionmonster:
## Description
This pull request addresses #371 by extending the per-page row mask machinery (originally introduced in #277 for `IndexedFetchPlan`) to `SequentialFetchPlan`, so that files without a Page Index (and any tail-read into a row group's first kept page) no longer fall back to the decode-and-discard path. The reader now drops leading pages entirely and trims the straddling page in place, even when no Offset Index is available.

## Key Changes
The overall plan is documented under a newly created `SEQUENTIAL_FETCH_PLAN_PAGE_MASKING.md` document, but can be summarized at a high-level as follows:

- **Introduced Per-Page Masking in SequentialFetchPlan**
    - Added `matchingRows` (RowRanges) and `rowGroupRowCount` fields, plumbed through respective build overloads.
    - Iterator switched to a look-ahead pattern (cached nextPage) so `hasNext` cannot lie when pages are dropped mid-stream.
    - Trailing early-exit via `RowRanges.endRow()` triggered once the last matching row has been emitted, header scanning stops.
- **Added Mask-Eligibility Gate**
    - `RowGroupIterator.masksApplicableForRowGroup(...)` checks each projected column status (e.g., flat, has OffsetIndex, nested-v2) before allowing the mask through; otherwise targets all ranges.
    - New `PageFormatProbe` peeks the first data page header per column to distinguish v1 from v2 nested.
- **Added `PageRecordCounter` to Limit Allocations During Masking**
    - For nested columns, walks the RLE/bit-packing rep-level stream in place and counts `rep_level == 0` occurrences without materializing a buffer (saves on the order of hundreds of MB of transient allocation per scan on nested-v2 row groups).
- **Extended Range Primitives for `RowRanges`**
    - Added `RowRanges.range(start, end)` to express "skip the leading N rows" without flipping the all-rows sentinel.
    - Added `RowRanges.endRow()` to support the trailing early-exit.

## Verification and Testing
Test coverage was added and extends both the masking-machinery unit tests and the integration paths exercised end-to-end via:
- `PageRecordCounterTest` to verify behavior across various RLE/bit-packed inputs through both the new walker and RleBitPackingHybridDecoder.readInts + count-zeros, asserting equal output. Catches drift from the canonical decode path.
- `MisalignedPageBoundariesNoIndexTest` to eercise the no-Page-Index fast path with per-column page boundaries that diverge mid-row-group (same alignment issue introduced in the previous body of work, but without the Offset Index escape hatch).
- `NestedV2NoIndexMaskingTest` to verify the rep-level walker against actual nested data.
- `PageFormatProbeTest` to verify behavior for v1/v2 dispatch using existing fixtures.
- `RowRangesTest` to exercise newly added primitive behavior (e.g. `endRow()`, etc.)

## Performance and Throughput
As part of typical performance side testing, I ran a regression sweep against which demonstrated the following:

| Benchmark | main | Current | Δ |
|---|---|---|---|
| FlatPerformanceTest (taxi indexed) | 4450 MB/s | 4514 MB/s | +1.4% _(negligible)_ |
| PageFilter — unfiltered | 740M rec/s | 743M rec/s | +0.4% _(negligible)_ |
| PageFilter — filtered (id < 10%) | 406M rec/s (12.3 ms) | **560M rec/s (8.9 ms)** | **+38%** |
| RecordFilter — no filter | 547M rec/s | 554M rec/s | +1.2% |
| RecordFilter — match-all | 119M rec/s | 121M rec/s | +1.8% |
| RecordFilter — selective (id < 1%) | 27.5M rec/s (3.6 ms) | **31.6M rec/s (3.2 ms)** | **+15%** |

Additionally memory/allocations appeared to be unchanged.

## Checklist


rionmonster:
@gunnarmorling 

This is a follow-up PR for #277 to bring similar per-page masking to data against pages _without_ OffsetIndex so that the row skipping behavior will be at parity for both cases. No real signs of any regressions related to the effort with regards to performance and/or resource allocations based on a pretty exhaustive then/now comparison.

Feel free to take a look whenever you have a chance!

gunnarmorling:
Hey @rionmonster, here's the review feedback from Claude. Could you take a look and apply what makes sense? Thanks a lot!

_(remainder via Claude Code)_

Review of the current head. Solid PR — the fast-path implementation is correct and well-documented. The items below are what I'd want addressed before merge, in roughly decreasing priority.

### Blockers

1. **Nested-v1-without-OffsetIndex fallback path is implemented but untested.** The design doc lists a `nested_v1_no_index.parquet` fixture and `NestedV1NoIndexFallbackTest`; neither is in the diff. The gate-closed branch at `RowGroupIterator.java` (line ~550, `matchingRows = RowRanges.ALL`) and the `IllegalStateException` guard in `SequentialFetchPlan.computeRecordsInPage` are unreachable from today's tests. Either add the fixture (parquet-mr `WriterVersion.PARQUET_1_0`) and a fallback test, or scope the gate so v1-nested-without-OI is rejected at plan time and the dead branch is removed.
2. **Trailing-page early-exit perf claim is unasserted.** The early-exit in `SequentialPageIterator.findNextEmittablePage` is correct on inspection, but no test asserts the iterator stops scanning headers after the last kept page. Add a `pageCount` assertion via `RowGroupScannedEvent` against a multi-page no-index fixture with a leading-prefix filter — otherwise a future cursor-semantics refactor can silently regress the win with no test failing.

### Concerns worth fixing or justifying

3. **`masksApplicableForRowGroup` runs twice per row group on the tail-fast-path.** `ParquetFileReader.canFastSkipTail` calls the gate per row group in `subset`; iteration then calls it again per row group via `computeFetchPlans`. The design promises caching on `SharedRowGroupMetadata` ("Detection happens once per row group and is cached…") — not implemented. On hundred-row-group files with even one nested-without-OI column that's a hundred redundant `PageFormatProbe` GETs per tail query. Cache the result on `SharedRowGroupMetadata` (the design even names it `MaskCapability`).
4. **`computeRecordsInPage` allocates a `byte[]` per page just to copy out of the slice.** `readBytes` already returns a zero-copy slice; `.get(byte[])` then allocates and copies into a heap array, per page, per nested-v2 column. Overload `PageRecordCounter.countTopLevelRecords` on `ByteBuffer` — same byte indexing, no allocation. Otherwise the "saves hundreds of MB of transient allocation" claim is partly undone by these per-page copies.
5. **`recordsInPage > 0 ? recordsInPage : numValues` sentinel** in the inline-stats placeholder branch. `recordsInPage == 0` is overloaded ("not computed" vs "legitimately empty page"). Make it explicit — either a `boolean recordsInPageComputed` flag (the iterator already has `nextPageComputed` for the same idea), or compute `recordsInPage` unconditionally.
6. **`endRow()` returns `Long.MAX_VALUE` for both `RowRanges.ALL` and `RowRanges.all(N)`**. Today's caller gates on `!isAll()`, but the method is `public` and the next caller may not. Either make it package-private with a single-caller comment, or return the correct end of the all-range when `count >= 0`.
7. **`canFastSkipTail` silently reads `inputFiles.get(0)`.** The multi-file rejection upstream keeps this safe today, but if anyone later removes that rejection, the gate decision is silently wrong for non-first files. Either assert single-file at this point or thread the per-row-group input file through.

### Smaller items

- `PageFormatProbe`'s peek ceiling (1 MiB) and `SequentialFetchPlan`'s `MAX_PAGE_HEADER_PEEK_SIZE` are two constants for the same thing — consolidate.
- `bitWidthFor` is duplicated between production and `PageRecordCounterTest` — make the production helper package-private and `import static`.
- `findNextEmittablePage` is now ~80 lines — a one-line "header / mask / emit" comment would help future readers grok the loop body.

### Confirmed not issues

- The "single S3 download" model holds: page-header parsing and decode happen on the same `ChunkHandle` buffer; header walking is not a separate I/O pass.
- Cross-column placeholder alignment via `mask.totalRecords()` is correct.
- Tail-mode early-exit cannot fire spuriously: `RowRanges.range(skip, total)` makes `endRow == rowGroupRowCount`.
- No `var` usage; `///` JavaDoc throughout; CLAUDE.md style preserved.
- Public API surface is internal-only; no `docs/content/` update required.

### Verdict

Once 1, 2, and ideally 3 are addressed, I'd approve. The fast-path correctness story is solid; what's missing is coverage of the *fallback* path and the *early-exit perf* claim.

rionmonster:
@gunnarmorling 

Sounds good — I'll tag you once I work through those. I also had to fuss at my local Claude as its own review did a poor job not catching quite a few of these. 😄

rionmonster:
@gunnarmorling 

I've pushed a ton of commits this morning to work through all of the above feedback (both larger and smaller items).

gunnarmorling (approved):
Awesome work, @rionmonster! Loving how we avoid decoding pages for tail requests and how things become more regular at the reader level. I feel the fetch plan design is proving to work out nicely. Thanks a lot!

gunnarmorling:
> I also had to fuss at my local Claude as its own review did a poor job not catching quite a few of these.

Have you tried doing a review using a separate Claude session? This yields much better results for me most of the time.

gunnarmorling:
Squashed and applied.

Issue #103, opened 2026-03-02
Limit the set of performance tests run on CI


gunnarmorling:
We only really need `FlatPerformanceTest` and `NestedPerformanceTest`, and these also only with a subsets of contenders; e.g. we don't need to test parquet-java continuously. This should cut down CI times quite a bit; currently, the performance-test stage takes about 3 min.

brunoborges:
For projects where I like seeing benchmarks, I have a separate GitHub Action workflow just for that. Then you can decide if you want to trigger it during every push/pr merge and/or release cycle.

PR #417, opened 2026-05-02
#103 Limit / Extract Default Performance Test Workflow

rionmonster:
## Description
This pull request addresses #103, which observed that the per-PR performance stage was adding ~3 minutes to every check while producing noisy signal that rarely caught real regressions. The work is split across three commits that trim what runs, where it runs, and how it's parameterized.

## Key Changes
The changes are CI-only and no production code is touched. Changes primarily include:
- **Limit CI to Primary Workflows**
  - Restrict the performance testing stage to `FlatPerformanceTest` and `NestedPerformanceTest`. All other workloads are now run on-demand/ad-hoc.
- **Extracted Performance Tests to Dedicated Workflow**
  - Introduced a new `.github/workflows/performance.yml` triggered on `push: [main]` + `workflow_dispatch`.
  - Maintainers can now investigate a suspect PR can dispatch it against any branch from the Actions tab.

## Verification and Testing
Since this is a CI-only change with no application code touched, so verification is structural. 
```
(todo)
```

## Checklist

### DIFF
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -17,6 +17,25 @@ on:
   push:
     branches: [ main ]
   workflow_dispatch:
+    inputs:
+      perf_runs:
+        description: 'Number of timed runs per contender'
+        required: false
+        default: '5'
+        type: string
+      tests:
+        description: 'Surefire -Dtest filter; pass "*" to run every perf test'
+        required: false
+        default: 'FlatPerformanceTest,NestedPerformanceTest'
+        type: string
+
+# Single source of truth for the taxi-data window. The downloader (run via
+# the test-data-setup module on cache miss) and the perf test both honor
+# `perf.start` / `perf.end`; keeping the cache key aligned ensures we don't
+# silently widen the window when the cache is regenerated.
+env:
+  PERF_START: '2023-01'
+  PERF_END: '2025-12'
 
 jobs:
   performance-tests:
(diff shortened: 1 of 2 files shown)


rionmonster:
@gunnarmorling 

In the interest of still carving away at the current `1.0.0.CR1` milestone, I decided to tackle this one as well. It separates the performance tests out to a separate workflow (not currently running on every push) although one that could be run ad-hoc. Additionally, while thinking about it I'm going to carve a separate issue to define a new bot that we can _hopefully_ use to trigger ad-hoc performance testing runs via GitHub comments (e.g. `@hardwoodbot run perf`) that can reply with some throughput numbers similar to the flat/nested calls we usually compare against.

Thoughts on that as a QoL improvement or platform for future usage?

gunnarmorling:
Hey @rionmonster, nice, I think this is an improvement overall. Not significantly moving the needle in terms of wall clock time for PR builds, unfortunately, as the native CI build verification stage still adds 3 min (these two stages ran in parallel before this change). The capability to run the perf jobs manually is a good step forward on its own right though, so I'm in favor of merging this.

As for the bot, this sounds good too. I'm curious how that would relate to the previous idea of tracking results via [Otava](https://otava.apache.org/docs/overview/). One main question there would be how stable numbers from different GHA runs are, i.e. do VMs have the same specs, could they be impacted by noisy neighbours, etc. Definitely an interesting area to explore, we'll need a comprehensive strategy for that.

gunnarmorling (approved):

gunnarmorling:
Squashed and applied; thanks a lot, @rionmonster!

PR #418, opened 2026-05-03
#250 Drain-side record filtering for column-local AND predicates

iifawzi:
Hello, this PR is an initial draft for #250 to spin up the discussions, moving the filtration to the drains. It has shown a great e2e improvement as shown below, while also significantly slower on the micro benchmarks given they're single threaded so it loses its benefits, also for the same reason, I excluded single leaf filtration as it turned out slower than the compiled row version. 

There's a lot of follow ups, mainly to support OR, and ranges, and others to optimize this direction further with fusion (similar to what we need to do at #193) or null handling optimization. Matchers are a bit complex due to bit packing 

### e2e

| Scenario                                              | Compiled (ms) | Drain (ms) | Ratio       | Path             |
|-------------------------------------------------------|--------------:|-----------:|------------:|------------------|
| `and2`  long+double                       — match-all |        53.19  |    36.40   |  **1.46×**  | drain (2 cols)   |
| `and3`  long+double+int                   — match-all |        76.45  |    39.37   |  **1.94×**  | drain (3 cols)   |
| `and4`  long+double+int+bool              — ~50% pass |       114.21  |    22.56   |  **5.06×**  | drain (4 cols)   |
| `and2`  selective ~0.1%                   — sparse    |         3.20  |     2.77   |    1.16×    | drain (2 cols)   |
| `and2`  ~50% selectivity                  — half      |       107.12  |    45.45   |  **2.36×**  | drain (2 cols)   |
| `and2`  empty result                      — no match  |         0.48  |     0.34   |    1.41×    | drain (2 cols)   |
| `single`  one leaf                        — match-all |        48.99  |    47.95   |    1.02×    | fallback (gate)  |
| `or`  long ‖ double                       — half      |        99.93  |    98.75   |    1.01×    | fallback (or)    |
| `range`  id BETWEEN x AND y AND value < c — ~10% pass |         8.96  |     8.68   |    1.03×    | fallback (dup)   |
| `intIn5`  selective ~5%                   — sparse    |        66.92  |    66.00   |    1.01×    | fallback (gate)  |

### Micro Benchmarks

| Shape    | Compiled ns/op | DrainSide ns/op | Drain / Compiled | Notes                              |
|----------|---------------:|----------------:|-----------------:|------------------------------------|
| single   | 0.412          | gated → fallback | n/a              | Single-leaf gate trips.            |
| and2     | 0.479          | 3.311           | 6.9× slower      | 2 columns, lowest leaf count.      |
| and3     | 0.561          | 3.551           | 6.3× slower      | 3 columns.                         |
| and4     | 0.651          | 2.481           | 3.8× slower      | 4 columns — best per-leaf cost.    |
| or2      | 0.413          | gated → fallback | n/a              | OR not eligible.                   |
| nested   | 0.986          | gated → fallback | n/a              | OR-inside-AND not eligible.        |
| intIn5   | 1.364          | gated → fallback | n/a              | Single-fragment IN-list.           |
| intIn32  | 3.126          | gated → fallback | n/a              | Same.                              |

## Claude description 

Record-level filter predicates have been evaluated on the consumer thread, serially across leaves, since the row reader's introduction. The compiled per-row matcher (#193) is tight, but its cost grows linearly in leaves and can't use the per-column drain threads that already touch every value.

This change moves eligible predicates onto the drains: each column-local leaf compiles to a per-(type, op) BatchMatcher that runs against the just- filled batch on its drain thread, writing matches into a long[] on the Batch. FlatRowReader intersects the per-column masks word-wise and iterates surviving rows via Long.numberOfTrailingZeros — no per-row RowMatcher dispatch on the consumer.

Eligibility is gated by BatchFilterCompiler: top-level AND of column-local leaves with at least two distinct projected columns, supported (type, op) pairs only. Single-fragment, OR, same-column-duplicate, and intermediate paths fall back to FilteredRowReader unchanged. The gate exists because single-fragment drain pays its overhead with no parallelism payoff — v1 measurements showed that case lost to compiled.

End-to-end on a 10 M-row file the drain-side path beats compiled by 1.46× on a 2-column AND, 1.94× on 3-column, and 5.06× on 4-column ~50%-pass. Single-threaded JMH still favours compiled (BitSet.get blocks autovec, nextSetBit is per-survivor), but those gaps close once the drains run in parallel. See _designs/DRAIN_SIDE_RECORD_FILTERING.md for the full results and the follow-ups list.

## Checklist


iifawzi:
I'm not sure if it worth it @gunnarmorling, it looks promising but i'm concerned about the maintainability, the gain increase is obvious with more than two leaves comparisons ( > ~2x speedup), but for two leaves which i assume is more common, it's still < 2x

it could be that with the follow ups it will get better. 

Edit: i feel like i'm a bit biased to micro benchmarks improvements from last PRs, wishing all performance tuning must be 40x lol.. 2x is already great, so that's still a win e2e

gunnarmorling:
Hey @iifawzi, thanks a lot, gonna check out this PR in a bit.

In the meantime, I've just invited you as a committer to the project (you should have gotten an email). It means you could, for instance, review and merge PRs from @rionmonster or me (or others), if you're up for it. You can also merge your own PRs if we have general consensus and you've just sorted out some remaining lose ends from the prior discussion (no need for yet another review cycle then), or quick fixes not worth of much discussion in the first place. I'd only ask you to not merge larger changes without any prior conversation (so e.g. a PR like this shouldn't go in without review). No pressure, you don't have to accept the invite, do as you see fit :) In any case, thank you so much for all you efforts and contributions around the project, I really appreciate it!

rionmonster:
@iifawzi 

Really great work! I've been following several of the discussions and threads across the projects and you've provided some really awesome contributions.

> Edit: i feel like i'm a bit biased to micro benchmarks improvements from last PRs, wishing all performance tuning must be 40x lol.. 2x is already great, so that's still a win e2e

So, so very relatable 😄

iifawzi:
Hi @gunnarmorling @rionmonster thanks for the trust, truly delighted to join, I've been enjoying being part of that, and willing to contribute more. 

I'm still building intuition and understanding of several components, and as we're moving so fast, it would take me sometime to be sure i'm up to date with high level architecture/direction, so until then, I will use the privileges with cautious when it comes to merge, but I will start keeping an eye on the PRs to help review, and to have clarity as well. (feel free to request review from me, when they land, if i missed). 

> So, so very relatable 😄 

Haha yes! a bit annoying it's not 40x :d. However I've been thinking about it, and I'm thinking that in order to have accurate estimations about whether going drain side is useful or not, I think we should be completing what we started @ #193 for consumer megamorphic site. When that optimization land, we would then be able to measure differences in latency compared to the fully optimized consumer side path

iifawzi:
Following up on previous comment, I also don't want to get slipped in these optimizations while leaving other core features/fixes, so curious about your thoughts here @gunnarmorling @rionmonster. Whether you think we should invest more or not, and whether it should be drain or consumer side, as a beginning

gunnarmorling:
Thanks a lot for this, @iifawzi, still processing that change, it's huge :)

> Hi @gunnarmorling @rionmonster thanks for the trust, truly delighted to join, I've been enjoying being part of that, and willing to contribute more.
>
> I'm still building intuition and understanding of several components, and as we're moving so fast, it would take me sometime to be sure i'm up to date with high level architecture/direction, so until then, I will use the privileges with cautious when it comes to merge, but I will start keeping an eye on the PRs to help review, and to have clarity as well. (feel free to request review from me, when they land, if i missed).

That's awesome! Welcome to the crew!

As for the change at hand; I'd like to better understand why are we not applying it generally, e.g. not for queries with a single column filter? Intuitively, I'd have expected every bit of computation we can move away from the consumer thread would be beneficial?

> Whether you think we should invest more or not, and whether it should be drain or consumer side, as a beginning

Very good question, definitely appreciate the thought. I feel this change probably needs more baking time, in particular considering we want to move to CR this week or next week and to 1.0 Final in the week of May 20th (week of Current London). So we probably shouldn't rush to get this in, if we're not feeling really confident. The maintenance effort definitely is a concern I share. That said, there might be an opportunity to apply some easy wins, like moving from `BitSet` to `long[]` for null representation, which might yield some perf gains with zero risk?

Taking a step back, other than improvements to performance, I think the following things are up high really high on the prio list:

- Do a review pass on the public API before 1.0; is this what we want to ship?
- Bloom filters (not needed for 1.0 necessarily, but should come shortly thereafter)
- Develop a comprehensive performance benchmark comparing to parquet-java for representative workloads
- Write support, which will be the key theme for 1.1; it's gonna be the single biggest blocker stopping many potential users from adopting Hardwood; if they need both read and write support, they won't move to Hardwood if it only supports the former, no matter how light-weight and fast it is

That said, I also love those performance improvements and people should work on what they enjoy working on. Purely from a project perspective, these things above (and maybe some others) are probably the most valuable ones right now.

iifawzi:
> As for the change at hand; I'd like to better understand why are we not applying it generally, e.g. not for queries with a single column filter? Intuitively, I'd have expected every bit of computation we can move away from the consumer thread would be beneficial?

Yes, I was expecting the same, benchmark didn't align, I believe it's due to the overhead of applying the filters on drain and then do intersection on the consumer side, I need to think about how this could be improved, and to identify exactly the reason of the latency increase for queries with single column filter, I will give it another look 

on the other hand, moving bitSet to long[] with bitpacking for null representation has been commented as an improvement by Claude to improve this PR, I started it as a separate commit against main couple of days ago, given it will be beneficial in general, I will refine it and let land shortly  


> That said, I also love those performance improvements and people should work on what they enjoy working on. Purely from a project perspective, these things above (and maybe some others) are probably the most valuable ones right now.

Yes, I totally agree, and prio list seems reasonable especially for what's expected before 1.0 final so such performance improvements are still be cool to tackle. 

That being said, I will consider this week a week of perf tuning 

EDIT: Drafting this PR as I'm planning to dig more into this in the upcoming days

core/src/main/java/dev/hardwood/internal/predicate/RecordFilterCompiler.java
```diff
@@ -401,55 +400,36 @@ static int compareBinary(byte[] left, byte[] right, boolean signed) {
...
     private static RowMatcher intInLeaf(String[] path, String name, int[] values) {
-        int[] sorted = values.clone();
-        Arrays.sort(sorted);
-        if (sorted.length >= IN_LIST_BINARY_SEARCH_THRESHOLD) {
-            return row -> {
-                StructAccessor a = resolve(row, path);
-                if (a == null || a.isNull(name)) return false;
-                return Arrays.binarySearch(sorted, a.getInt(name)) >= 0;
-            };
-        }
```
iifawzi, line 412: binary search has slipped from previous PR adding the compiler, it's unrealistic that a user will pass 16 value in the values array for In predicate, to simplify the code, I'm removing it

core/src/main/java/dev/hardwood/internal/reader/FlatRowReader.java
```diff
@@ -176,7 +280,74 @@ public boolean hasNext() {
...
+                return Math.min(bit, limit);
+            }
+        }
+        return limit;
+    }
+
+    /// Finds the next set bit in `words` at or above `from`, bounded above by `limit` (exclusive).
+    /// Returns `-1` if no such bit exists. Used by the drain-side iteration path so
+    /// `hasNext()`/`next()` stay monomorphic without a wrapping reader.
+    private static int nextSetBit(long[] words, int from, int limit) {
```
iifawzi, line 327: Hello @gunnarmorling @rionmonster I've applied several iterations, it's a bit up to par now. 

Coming back to the question of why it was not performing very well for single leaf filter (match all id < 0) it was due to the overhead of determining the next row , due to the logic of `nextSetBit` here, and the intersection logic (little overhead, but improved as well)

To improve that, I added a fast path to avoid going through `nextSetBit`, if there's a long sequence of 1s [(Here you can find the change)](https://github.com/hardwood-hq/hardwood/pull/418/changes/88270d3e6b8bc5a7bd0ac6583d5331f3b30fda51).  

This way, after getting the first set bit, we will run a scan to determine `runEndExclusive`, which will then be used to advance to next matches, skipping `nextSetBit` overhead. 

```JAVA
      if (rowIndex + 1 < runEndExclusive) {
          pendingRowIndex = rowIndex + 1;
          return true;
       }
```

It turned out to be very useful as well for long sequences of matches in general (sorted columns), I added a test `Sorted cluster` to verify that. 

Now Numbers are showing clear win on all the paths that go through the drain, aligning with our expectations

```
./mvnw test -Pperformance-test -pl performance-testing/end-to-end -Dtest="RecordFilterBenchmarkTest" -Dperf.runs=10
```

| Contender                                | Path (drain on) | Drain ON (ms) | Drain OFF (ms) | Speedup |
|------------------------------------------|-----------------|--------------:|---------------:|--------:|
| No filter (baseline)                     | —               |          19.7 |           19.2 |   0.97× |
| Match-all (id≥0)                         | Drain           |          23.6 |           27.7 |   1.17× |
| Selective (id<1%)                        | Drain           |           1.8 |            2.0 |   1.11× |
| Compound match-all (id≥0 & value<+∞)     | Drain           |          23.2 |           50.0 |   2.16× |
| Page+record (top 1% & value<500)         | Consumer        |           3.9 |            2.8 |   0.72× |
| And3 match-all (id+value+tag)            | Drain           |          22.6 |           63.2 |   2.80× |
| And4 ~50% (id+value+tag+!flag)           | Drain           |          36.9 |          104.9 |   2.84× |
| Compound selective (id<10K & value<+∞)   | Drain           |           0.8 |            1.1 |   1.38× |
| Compound mid 50% (id≥0 & value<500)      | Drain           |          43.5 |          100.8 |   2.32× |
| Sorted cluster 50% (id<N/2 & tag≥0)      | Drain           |          12.6 |           29.8 |   2.37× |
| Empty result (id<0 & value<+∞)           | Drain           |           0.3 |            0.4 |   1.33× |
| OR (id<0 or value<500)           | Consumer        |          92.1 |           90.0 |   0.98× |
| Range+value (id BETWEEN 1M..2M)          | Consumer        |           7.9 |            8.6 |   1.09× |
| intIn (tag IN [1,5,10,25,50])            | Drain           |          24.1 |           79.1 |   3.28× |


When this land, I will follow up on it with the following easy wins, but to keep the PR small (sorry for the 4k :p, better to do them in separate PRs, I will try in general to avoid such large PRs in the future). 

1. More than one predicate on same leaf - Now if there's multiple filtration on same leaf, it goes to the row filtered wrapper. 
2. Drain based OR Predicates, this PR only supports AND, I believe OR would also be a significant win. 
3. SIMD for the binary operations
4. and I have the null bitSet optimization on the radar as well

I will create issues later for them before creating the PRs. 

Let me know your thoughts!

iifawzi: to run benchmarks with drain stopped, you can `git reset --hard HEAD~1` and 
```
./mvnw test -Pperformance-test -pl performance-testing/end-to-end -Dtest="RecordFilterBenchmarkTest" -Dperf.runs=10 -Dhardwood.drainSide.enabled=false
```

as i removed the flag in last commit

rionmonster: @iifawzi 

Thanks for these changes and keeping after it on a performance front (🤞 on those elusive 40x gains)! I’ll try and pull these down and dig a bit deeper tonight or when I get some free time!

Much appreciated!

gunnarmorling:
@iifawzi, hey, can you remind me, where are we with this one? Do you want to extract that bitset -> long change, or do you think it's good to go altogether in its current shape?

iifawzi:
> @iifawzi, hey, can you remind me, where are we with this one? Do you want to extract that bitset -> long change, or do you think it's good to go altogether in its current shape?

I'm willing to tackle https://github.com/hardwood-hq/hardwood/issues/440 this week, if it landed before this one, I will rebase and do the changes, if this one landed first, the bit checks part will be changed in 440.

i'd say that this can land in its current form, when approved, to avoid the diff getting larger

gunnarmorling:
Hey @iifawzi, here are some review comments from Claude; can you apply what makes sense?

---

# PR #418 — Drain-side record filtering — Review actions

<https://github.com/hardwood-hq/hardwood/pull/418>

## Blockers for leaving draft

- [x] Extend `DrainSideOracleTest` to cover every supported `(type, op)` pair, not just `Long`/`Double`. Missing: `int`, `float`, `boolean × {EQ, NOT_EQ}`, `IntIn`, `LongIn`, `IsNull`, `IsNotNull`. The design doc claim "for every supported (type, op) pair … return the same surviving row indices" is not actually backed today.
- [x] Reconcile gate wording with the implementation. Design doc says "≥ 2 distinct projected columns"; code is `leaves.size() < 2`. Same-column duplicates pass the leaf-count gate and are caught later by the `result[projected] != null` check. Pick one and align doc + code.
- [x] Fix `BatchFilterCompiler` class JavaDoc — it lists "a single column-local leaf with a supported (type, op)" as eligible, but the gate rejects single-leaf queries.

## Correctness / fail-early

- [x] Throw or assert in `FlatRowReader.next()` on the drain path when `pendingRowIndex < 0` (i.e. `next()` called without prior `hasNext()`). Currently silently sets `rowIndex = -1`, which then poisons typed-array accessors.
- [x] Remove the unreachable `default -> null` arms in `BatchFilterCompiler.compileLeaf` (`BooleanPredicate` switch and outer switch). Either rely on exhaustive switch, or replace with `throw new IllegalStateException(...)` per fail-early.
- [x] Confirm or enforce that `ResolvedPredicate.And` is always flattened at resolution time. If not, either flatten recursively in `BatchFilterCompiler.flattenAnd` or assert in `tryCompile` so the assumption fails loudly.

## Concurrency / API shape

- [x] Make `FlatColumnWorker`'s drain-side filter constructor-injected and `final`. Removes the unenforced "set before `start()`" contract and the JMM concern about cross-thread visibility of a plain field. One caller to migrate (`FlatRowReader.create`).
- [x] Drop `BatchMatcher.columnIndex()` from the interface (and the field from every implementation's constructor) if nothing in production paths consumes it — `FlatRowReader.create` already knows the projected index it's iterating.

## Duplication / structural

- [ ] A/B-test a shared scalar helper (`IntPredicate`-style comparison passed in) against the 28 hand-written per-`(type, op)` matchers before this pattern locks in for another 56 files (`notNullWords` + Vector API matchers). JIT specializes on call-site type, not body shape — shared helper may not regress.
- [x] Consolidate `BINARY_SEARCH_THRESHOLD = 16` (currently duplicated in `IntInBatchMatcher`, `LongInBatchMatcher`, and `RecordFilterCompiler`) into one shared constant in `internal.predicate`.

## Compiler micro-issues

- [x] In `BatchFilterCompiler.tryCompile`, fold the same-column-duplicate check into the first validation pass so the rejection path doesn't waste the `result[]` allocation. A `long` bitmask of seen projected indices is enough for typical projection widths; `BitSet` for wide schemas.
- [x] Drop the redundant `pendingRowIndex = -1` reset inside `loadNextBatch` — `hasNext()` overwrites it via `nextSetBit` before any read.

## Performance follow-ups (defer if out of v1 scope)

- [x] Add early-exit to `intersectMatches` when `combined` becomes all-zero after a column merge — avoids running remaining columns on a known-empty batch when one leaf is highly selective.

## Test additions

- [x] Add `BatchFilterCompilerTest.IneligibleShapes` cases for: (a) `And(LongPredicate)` of size 1 — exercises the leaf-count gate via the `And` branch; (b) `And` containing a `BinaryPredicate` child — `Binary` is currently only tested as a standalone leaf.

## Documentation polish

- [x] Soften design doc opener ("Beats the compiled per-row path on every multi-column scenario measured") to carry the parallelism caveat up front — JMH single-threaded numbers show drain 3.8–6.9× slower; the honesty appears later in the doc but the lede should set expectations.
- [x] Add a one-line comment in `RecordFilterScenariosBenchmarkTest` explaining that `withRowGroupSize((long) TOTAL_ROWS * 16)` is a byte size intentionally forcing one row group, not a row count.
- [x] Update `ROADMAP.md` §9.4 with the drain-side entry the design doc says belongs there.

iifawzi:
Hey @gunnarmorling, I've applied all the changes. Some thoughts on the following points:

> Consolidate BINARY_SEARCH_THRESHOLD = 16 (currently duplicated in IntInBatchMatcher, LongInBatchMatcher, and RecordFilterCompiler) into one shared constant in internal. predicate.

I believe you haven't pulled latest commits before Claude review, I deleted earlier, all binary search operations, It's something that have slipped from the earlier RecordFilterCompiler, and I don't think is realistic or needed, having more than 16 column in an `In` predicate doesn't sound realistic, and doing binary search for 3 ~ 4 ~ 6 columns would be an overkill, so I removed the binary search from the record filter compiler, and from this PR, earlier. 

> In BatchFilterCompiler.tryCompile, fold the same-column-duplicate check into the first validation pass so the rejection path doesn't waste the result[] allocation. A long bitmask of seen projected indices is enough for typical projection widths; BitSet for wide schemas.

I added support for the same column filters @ https://github.com/hardwood-hq/hardwood/issues/250. However, there's room for improvement here, which I filed as an enhancement https://github.com/hardwood-hq/hardwood/issues/454


> A/B-test a shared scalar helper (IntPredicate-style comparison passed in) against the 28 hand-written per-(type, op) matchers before this pattern locks in for another 56 files (notNullWords + Vector API matchers). JIT specializes on call-site type, not body shape — shared helper may not regress.

I tested this with a single class for IntBatchMatcher, receiving the operation in constructor, so it's a single class instead of 6 (eq, not-eq, etc), and it turned 4x slower, It's because the inner site call (`operation.test()`) gets profiled as a megamorphic given operation has different shapes (eq, not-eq, lgt, etc..) 



Finally, I added https://github.com/hardwood-hq/hardwood/issues/456 for follow-up/discovery of SIMD instructions. and https://github.com/hardwood-hq/hardwood/issues/457 for OR support follow-up

gunnarmorling:
Awesome improvement, thanks a lot for tackling this, @iifawzi! Neat with that "consecutive matches" fast path, too.  Really cool to see this land! Pushed one more commit for a few minor tweaks.

gunnarmorling:
Squashed and applied. Thx, @iifawzi!

PR #420, opened 2026-05-04
#408 display version info (tag + commit sha) in the cli

### LINKED ISSUE #408: Display version info (tag + commit SHA) in the CLI
## Problem
The `hardwood` CLI currently exposes only the Maven `project.version` via `--version` (see [`HardwoodCommand.java`](cli/src/main/java/dev/hardwood/cli/command/HardwoodCommand.java) `VersionProviderWithConfigProvider`). That makes it hard to pinpoint which exact build a user is running when they report an issue, especially for `-SNAPSHOT` builds or native binaries cut from arbitrary commits.

There is also no way to see version information from inside the `hardwood dive` TUI at all.

## Expected
- `hardwood --version` prints both the release tag/version and the git commit SHA (e.g. `hardwood 0.5.0-Beta2 (abc1234)`); for dirty builds, mark them as such.
- The same info is surfaced in the `hardwood dive` TUI's help modal, so users can read it off without exiting the TUI.
- The commit SHA must be baked in at build time (e.g. via a generated resource or a Maven filtered property), so it works for both the JVM jar and the native binary.

mohamedibrahim54:
closes #408 
Display version info (tag + commit SHA) in the CLI via --version and display version information from inside the hardwood dive TUI help model

> hardwood --version
hardwood 1.0.0-SNAPSHOT (fe218a0)

when there are changes not commit yet, `-dirty` is added to sha
> hardwood 1.0.0-SNAPSHOT (a093aab-dirty)

#### TUI help model
Screenshot from 2026-05-04 06-19-12

also, Inject Project info into the Manifest for official record-keeping.

## Checklist

- [x] The commit message is prefixed with the issue key ("#408")

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -11,6 +11,7 @@ Run ./mvnw verify to build the project.
 When doing changes in hardwood-core, install that module before running the performance tests or any other module.
 When running Maven commands, always apply a timeout of 180 seconds to detect deadlocks early on.
 Enable -Pperformance-test to run performance tests.
+All plugin versions must be declared in the parent `pom.xml`'s `<pluginManagement>`. Module POMs reference plugins by `groupId`/`artifactId` only, never with a `<version>` of their own.
 
 # Design
 
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -567,7 +567,6 @@
       <plugin>
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>buildnumber-maven-plugin</artifactId>
-        <version>3.3.0</version>
         <executions>
           <execution>
             <phase>initialize</phase>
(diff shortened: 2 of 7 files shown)


gunnarmorling:
Thanks a lot, @mohamedibrahim54! Pushed one more commit for a few minor fixes like pulling up all plug-in versions to the parent.

gunnarmorling:
Rebased and applied. Thanks a lot, @mohamedibrahim54! Note I went with calling git via the antrun plug-in eventually, so as to also correctly capture the tainted state in the case where there are new files in the workspace (but not modified ones), which wasn't exposed by the plug-in.

mohamedibrahim54:
> Rebased and applied. Thanks a lot, @mohamedibrahim54! Note I went with calling git via the antrun plug-in eventually, so as to also correctly capture the tainted state in the case where there are new files in the workspace (but not modified ones), which wasn't exposed by the plug-in.

You're very welcome! That makes total sense about using the antrun plug-in to catch the untracked files. Thanks for reviewing and getting this applied!

PR #421, opened 2026-05-04
#84 Compare Nested Columns in ParquetComparisonTest

### LINKED ISSUE #84: Compare nested columns in column-level ParquetComparisonTest
The column-level comparison in \`ParquetComparisonTest\` (\`compareColumnsWithReference\`) skips 12 files that contain nested/repeated columns. Columns with \`maxRepetitionLevel > 0\` are also skipped within files that do run.

This means the ColumnReader API's handling of offsets, level nulls, and element nulls for list, map, and repeated columns is not validated against parquet-java.

The test should be extended to reconstruct nested values from the ColumnReader's offset and null bitmap arrays and compare them against the reference.

rionmonster:
## Description
This pull request closes #84, which observed that `compareColumnsWithReference` skipped 12 nested-column files via `COLUMN_SKIPPED_FILES` and bypassed every `maxRepetitionLevel > 0` column within files that did run. As a result,
  the public `ColumnReader` API's handling of offsets, level-null bitmaps, and element-null bitmaps for list / map / repeated columns was never validated against parquet-java.

## Key Changes
The primary work centers on a new offsets-driven reconstructor in `Utils.java` that produces a per-record nested value from the public `ColumnReader` surface and matches it against the same shape extracted from parquet-java's Avro `GenericRecord` along the column's `FieldPath`. Both single-file and multi-file column-comparison paths now flow through this reconstructor instead of skipping nested columns.

In summary form, changes include:
- **Added Nested Column Reconstruction in `Utils.java`**
    - `compareNestedColumnReader` iterates batches and produces per-record values from `ColumnReader`'s offsets / level-nulls / element-nulls.
    - `reconstructNested` recursively walks `getOffsets(level)` and `getLevelNulls(level)` to build a `List<Object>` shape, dropping to `getElementNulls()` and the typed value array at the leaf.
    - `extractAvroAlongPath` walks the parquet-java `GenericRecord` along the column's `FieldPath`, eliding structural names (`list`, `element`, `key_value`) by treating any component that isn't an Avro field on the current record as
   transparent. Map keys/values project by name; `List` components distribute the rest of the path per element.
    - `deepCompareNested` matches the two reconstructions structurally, reusing the same `String`/`byte[]` coercion and numeric tolerance as `compareColumnValue`.
- **Wired Through Single-File and Multi-File Test Paths**
    - `compareColumns` (single-file via `compareColumnsParquetFile`) dispatches `maxRepetitionLevel > 0` columns to the new nested path.
    - `multiFileColumnReadersMatchReference` drops its identical `maxRepetitionLevel > 0` skip.
- **Shrunk `COLUMN_SKIPPED_FILES` from 12 to 3**
    - 9 of the 12 files round-trip through the new reconstructor. Of those, four (`map_no_value`, `nested_maps.snappy`, `repeated_no_annotation`, `repeated_primitive_no_list`) were already gated upstream by `SKIPPED_FILES`, so the
  column test still skips them at the outer level for the same parquet-java reasons.
    - The remaining three (`null_list`, `nullable.impala`, `nonnullable.impala`) stay skipped because they contain empty repeated groups: an empty list and a list with one null element produce identical offset / null shapes.
  Distinguishing them requires the underlying definition level, which the public API doesn't expose. The skip list now documents this inline.
- **Added Per-Column Skip for Variant `typed_value` Subtrees**
    - `isUnderVariantShred` walks the schema along the column's `FieldPath` and returns true when a `typed_value` step occurs after a Variant-annotated ancestor.
    - parquet-java's `AvroParquetReader` collapses shredded variants onto the canonical `var.value` binary, so the reference side has no data for shredded subcolumns. The skip is per-column rather than per-file, so each variant's
  `metadata` and canonical `value` columns are still validated.

## Verification and Testing
As expected, the complete `parquet-testing-runner` module suite of tests passes with the new reconstructor running on all previously-skipped nested files:

```
[INFO] Tests run: 567, Failures: 0, Errors: 0, Skipped: 64
[INFO] BUILD SUCCESS
```

## Checklist

### DIFF
--- a/parquet-testing-runner/src/test/java/dev/hardwood/testing/Utils.java
+++ b/parquet-testing-runner/src/test/java/dev/hardwood/testing/Utils.java
@@ -555,22 +555,16 @@ public class Utils {
 
     // ==================== Column-Level Comparison ====================
 
-    /// Additional files to skip in column-level comparison tests.
-    /// Files with nested/repeated columns where column-level comparison
-    /// requires list reconstruction from offsets (deferred).
+    /// Additional files to skip in column-level comparison tests beyond [#SKIPPED_FILES].
+    /// All listed files contain empty repeated groups: the public [ColumnReader] API
+    /// exposes per-record offsets, level-null bitmaps, and element-null bitmaps but
+    /// not definition levels, which are needed to distinguish an empty list (size 0,
+    /// no elements) from a list containing a single null element (size 1). Both
+    /// produce the same offset/null shape, so reconstruction can't tell them apart.
     static final Set<String> COLUMN_SKIPPED_FILES = Set.of(
-            "list_columns.parquet",
-            "nested_lists.snappy.parquet",
-            "nested_maps.snappy.parquet",
-            "nested_structs.rust.parquet",
-            "nonnullable.impala.parquet",
-            "nullable.impala.parquet",
             "null_list.parquet",
-            "old_list_structure.parquet",
-            "repeated_no_annotation.parquet",
-            "repeated_primitive_no_list.parquet",
-            "incorrect_map_schema.parquet",
-            "map_no_value.parquet"
+            "nullable.impala.parquet",
+            "nonnullable.impala.parquet"
     );
 
     /// Compare a Parquet file column-by-column using ColumnReaders against parquet-java reference data.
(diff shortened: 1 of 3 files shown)


rionmonster:
@gunnarmorling 

Here's another towards the upcoming `1.0.0.CR1` milestone that has been around for a while. This was the initial pass at it, however a review towards the end mentioned the following as a follow-up which may/may not be worth including in this (since it has a larger blast-radius and affects the public ColumnReader API and not just tests/utilities).

It basically just involves exposing or adding the ability to distinguish between empty lists and a list with a single null element. I would let a consumer reliably reconstruct the user-visible shape (e.g., aggregations (array_length, cardinality, count-of-elements), equality checks, etc.) and any re-emit of the data all give different answers for [] vs [null], and without the disambiguator a columnar consumer has no way to produce the right one.

Thoughts on that? I wanted to mention it explicitly since it would touch the public API (and not just the tests like this PR proposes)

## Claude's Summary

**Problem**
The public ColumnReader API can't distinguish an empty list from a list with one null element — both encode identically in offsets / levelNulls / elementNulls, and the disambiguating definition level is computed internally but never exposed.

A nested container has four meaningful states:

  1. Null list — the container itself is null.
  2. Empty list — the container exists but has zero elements.
  3. List with null element(s) — the container has elements, some/all null.
  4. List with value(s) — the container has elements with non-null values.

The current API distinguishes three: levelNulls isolates state 1, elementNulls isolates state 4 from {2, 3}. States 2 and 3 collapse — both encode as a single phantom entry with elementNulls[i] = true. Only the underlying definition level separates them (def < elementDefLevel ⇒ empty), and NestedBatchIndex already uses exactly that in PqListImpl.computeRange. The internal reader gets it right; the public surface is just missing the bit.

**Why it matters**
Without the disambiguator, a columnar consumer cannot faithfully reconstruct the user-visible shape: array_length / cardinality, equality checks, and any re-emit of the data give different answers for [] vs [null]. There's also a completeness gap relative to flat columns — a flat consumer can fully reconstruct values from (values, elementNulls), whereas a nested consumer with (values, offsets, levelNulls, elementNulls) currently cannot.

**Proposed solution**
Surface the missing signal. Two reasonable shapes:
  - Derived (preferred): add BitSet getEmptyListMarkers(int level) parallel to getLevelNulls(int level). Matches the existing style of pre-computing bitmaps in the drain thread; cost is essentially free since
  NestedLevelComputer.computeLevelNulls already walks the same def-level array.
  - Raw (fallback): add int[] getDefinitionLevels(), mirroring the typed-value array accessors. Most flexible, but pushes interpretation onto the caller.

The data already flows — NestedBatch.definitionLevels is populated by the drain thread and computeLevelNullThresholds returns the per-level thresholds — so the change is additive and small (~1 day of work end-to-end including tests and the docs/content/usage.md update).

**Impact**
Unblocks the three files currently skipped in #84's COLUMN_SKIPPED_FILES (null_list.parquet, nullable.impala.parquet, nonnullable.impala.parquet) — all three exist specifically to exercise empty-repeated-group semantics. More broadly, closes a real ergonomic gap for any downstream consumer reconstructing nested values from the columnar API.


_(Update: I've gone ahead and spun this off as an issue in #422)_

parquet-testing-runner/src/test/java/dev/hardwood/testing/Utils.java
```diff
@@ -555,42 +557,72 @@ static Object convertToComparable(Object value) {
...
-            "nested_maps.snappy.parquet",
-            "nested_structs.rust.parquet",
-            "nonnullable.impala.parquet",
-            "nullable.impala.parquet",
             "null_list.parquet",
-            "old_list_structure.parquet",
-            "repeated_no_annotation.parquet",
-            "repeated_primitive_no_list.parquet",
-            "incorrect_map_schema.parquet",
-            "map_no_value.parquet"
```
gunnarmorling, line 573: Really sweet to see this list shrinking!

gunnarmorling:
> Thoughts on that? I wanted to mention it explicitly since it would touch the public API (and not just the tests like this PR proposes)

Yes, great catch! Definitely worth fixing for the 1.0 release (either via this week's CR1, or a CR2 next week).

gunnarmorling:
LGTM overall! I've pushed one more commit to clean up a few things, most notably for getting rid of the single-file / multi-file reader differentation which still was a left-over after the reader unification.

PR #426, opened 2026-05-05
#422 Distinguish Empty / Single Null Lists in ColumnReader API

### LINKED ISSUE #422: Distinguish Empty Lists / Single-Null Lists in ColumnReader API
## Problem
The public `ColumnReader` API exposes `offsets`, `levelNulls`, and `elementNulls` for nested columns, but it can't disambiguate an empty list from a list with one null element — both encode identically in those three arrays. This stems as a follow-up from the effort for #84 and was identified as a future improvement to support consumer disambiguity for these types of data. 

A nested container has four meaningful states:

  1. **Null list** — the container itself is null.
  2. **Empty list** — the container exists but has zero elements.
  3. **List with null element(s)** — the container has elements, some/all null.
  4. **List with value(s)** — the container has elements with non-null values.

The current API distinguishes three: `levelNulls` isolates state 1, `elementNulls` isolates state 4 from {2, 3}. States 2 and 3 collapse — both encode as a single phantom entry with `elementNulls[i] = true`. Only the underlying definition level separates them (`def < elementDefLevel` ⇒ empty), and `NestedBatchIndex` already uses exactly that signal in `PqListImpl.computeRange`. The internal reader gets it right; the public surface is just missing the bit.

A columnar consumer can't faithfully reconstruct the user-visible shape without it: `array_length`, equality checks, and any re-emit of the data give different answers for `[]` vs `[null]`. There's also a completeness gap relative to flat columns — a flat consumer can fully reconstruct values from `(values, elementNulls)`, whereas a nested consumer with `(values, offsets, levelNulls, elementNulls)` currently can't.

## Proposal
Surface the missing signal on `ColumnReader`. Two reasonable shapes:

  - **Derived (preferred)**: add `BitSet getEmptyListMarkers(int level)` parallel to `getLevelNulls(int level)`. Matches the existing style of pre-computing bitmaps in the drain thread; cost is essentially free since
  `NestedLevelComputer.computeLevelNulls` already walks the same def-level array.
  - **Raw (fallback)**: add `int[] getDefinitionLevels()` mirroring the typed-value array accessors. Most flexible, but pushes interpretation onto the caller.

The data already flows: `NestedBatch.definitionLevels` is populated by the drain thread an `NestedLevelComputer.computeLevelNullThresholds` returns the per-level thresholds. Adding an `emptyListMarkers` `BitSet[]` alongside `levelNulls`, computed in the same pass, is purely additive.

  ## Net effect

  - Drops three files from `COLUMN_SKIPPED_FILES` (`null_list.parquet`, `nullable.impala.parquet`, `nonnullable.impala.parquet`) added under #84 specifically because of this gap.
  - Closes the ergonomic gap for any downstream consumer reconstructing nested values from the columnar API.
  - `docs/content/usage.md`'s nested-column example (lines 618–646) gains a short paragraph describing the four states + new method, plus a corresponding example line.

  ## References

  - Follow-up to #84 (column-level nested validation).
  - Internal disambiguator: `PqListImpl.computeRange` uses `def < elementDefLevel` for empty-vs-element.
  - Threshold computation: `NestedLevelComputer.computeLevelNullThresholds`.
  - Public API home: `core/src/main/java/dev/hardwood/reader/ColumnReader.java`, `docs/content/usage.md` § "Nested and Repeated Columns".

  Once it's filed, you can update the inline comment in Utils.java's COLUMN_SKIPPED_FILES to point at the new issue number, and reference it from your PR review reply.

rionmonster:
## Description
This pull request closes #422, which observed that the public `ColumnReader` API couldn't distinguish an empty list from a list containing one null element — both encoded identically in `offsets` / `levelNulls` / `elementNulls`. A nested container has four meaningful states (null, empty, non-empty with null element(s), non-empty with value(s)), but the public API only distinguished three; the underlying definition level was the disambiguator, computed internally and used by `NestedBatchIndex` / `PqListImpl.computeRange`, but never surfaced.

## Key Changes
- **Expose empty-list disambiguation on `ColumnReader`** via a new public `BitSet getEmptyListMarkers(int level)` parallel to `getLevelNulls`, computed in the same drain-thread pass. Documented in `usage.md` and pinned by
  `NestedLevelComputerTest`.
- **Adopt the disambiguator in column-level `ParquetComparisonTest`** by having `reconstructNested` short-circuit on the new bitmap, dropping `null_list.parquet` from `COLUMN_SKIPPED_FILES`.
- **Handle legacy MAP encoding and list-of-map paths in the test walker** by accepting `map` as a structural alias for `key_value` and eliding list-structural names inside Map projections, dropping `nonnullable.impala.parquet`.
- **Canonicalize map iteration order in the column-level comparator** by sorting innermost lists for map-projection columns since `AvroParquetReader` returns `HashMap`-backed maps, dropping `nullable.impala.parquet` and emptying `COLUMN_SKIPPED_FILES`.
  - Removed entire `COLUMN_SKIPPED_FILES` as it is now empty after these changes.

## Verification and Testing
The primary verification and major change associated with this effort centers around the introduction of a new `NestedLevelComputerTest` class explicitly designed to verify the successful disambiguation between the various list types that are supported. In addition to this, the expected suite of tests all pass end-to-end:
```
[INFO] core .................................. Tests run: 893, Failures: 0, Errors: 0, Skipped: 0
[INFO] parquet-testing-runner ................. Tests run: 564, Failures: 0, Errors: 0, Skipped: 61
[INFO] BUILD SUCCESS
```

## Checklist

### DIFF
--- a/docs/content/usage.md
+++ b/docs/content/usage.md
@@ -617,7 +617,14 @@ try (Hardwood hardwood = Hardwood.create();
 
 ### Nested and Repeated Columns
 
-For nested columns (lists, maps), `ColumnReader` provides multi-level offsets and per-level null bitmaps:
+For nested columns (lists, maps), `ColumnReader` provides multi-level offsets and per-level null bitmaps. A list/map container has four states the reader exposes separately:
+
+- **Null** — `getLevelNulls(level)` set at the container's index.
+- **Empty** — `getEmptyListMarkers(level)` set; the container exists but has no entries.
+- **Non-empty with null element(s)** — neither `getLevelNulls` nor `getEmptyListMarkers` set; `getElementNulls()` flags the null leaves inside the range.
+- **Non-empty with values** — only `getElementNulls()` indicates per-leaf nullability (or is `null` if all leaves are required).
+
+Without `getEmptyListMarkers`, an empty list is indistinguishable from a list containing a single null element — both encode as a single phantom entry with `getElementNulls()` set.
 
 ```java
 try (ColumnReader reader = fileReader.columnReader("tags")) {
@@ -625,12 +632,14 @@ try (ColumnReader reader = fileReader.columnReader("tags")) {
         int recordCount = reader.getRecordCount();
         int valueCount = reader.getValueCount();
         byte[][] values = reader.getBinaries();
-        int[] offsets = reader.getOffsets(0);            // record -> value position
-        BitSet recordNulls = reader.getLevelNulls(0);    // null list records
-        BitSet elementNulls = reader.getElementNulls();  // null elements within lists
+        int[] offsets = reader.getOffsets(0);                 // record -> value position
+        BitSet recordNulls = reader.getLevelNulls(0);         // null list records
+        BitSet emptyMarkers = reader.getEmptyListMarkers(0);  // empty (size 0) lists
+        BitSet elementNulls = reader.getElementNulls();       // null elements within lists
 
         for (int r = 0; r < recordCount; r++) {
-            if (recordNulls != null && recordNulls.get(r)) continue;
+            if (recordNulls != null && recordNulls.get(r)) continue;     // null list
+            if (emptyMarkers != null && emptyMarkers.get(r)) continue;   // empty list
             int start = offsets[r];
             int end = (r + 1 < recordCount) ? offsets[r + 1] : valueCount;
             for (int i = start; i < end; i++) {
--- a/docs/content/usage.md
+++ b/docs/content/usage.md
@@ -626,16 +626,24 @@ For nested columns (lists, maps), `ColumnReader` provides multi-level offsets an
 
 `getEmptyListMarkers()` is what lets you tell an empty list apart from a list containing a single null element: in the underlying encoding, both produce a single phantom entry that `getElementNulls()` flags as null, so without this bitmap they'd be indistinguishable.
 
-For a column `optional group tags (LIST) { repeated group list { optional binary element (STRING); } }` holding these four records:
+For a column `optional group tags (LIST) { repeated group list { optional binary element (STRING); } }`, the four container states look like this:
 
 | Record index | Logical value | `getLevelNulls(0)` | `getEmptyListMarkers(0)` | `offsets[0]` | `getElementNulls()` at value index |
 |---|---|---|---|---|---|
-| 0 | `null`        | `1` | `0` | `0` | `1` (phantom) |
-| 1 | `[]`          | `0` | `1` | `1` | `1` (phantom) |
-| 2 | `[null]`      | `0` | `0` | `2` | `1` (real null element) |
-| 3 | `["x"]`       | `0` | `0` | `3` | `0` (`values[3] = "x"`) |
+| 0 | `null`     | `1` | `0` | `0` | `1` (phantom) |
+| 1 | `[]`       | `0` | `1` | `1` | `1` (phantom) |
+| 2 | `[null]`   | `0` | `0` | `2` | `1` (real null element) |
+| 3 | `["x"]`    | `0` | `0` | `3` | `0` (`values[3] = "x"`) |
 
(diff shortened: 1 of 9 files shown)


rionmonster:
@gunnarmorling 

Yet another for the upcoming milestone (and tying up the loose ends from the previous PR). It also completely removes the `COLUMN_SKIPPED_FILES` that we had, which is fun!

docs/content/usage.md
```diff
@@ -617,20 +617,29 @@ try (Hardwood hardwood = Hardwood.create();
...
 
-For nested columns (lists, maps), `ColumnReader` provides multi-level offsets and per-level null bitmaps:
+For nested columns (lists, maps), `ColumnReader` provides multi-level offsets and per-level null bitmaps. A list/map container has four states the reader exposes separately:
+
+- **Null** — `getLevelNulls(level)` set at the container's index.
+- **Empty** — `getEmptyListMarkers(level)` set; the container exists but has no entries.
+- **Non-empty with null element(s)** — neither `getLevelNulls` nor `getEmptyListMarkers` set; `getElementNulls()` flags the null leaves inside the range.
+- **Non-empty with values** — only `getElementNulls()` indicates per-leaf nullability (or is `null` if all leaves are required).
+
+Without `getEmptyListMarkers`, an empty list is indistinguishable from a list containing a single null element — both encode as a single phantom entry with `getElementNulls()` set.
```
gunnarmorling: I'd frame it positively: `getEmptyListMarkers()` allows to distinguish...

That said, how do you feel about the `ColumnReader` API in general? Admittedly, it's the part of the API I like the least (independent from this addition, which makes it functionally complete). It does the job, but it feels really complex and unintuitive; perhaps that just the nature of the beast though. Curious about your thoughts?

As for the docs, I think showing an example with some data would be nice; i.e. a few records with a list attribute, showcasing the different states.

parquet-testing-runner/src/test/java/dev/hardwood/testing/Utils.java
```diff
@@ -557,20 +558,6 @@ static Object convertToComparable(Object value) {
 
     // ==================== Column-Level Comparison ====================
 
-    /// Additional files to skip in column-level comparison tests beyond [#SKIPPED_FILES].
```
gunnarmorling, line 560: Noice!

gunnarmorling (commented):
LGTM overall. A few suggestions/comments inline.

docs/content/usage.md
rionmonster: @gunnarmorling 

Makes sense — I also like the idea of an example perhaps, specifically with data to help drive the point home a bit more clearly!

> That said, how do you feel about the ColumnReader API in general? Admittedly, it's the part of the API I like the least (independent from this addition, which makes it functionally complete). It does the job, but it feels really complex and unintuitive; perhaps that just the nature of the beast though. Curious about your thoughts?

I really need to carve out some time to _zoom out_ on it and spend some brainpower on it. I know it's grown pretty organically as we've added on complexity, which as you mentioned might just be par for the course, but it would be worth potentially identifying some areas to explore to simplify where we can (and if it makes sense with respect to performance and clarity).

gunnarmorling: One option could also be to mark the column reader as "Incubating" or "Experimental" for the 1.0 release for now. The row reader API in contrast feels pretty solid to me and I'd be ready to commit to it (opposing opinions welcome, of course).

rionmonster: I've gone ahead and updated the PR reframing the verbiage and including an example of the four cases involving the empty-list disambiguation as well!

rionmonster: > One option could also be to mark the column reader as "Incubating" or "Experimental" for the 1.0 release for now. The row reader API in contrast feels pretty solid to me and I'd be ready to commit to it (opposing opinions welcome, of course).

I actually like this idea, especially while things are still evolving before the 1.0 release. Making the leap from "Experimental" to "Stable" is much easier than attempting to revert or deprecate something from a seemingly reliable state.

gunnarmorling:
Thanks a lot, @rionmonster! Pushed a commit to adjust docs a bit, will merge.

gunnarmorling:
And merged. Thanks a lot, @rionmonster!

PR #428, opened 2026-05-05
#427 Adding Experimental annotation and column reader doc updates

gunnarmorling:


gunnarmorling:
Hey @rionmonster, added the `Experimental` annotation we discussed and also made a few quick doc changes for `ColumnReader`.

docs/content/usage.md
```diff
@@ -481,74 +481,14 @@ try (ParquetFileReader fileReader = ParquetFileReader.open(InputFile.of(path));
...
-}
-```
-
 ## Column-Oriented Reading (ColumnReader)
 
 The `ColumnReader` provides batch-oriented columnar access with typed primitive arrays, avoiding per-row method calls and boxing. This is the fastest way to consume Parquet data when you process columns independently.
 
-### Single-File Column Reading
+!!! warning "Experimental API"
+    The `ColumnReader` is under active development; The shape of the batch accessors and nested-offset representation may change in future releases without prior deprecation.
```
rionmonster: Tiniest of nits; I think the "The" here should be lowercased as "the" (or we just change the semicolon to a period).

docs/content/usage.md
```diff
@@ -645,6 +584,52 @@ try (ColumnReader reader = fileReader.columnReader("tags")) {
...
+import dev.hardwood.reader.RowReader;
+
+List<InputFile> files = InputFile.ofPaths(
+    Path.of("data_2024_01.parquet"),
+    Path.of("data_2024_02.parquet"),
+    Path.of("data_2024_03.parquet")
+);
+
+try (Hardwood hardwood = Hardwood.create();
+     ParquetFileReader parquet = hardwood.openAll(files);
```
rionmonster, line 605: Another minor nit: the comment above mentions `open() but the example uses `openAll()`. Likely fine, just unsure if those should align.

rionmonster (approved):
@gunnarmorling 

Approved! I left two minor nits but otherwise I think things make sense. I'll defer to you and your greater Java expertise on the best way to handle Experimental/Incubating functionality (as I think there are a lot of ways to handle that).

docs/content/usage.md
gunnarmorling: Ah yes, great catch. It needs to be `openAll()` up their too; only that method accepts a list.

gunnarmorling:
Thanks for reviewing, addressed both remarks and force-pushed.

> I'll defer to you and your greater Java expertise on the best way to handle Experimental/Incubating functionality

Yeah, that annotation path is pretty well-chartered territory, it's just our means of communicating that we reserve the right to break that API in backwards-incompatible ways (if need be), not following semantic version principles. Essentially saying "use at your own risk". Similar to what the JDK itself does with its "incubating"/"preview" notions.

Issue #431, opened 2026-05-07
feat(reader): RowGroupPredicate for split-aware row group selection


gunnarmorling:
## Motivation

Integrations that partition a Parquet file across parallel readers (Flink's `BulkFormat`, Spark's file-source reader, MapReduce-style splits) need to assign each reader a disjoint byte range and read only the row groups owned by that range. Today `ParquetFileReader.open(InputFile)` always reads every row group, so:

- The Flink integration's `BulkFormat.isSplittable()` has to return `false`, which gives up file-level parallelism.
- Test scenarios that hand the reader byte-bounded splits (e.g. `ParquetColumnarRowInputFormatTest` in the Flink-on-Hardwood experiment at MartijnVisser/flink@f8303bc) double-read row groups when run against multiple splits.

## Proposal

Introduce a sealed `RowGroupPredicate` type and a `filter(RowGroupPredicate)` overload on `ColumnReaderBuilder`:

```java
public sealed interface RowGroupPredicate permits ByteRange, And {
    static RowGroupPredicate byteRange(long startInclusive, long endExclusive);
    static RowGroupPredicate and(RowGroupPredicate... children);
    record ByteRange(long startInclusive, long endExclusive) implements RowGroupPredicate {}
    record And(List<RowGroupPredicate> children) implements RowGroupPredicate {}
}
```

```java
ColumnReader col = file.buildColumnReader("price")
        .filter(FilterPredicate.gt("price", 100))                    // column-stats
        .filter(RowGroupPredicate.byteRange(splitStart, splitEnd))   // row-group layout
        .build();
```

Both `filter(...)` overloads apply (intersection): a row group is read iff it passes both. `ByteRange` uses the standard midpoint convention (start of first column chunk + half compressed size in `[start, end)`), so every row group lands in exactly one byte range across a partitioning of the file.

## Why a separate type from `FilterPredicate`

`FilterPredicate` predicates target column values via per-column statistics. `RowGroupPredicate` predicates target row-group physical layout. The combinators have different semantics (`or`/`not` of byte-ranges is incoherent at the row-group level) and the evaluators read different metadata. Sharing the verb `filter()` keeps the call site uniform; sharing the type would conflate two different abstractions.

## Why "row group" is in the type name

[...]

PR #432, opened 2026-05-07
#431 Add RowGroupPredicate for split-aware row group selection

gunnarmorling:
Hadoop-style integrations (Flink BulkFormat, Spark file source, …) need to assign each parallel reader a byte range and read only the row groups owned by it. Previously ParquetFileReader.open(InputFile) always read every row group, so such integrations had to forfeit file-level parallelism. Introduce RowGroupPredicate.byteRange(start, end) that selects row groups by midpoint and a filter(RowGroupPredicate) overload on every reader builder, composing with FilterPredicate via intersection.

### DIFF
--- a/core/src/main/java/dev/hardwood/reader/ParquetFileReader.java
+++ b/core/src/main/java/dev/hardwood/reader/ParquetFileReader.java
@@ -231,29 +231,30 @@ public class ParquetFileReader implements AutoCloseable {
     // ============================================================
 
     RowReader buildRowReader(ColumnProjection projection, FilterPredicate filter, long maxRows) {
-        return buildRowReader(projection, filter, maxRows, 0L);
+        return buildRowReader(projection, filter, null, maxRows, 0L);
     }
 
-    RowReader buildRowReader(ColumnProjection projection, FilterPredicate filter, long maxRows,
-                             long firstRow) {
+    RowReader buildRowReader(ColumnProjection projection, FilterPredicate filter,
+                             RowGroupPredicate rowGroupFilter, long maxRows, long firstRow) {
+        // Apply the row-group predicate (e.g. byte-range) up front so `firstRow` indexes
+        // into the kept sequence — a caller doing split-aware reading can seek inside *its*
+        // split. Stats-based row-group dropping (via FilterPredicate) stays inside the
+        // RowGroupIterator, so its semantics under `firstRow` are unchanged from before.
+        List<RowGroup> filteredRowGroups = rowGroupFilter == null
+                ? firstFileMetaData.rowGroups()
+                : filterRowGroups(null, rowGroupFilter);
+
         if (firstRow == 0L) {
-            return buildRowReader(projection, filter, maxRows, firstFileMetaData.rowGroups());
+            return buildRowReader(projection, filter, maxRows, filteredRowGroups);
         }
         // Locate the row group containing `firstRow` by walking cumulative
-        // RowGroup.numRows(). For multi-file readers this indexes into the
-        // first file only — cross-file firstRow is out of scope.
-        List<RowGroup> all = firstFileMetaData.rowGroups();
-        long totalRows = firstFileMetaData.numRows();
-        if (firstRow > totalRows) {
-            throw new IllegalArgumentException(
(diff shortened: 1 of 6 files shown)


gunnarmorling:
Hey @rionmonster, @iifawzi, this enables rowgroup-level parallelism for sharded consumers like Flink.

In theory, this would also be doable by adding `head()` and `firstRow()` to the column reader API (arguably, they are missing anyways, so I've logged #433 for adding them, ensuring parity with the row reader API), leaving it to consumers to query the metadata and do the math themselves for expressing row group limits via offsets/limits. I've still opted for the more explicit `RowGroupPredicate` API for a) ease-of-use, and b) it let's consumers process multiple non-contiguous row groups, which wouldn't be possible with offset/limit alone.

Any feedback welcome!

core/src/main/java/dev/hardwood/reader/ParquetFileReader.java
```diff
@@ -231,43 +231,52 @@ public ColumnReadersBuilder buildColumnReaders(ColumnProjection projection) {
...
+        long totalRows = 0;
+        for (RowGroup rg : filteredRowGroups) {
+            totalRows += rg.numRows();
+        }
         if (firstRow > totalRows) {
             throw new IllegalArgumentException(
-                    "firstRow " + firstRow + " exceeds the file's total row count "
+                    "firstRow " + firstRow + " exceeds the (row-group-filtered) total row count "
                     + totalRows);
         }
```
iifawzi: nit: perhaps we can have it as a part of the loop below checking the target rg
```java
long cumulative = 0L;
int targetRg = filteredRowGroups.size(); // sentinel: at-the-end yields empty reader
long withinRg = 0L;

for (int i = 0; i < filteredRowGroups.size(); i++) {
    long rgRows = filteredRowGroups.get(i).numRows();

    if (firstRow < cumulative + rgRows) {
        targetRg = i;
        withinRg = firstRow - cumulative;
        break;
    }

    cumulative += rgRows;
}

if (targetRg == filteredRowGroups.size()) {
    if (firstRow > cumulative) {
        throw new IllegalArgumentException(
                "firstRow " + firstRow
                + " exceeds the (row-group-filtered) total row count "
                + cumulative);
    }

    // firstRow == totalRows — empty reader.
    return buildRowReader(
            projection,
            filter,
            maxRows,
            List.<RowGroup>of());
}
```

iifawzi (approved):

core/src/main/java/dev/hardwood/reader/ParquetFileReader.java
```diff
@@ -403,9 +428,18 @@ private static List<RowGroup> tailRowGroups(List<RowGroup> rowGroups, long tailR
...
+    /// Filter row groups by an optional column-statistics predicate and an optional
+    /// [RowGroupPredicate]. A row group is kept if and only if it passes both.
+    private List<RowGroup> filterRowGroups(
+            ResolvedPredicate filter, RowGroupPredicate rowGroupFilter) {
         List<RowGroup> all = firstFileMetaData.rowGroups();
         List<RowGroup> kept = all.stream()
-                .filter(rg -> !RowGroupFilterEvaluator.canDropRowGroup(filter, rg))
+                .filter(rg -> filter == null
+                        || !RowGroupFilterEvaluator.canDropRowGroup(filter, rg))
+                .filter(rg -> rowGroupFilter == null || matches(rg, rowGroupFilter))
```
iifawzi: could switching the order to filter out row groups outside of the RowGroupPredicate be better? i'm feeling row group predicative would be more selective

core/src/main/java/dev/hardwood/reader/ParquetFileReader.java
gunnarmorling: Yes, good idea.

core/src/main/java/dev/hardwood/reader/ParquetFileReader.java
gunnarmorling: Yes, definitely better to do that first, will change.

gunnarmorling:
Applied; thanks a lot for reviewing, @iifawzi!

Issue #435, opened 2026-05-07
feat(schema): getMapKey() / getMapValue() parity with getListElement()


gunnarmorling:
## Motivation

`SchemaNode.GroupNode` exposes `getListElement()` to navigate the standard 3-level (and legacy 2-level) LIST encoding. The MAP equivalent is missing — consumers walk `group.children().get(0)` to the `key_value` subgroup, then `.children().get(0)` and `.children().get(1)` for key and value, with type-checks and back-compat rules duplicated at every call site.

This came up while writing the Flink-on-Hardwood vectorized integration ([MartijnVisser/flink#hardwood-poc](https://github.com/MartijnVisser/flink/tree/hardwood-poc)) — the integration's MAP-handling code is ~10 lines of structural navigation that anyone touching MAP groups would re-write the same way.

## Proposal

Add to `SchemaNode.GroupNode`:

```java
/// For MAP groups, returns the key node from the standard encoding
/// (`map.key_value.key`). Returns `null` if not a MAP, if the structure is
/// malformed, or if the encoding doesn't expose a key/value pair.
public SchemaNode getMapKey();

/// For MAP groups, returns the value node. See [#getMapKey()].
public SchemaNode getMapValue();
```

Implementation is a small private helper that validates the standard encoding shape (single REPEATED `key_value` child with two grand-children) and returns the matching child by index. `null`-on-malformed matches the existing `getListElement()` behavior — the caller decides whether that's an error.

## Tests

- `getMapKey()` / `getMapValue()` against existing nested fixtures with maps; if none exists for top-level MAPs, add a small synthetic fixture.
- Round-trip: standard 3-level encoding produces non-null key/value nodes.
- Negative cases: non-MAP group → null; MAP group with malformed `key_value` shape → null.

## Docs

Update `docs/content/usage.md` § "Nested and Repeated Columns" to mention `getMapKey()` / `getMapValue()` alongside `getListElement()`.

## Related

- #422 (closed) — added `getEmptyListMarkers` for list-state disambiguation. Same area (nested-column ergonomics), different concern.
- #437 — schema visitor pattern. Higher-level recursive-walking abstraction, deferred until a second nested-type consumer surfaces.
- #431 — `RowGroupPredicate`. Unrelated, but the same Flink experiment surfaced this gap.

gunnarmorling:
Rescoped down to just `getMapKey()` / `getMapValue()`.

[...]

Issue #434, opened 2026-05-07
feat(reader): coordinated nextBatch() on ColumnReaders


gunnarmorling:
## Motivation

`ColumnReaders` already shares a single `RowGroupIterator` across all projected columns, but each underlying `ColumnReader.nextBatch()` is still independent. Consumers reading multiple columns in lockstep (which is the entire point of the columnar projection API) have to advance every reader by hand and reconcile per-batch record counts themselves. The `ColumnReaders` javadoc itself shows this in the example:

```java
while (col0.nextBatch() & col1.nextBatch() & col2.nextBatch()) {
    int count = col0.getRecordCount();
    // ...
}
```

The bitwise-`&` is doing duty for "advance them all and stop when any is exhausted." Real consumers also need to assert that all readers report the same record count per batch — Hardwood's drain workers all use the same `DEFAULT_BATCH_SIZE` and consume from the same iterator, so they should align, but the consumer ends up writing a defensive check anyway.

This came up concretely when I worked on the Flink-on-Hardwood vectorized integration ([MartijnVisser/flink#hardwood-poc](https://github.com/MartijnVisser/flink/tree/hardwood-poc)). The integration's per-row-batch coordination is ~25 lines of `advanceHardwoodBatch()` plus an "Hardwood leaf batch sizes diverged" sanity throw — all of which would be one method call if `ColumnReaders` provided coordinated advancement.

A side effect of the gap: integrations that don't already use `ColumnReaders` (the Flink experiment doesn't — it opens N independent `ColumnReader`s) miss the shared-iterator efficiency entirely. A coordinated `nextBatch()` would make `ColumnReaders` more obviously the right choice for multi-column consumers.

## Proposal

Add to `ColumnReaders`:

```java
/// Advance every underlying ColumnReader to its next batch in lockstep.
/// Returns false when any reader is exhausted (in which case all readers
/// are considered done — partial advancement is impossible because every
/// reader consumes from the same RowGroupIterator).
///
/// Throws IllegalStateException if the readers report mismatched record
/// counts for the just-published batch — a safety net guarding against
/// internal bugs in the per-column drain workers.
public boolean nextBatch();

/// Number of records in the most recently published batch. Equal to every
/// underlying reader's getRecordCount() (validated by nextBatch()).
public int getRecordCount();
```

Usage becomes:

```java
try (ColumnReaders columns = parquet.buildColumnReaders(projection).build()) {
    while (columns.nextBatch()) {
[...]

Issue #61, opened 2026-02-25
ColumnReader batch alignment across multiple readers is not guaranteed by contract


gunnarmorling:
When consuming batches from multiple ColumnReader instances in lockstep (as shown in the README and done in FlatPerformanceTest), the code assumes all readers return batches of the same size:

```
while (col0.nextBatch() & col1.nextBatch() & col2.nextBatch()) {
  int count = col0.getRecordCount();  // used for all three columns
  double[] v0 = col0.getDoubles();
  double[] v1 = col1.getDoubles();
  double[] v2 = col2.getDoubles();
  for (int i = 0; i < count; i++) { ... }
}
```

This happens to work today because all ColumnReader instances use the same DEFAULT_BATCH_SIZE (262,144) and ColumnAssemblyBuffer fills batches greedily without regard to page boundaries. Since Parquet guarantees equal row counts across columns, the batch sizes end up aligned.

However, this is an implicit invariant — not documented or enforced. A future change (e.g., aligning batches to row group or page boundaries) could silently break this, causing data corruption or ArrayIndexOutOfBoundsException.

Options:

1. Document the guarantee — specify that ColumnReader instances created from the same file/MultiFileColumnReaders always produce aligned batches
1. Enforce it structurally — e.g., have MultiFileColumnReaders coordinate batch advancement so alignment is a structural property rather than a coincidence
1. Add defensive checks — assert that all readers in a lockstep group return the same getRecordCount()

gunnarmorling:
#434 (coordinated `nextBatch()` on `ColumnReaders`) implements option 2 of this issue (and option 3 — it adds an `IllegalStateException` guard on mismatched record counts) for the multi-column projection path. Once it lands, `ColumnReaders` becomes the structurally-safe path: a single `nextBatch()` call drives every underlying reader and validates alignment internally.

That leaves #61's residual scope as the **independent-reader** case — consumers using `parquet.columnReader("name")` × N (the pattern in the README and `FlatPerformanceTest`, and what the [Flink hardwood-poc experiment](https://github.com/MartijnVisser/flink/tree/hardwood-poc) currently uses). For that path, the right resolution is option 1: document the invariant on `ColumnReader.nextBatch()` ("every `ColumnReader` over the same file produces aligned batches") and steer multi-reader consumers toward `ColumnReaders` as the canonical, structurally-safe alternative.

Could close this once #434 ships and the README + relevant javadoc are updated to:

[...]

PR #443, opened 2026-05-09
#434 Coordinated nextBatch() on ColumnReaders

gunnarmorling:
This adds three independent, yet related improvements to the column reader API:

- #434, simplifying life for consumers reading multiple columns in lockstep
- #61, making explicit that batches from multiple readers are aligned
- #435 Adding `SchemaNode.GroupNode.getMapKey()` / `getMapValue()`, for parity with the already existing `getListElement()`


gunnarmorling:
@rionmonster, @iifawzi, just in case one of you wants to take a look, it closes a few gaps around column reader usage identified while working on integrating this into Flink's Parquet module. I have a more holistic rework of the column reader API in progress too (the current form just isn't great for nested schemas), but these things here are independent from that and are an improvement.

core/src/main/java/dev/hardwood/reader/ColumnReaders.java
```diff
@@ -91,6 +96,85 @@ public ColumnReader getColumnReader(int index) {
...
+            int count = readersByIndex[i].getRecordCount();
+            if (count != firstCount) {
+                throw new IllegalStateException(
+                        "ColumnReader batch sizes diverged: column '"
+                                + readersByIndex[0].getColumnSchema().name() + "' has "
+                                + firstCount + " records, column '"
+                                + readersByIndex[i].getColumnSchema().name() + "' has "
+                                + count);
+            }
+        }
```
iifawzi, line 157: I was thinking that if this is solely for regression detections, we can rely on tests for that and save the iterations on such a hot path, but I believe being defensive here could be useful just in case there's a mixed usage of ColumnReaders & columnReader that one has been advanced manually, causing inconsistency. Without this defensive check here, consumers would end up with corrupted data, so it seems reasonable

iifawzi (approved):
Looks good to me

gunnarmorling:
Rebased and applied; thanks for reviewing, @iifawzi!

Issue #436, opened 2026-05-07
feat(reader): distinguish struct-null from field-null on flat ColumnReader


gunnarmorling:
## Problem

For a flat leaf inside an `OPTIONAL` non-repeated group ancestor, the leaf's definition levels encode three states:

| def-level | meaning |
|---|---|
| 0 | the ancestor group is null (struct null) |
| 1 | the ancestor group exists, this field is null (field null) |
| 2 | this field has a value |

Hardwood's flat `ColumnReader.getElementNulls()` collapses every `def < maxDefinitionLevel` into a single null bit, so states {0, 1} merge. A consumer that needs to faithfully reconstruct user-visible shapes — "the struct is null" vs "all of the struct's fields happen to be individually null" — can't tell them apart.

This came up directly in the Flink-on-Hardwood vectorized integration ([MartijnVisser/flink#hardwood-poc](https://github.com/MartijnVisser/flink/tree/hardwood-poc)). For a top-level `RowType` field whose Parquet group is OPTIONAL, the integration today uses a heuristic: "if every child reports null at row r, mark the struct itself null in `HeapRowVector`." That's correct only when the writer never produces (struct non-null, all fields independently null) rows. For the Flink test fixtures it happens to be safe; in general it isn't.

This is **distinct from #422**, which solved the analogous ambiguity for *repeated* containers (empty list vs list-with-one-null-element). #422 added `getEmptyListMarkers(int level)` to `ColumnReader`, but that accessor is only valid for `level < maxRepetitionLevel` — it doesn't apply to flat leaves under non-repeated nullable ancestors. Different code path, different abstraction.

## Proposal

Two viable shapes, mirroring #422's framing:

### Derived (preferred)

Add to `ColumnReader` for flat leaves whose schema chain has at least one OPTIONAL non-repeated group ancestor:

```java
/// For flat columns whose schema chain includes an OPTIONAL non-repeated
/// group ancestor (e.g. a leaf inside an OPTIONAL struct), returns a
/// BitSet where set bits indicate that an ancestor group is null at that
/// row — as distinct from getElementNulls(), which would then mean
/// "ancestor non-null but the leaf field is null".
///
/// Returns null when no nullable group ancestor exists (i.e. the column
/// is required-or-flat-optional with no nullable parent), since there's
/// nothing to distinguish.
public BitSet getAncestorGroupNulls();
```

[...]

PR #450, opened 2026-05-11
#430 Column reader re-design

### LINKED ISSUE #430: feat(reader): ColumnReader layer-model rework
## Status

Decided. Implementation tracker. Authoritative design lives in [_designs/COLUMN_READER_ARROW_LAYOUT.md](https://github.com/hardwood-hq/hardwood/blob/main/_designs/COLUMN_READER_ARROW_LAYOUT.md).

## Decision

The `ColumnReader` columnar surface is reshaped around a uniform per-layer model:

- Each `OPTIONAL` group, `LIST`, and `MAP` along the column's schema chain contributes one layer; `REQUIRED` groups don't. `STRUCT` layers are exposed (closing the struct-null vs. field-null gap from #436).
- Per-layer accessors: `getLayerCount()`, `getLayerKind(int)` (`STRUCT` or `REPEATED`), `getLayerValidity(int)` (set bit = present, sparse-`null` for all-present), `getLayerOffsets(int)` (sentinel-suffixed, `REPEATED` only).
- Leaf accessors: `getLeafValidity()` plus typed value arrays. Varlength leaves use `getBinaryValues()` + `getBinaryOffsets()` (capacity-sized bytes, sentinel-suffixed offsets); `getBinaries()` and `getStrings()` retained as convenience wrappers that allocate per row.
- Raw def-/rep-level escape hatch: `getDefinitionLevels()` / `getRepetitionLevels()` for the long tail.
- `getElementNulls`, `getLevelNulls`, `getEmptyListMarkers`, `getNestingDepth` removed; empty-list state is encoded structurally as ``offsets[i+1] - offsets[i] == 0``.
- Validity polarity flips to *set bit = present* end-to-end (workers, internal storage, public API).
- Internal: `NestedBatch` / `BatchExchange.Batch` re-keyed onto layer-indexed validity; varlength `values` slot becomes `(byte[], int[])`. `RowReader`'s public API is unchanged.

## Not Arrow-bit-compatible

The model follows Arrow *in spirit* — same per-layer concept, same predicate polarity, same offsets-plus-values varlength shape — but the backing storage is `BitSet` / `int[]` / `byte[]`, not Arrow IPC buffers. Bit-for-bit interop is a separate concern.

## Subsumes

- #436 — struct-null vs. field-null on flat ColumnReader. Falls out of the new STRUCT layer slots.

## Related

- #422 — `getEmptyListMarkers`, retired by the new offsets contract.
- #153 — future `ByteBuffer` / `IntBuffer` storage swap. Preserves the public method shape, polarity, sentinels, and capacity-sized contract from this rework.

## External consumer follow-up

The Flink-on-Hardwood [hardwood-poc](https://github.com/MartijnVisser/flink/tree/hardwood-poc) branch is built against today's API and will break when this rework lands. Adjusting it is a separate, later change tracked after `1.0.0.CR1` ships.

## Target

`1.0.0.CR1`. Hard cutover under the existing `@Experimental` annotation; no deprecation cycle.

gunnarmorling:
_(description via Claude)_

## Rework `ColumnReader` around per-layer validity and offsets

Closes #430. Subsumes #436.

The previous `ColumnReader` API counted only Parquet repetition levels, which forced two compromises on consumers:

1. **Struct-null and leaf-null were indistinguishable.** A leaf inside `optional struct { optional int x }` couldn't tell "struct is null" from "struct present, x is null" — both collapsed onto one cleared bit on the leaf's null bitmap (#436).
2. **LIST/MAP four-state encoding required two bitmaps.** Distinguishing `null` / `[]` / `[null]` / `[v]` needed `getLevelNulls` *plus* `getEmptyListMarkers`.

This PR exposes a column's schema chain as a sequence of layers — `STRUCT` for user-authored `OPTIONAL` groups, `REPEATED` for `LIST`/`MAP` groups — each carrying its own validity, and for `REPEATED` its own sentinel-suffixed offsets buffer sized to real items only. Both compromises go away: per-layer validity disambiguates struct-null from leaf-null at any depth, and `offsets[r+1] - offsets[r] == 0` distinguishes empty from non-empty without a separate empty-marker bitmap.

### Before / after — a simple list scan

```java
// Before
BitSet recordNulls  = reader.getLevelNulls(0);
BitSet emptyMarkers = reader.getEmptyListMarkers(0);
BitSet elementNulls = reader.getElementNulls();
int[]  offsets      = reader.getOffsets(0);
int    valueCount   = reader.getValueCount();
for (int r = 0; r < recordCount; r++) {
    if (recordNulls != null && recordNulls.get(r)) continue;
    if (emptyMarkers != null && emptyMarkers.get(r)) continue;
    int start = offsets[r];
    int end   = (r + 1 < recordCount) ? offsets[r + 1] : valueCount;
    for (int i = start; i < end; i++) {
        if (elementNulls == null || !elementNulls.get(i)) sum += values[i];
    }
}
```

```java
// After
Validity listValidity = reader.getLayerValidity(0);
Validity leafValidity = reader.getLeafValidity();
int[]    offsets      = reader.getLayerOffsets(0);   // length recordCount + 1
for (int r = 0; r < recordCount; r++) {
    if (listValidity.isAbsent(r)) continue;
    int start = offsets[r];
    int end   = offsets[r + 1];                     // no last-record ternary
    if (start == end) continue;                     // empty list, no separate bitmap
    for (int i = start; i < end; i++) {
        if (leafValidity.isPresent(i)) sum += values[i];
    }
}
```

Three bitmaps drop to two. The last-record ternary goes away (offsets are sentinel-suffixed). Validity reads naturally as a present-polarity predicate (`isPresent` / `isAbsent`) instead of the double-negative `!nulls.get(i)`. The `null`-check footgun on the validity reference is gone — `Validity` is a sealed type with an `ALL_PRESENT` singleton for the no-absents fast path, and both predicates handle either case correctly.

### Other consumer-facing changes

- **`LayerKind` makes STRUCT vs REPEATED a first-class type-level distinction**; no more inferring list-ness from "did `getOffsets` return something."
- **Varlength leaves** get an Arrow-shaped `getBinaryValues()` + `getBinaryOffsets()` buffer pair for hot paths; `getBinaries()` / `getStrings()` are retained as per-row-allocating convenience accessors.
- **`values[i]` means the i-th real leaf**, not an arbitrary position in the raw def/rep stream that may or may not be meaningful.

Hard cutover under `ColumnReader`'s existing `@Experimental` annotation; no deprecation cycle.


gunnarmorling:
Hey @rionmonster, @iifawzi, would love your take on this one, if you have the time. It substantially improves ergonomics when dealing with nested schemas in the column reader API (e.g. thinking of that null list ambiguation gap you recently closed, @rionmonster). It also moves from `BitSet` to a dedicated `Validity` type in the API, which should grant us some more flexibility in terms of implementation changes going forward.

rionmonster:
> Hey @rionmonster, @iifawzi, would love your take on this one, if you have the time. It substantially improves ergonomics when dealing with nested schemas in the column reader API (e.g. thinking of that null list ambiguation gap you recently closed, @rionmonster). It also moves from `BitSet` to a dedicated `Validity` type in the API, which should grant us some more flexibility in terms of implementation changes going forward.

Sure thing - the should have some downtime today (traveling), so we ll see if I can dive in and put some eyes on it! 👍

iifawzi:
Taking a deeper look with care, after work, tomorrow

gunnarmorling:
I'm gonna go ahead and merge this one, so as to unblock a few other things. Feedback still welcome if you wanna check out the PR in more depth. The column reader is marked experimental, so we still can make adjustments also after 1.0, so I'm not too worried. I think it's in a much better shape than before.

PR #458, opened 2026-05-11
#445 Decoded getValue() across reader API + typed INTERVAL accessors

### LINKED ISSUE #445: PqList element access: missing INTERVAL/FLOAT16/VARIANT accessors and incorrect raw/converted doc
## Background

`PqList` and `PqMap.Entry` element access have related defects that surfaced during PR #439 review:

### 1. `ValueConverter.convertValue` omits `IntervalType`

`core/src/main/java/dev/hardwood/internal/reader/ValueConverter.java` switches on logical type and falls back to physical type. `IntervalType` is not in the chain, so an INTERVAL element (FLBA(12)) falls through to `convertToBinary` and surfaces as `byte[]`. There is no `ValueConverter.convertToInterval` method either — the conversion path was never wired up. Scalar `getInterval(name)` returns `PqInterval` correctly; the list path silently disagrees.

### 2. `PqList.get(int)` and `values()` JavaDoc claim raw, but the impl converts

`core/src/main/java/dev/hardwood/row/PqList.java`:

```java
/// Get a raw element by index without type conversion.
Object get(int index);

/// Iterate over elements as raw objects without type conversion.
Iterable<Object> values();
```

`PqListImpl.getLeafValue` and `PqListImpl.values()` both route through `ValueConverter.convertValue`, which decodes UUID/BigDecimal/Date/Time/Timestamp/String/Bson/Float16. The decoded behavior is what users want; the doc text just predates those conversions being added. Fix the doc, not the code.

(Note: `PqMap.Entry.getKey()` and `getValue()` *do* return raw — `PqMapImpl.readKey`/`readValue` go straight to `batch.getValue` with no conversion. No doc bug there.)

### 3. Missing typed list accessors on `PqList`

| Element type | Typed accessor | `values()` result | `get(int)` result |
|---|---|---|---|
| INTERVAL | **missing** `intervals()` | `byte[]` (no `IntervalType` in `convertValue`) | `byte[]` |
| VARIANT | **missing** `variants()` | `?` | `?` |

For FLOAT16 lists, `floats()` should subsume them — `Iterable<Float>` is uniformly boxed regardless of underlying encoding, matching the existing pattern where `getDate` accepts INT32+DateType and `getTimestamp` accepts INT64+TimestampType *or* legacy INT96. So `convertToFloat` should dispatch on element schema (FLOAT physical → existing path; FLBA(2)+Float16Type → 2-byte decode). No new `float16s()` method needed.

### 4. Missing typed accessors on `PqMap.Entry`

`core/src/main/java/dev/hardwood/row/PqMap.java`:

**Values** — load-bearing omissions, mirror the PqList gaps:
- `getIntervalValue()` returning `PqInterval`
- `getVariantValue()` returning `PqVariant`
- (FLOAT16 values: subsume into `getFloatValue()` — same NPE-on-null primitive contract works for either encoding.)

**Keys** — less common in practice but still real omissions:
- `getTimeValue()` exists; `getTimeKey()` does not.
- `getDecimalValue()` exists; `getDecimalKey()` does not.
- (Float, double, boolean keys are spec-legal but vanishingly rare; punt on those unless someone asks.)

## Proposed fix (one PR)

1. Add `ValueConverter.convertToInterval(rawValue, schema)` validating `IntervalType`, delegating to `LogicalTypeConverter.convertToInterval`.
2. Add `case LogicalType.IntervalType -> convertToInterval(...)` to `ValueConverter.convertValue`.
3. Make `ValueConverter.convertToFloat` dispatch on schema: physical FLOAT (existing) or FLBA(2)+Float16Type (decode via `LogicalTypeConverter.convertToFloat16`). Then `PqList.floats()` and `PqMap.Entry.getFloatValue()` work transparently against FLOAT16 element columns.
4. Add `intervals()` and `variants()` to `PqList` + `PqListImpl` following the existing `uuids()` / `decimals()` pattern.
5. Add `getIntervalValue()`, `getVariantValue()`, `getTimeKey()`, `getDecimalKey()` to `PqMap.Entry` + `PqMapImpl`.
6. Update `PqList.get(int)` and `PqList.values()` JavaDoc to describe what the methods actually do — return the value decoded to its logical type (UUID, BigDecimal, PqInterval, Float for FLOAT16, etc.), with FLBA/BYTE_ARRAY surfacing as `byte[]` only when no logical type is annotated.

## Tests

- INTERVAL list: scalar `getInterval` and list `intervals()` / `values()` / `get(0)` all yield equivalent `PqInterval`.
- INTERVAL map value: scalar `getInterval` and map entry `getIntervalValue()` yield equivalent `PqInterval`.
- FLOAT16 list: `floats()` yields `Float` values matching `getFloat16` output; no separate accessor exposed.
- VARIANT list and VARIANT map value: yield `PqVariant`.
- TIME-keyed map and DECIMAL-keyed map: `getTimeKey()` / `getDecimalKey()` work.

## Out of scope

- Scalar accessor semantics (already correct).
- Stats / predicate pushdown over list / map elements.
- Float/double/boolean map keys (spec-legal, vanishingly rare).

iifawzi:
## Summary

Resolves [hardwood#445](https://github.com/hardwood-hq/hardwood/issues/445) plus the design follow-ups in [comment #4415695987](https://github.com/hardwood-hq/hardwood/issues/445#issuecomment-4415695987) and [comment #4415750885](https://github.com/hardwood-hq/hardwood/issues/445#issuecomment-4415750885).

Three threads:

1. **Bug fix.** `ValueConverter.convertValue` had no `IntervalType` case, so INTERVAL elements inside a `PqList` surfaced as raw `byte[]`. Wired `convertToInterval` in.
2. **Typed-accessor gaps.** Added `PqList.intervals()`, `PqMap.Entry.getIntervalValue()`, `getTimeKey()`, `getDecimalKey()`. (FLOAT16 and VARIANT-in-list / VARIANT-as-map-value are deferred — neither has reader support today; shredded variants inside repeated contexts are explicitly rejected.)
3. **Consistent generic access.** `getValue()` on `RowReader` / `PqStruct` / `PqMap.Entry` now returns the value **decoded** to its logical-type form, matching what `PqList.get` / `PqList.values()` already did. A new parallel `getRawValue()` / `PqMap.Entry.getRawKey()` / `getRawValue()` exposes the underlying physical form. Dive's `t` toggle is now threaded through nested rendering so it flips Map / Struct primitives between decoded and raw at every depth — previously only the top-level primitive cell respected the toggle, and Map / Struct cells rendered raw hex regardless of mode (the original screenshot bug).

## Breaking changes

This changes the contract of `RowReader.getValue` / `PqStruct.getValue` / `PqMap.Entry.getKey()` / `PqMap.Entry.getValue()` from raw to decoded. Callers that expected, e.g., a `Long` for a TIMESTAMP now get an `Instant`; callers that wanted a UTF-8 `byte[]` for a STRING now get a `String`. Use the new `getRawValue` / `getRawKey` for the physical form. The maintainer signed off on the breaking change in the issue thread ("we're still pre 1.0 Final").

## Fixture & annotators

A new fixture `core/src/test/resources/typed_accessors_issue_445.parquet` covers INTERVAL list elements (including a null element), INTERVAL-valued maps (including a null value), TIME-keyed maps, and DECIMAL-keyed maps. PyArrow can't emit INTERVAL natively and can't target inner list / map SchemaElements, so `tools/parquet_annotators.py` gains path-aware helpers (`annotate_element_at_path_as_interval` / `_as_time` / `_as_decimal`).

## Test plan

- [x] `TypedAccessorsIssue445Test` (7 cases): INTERVAL list typed + decoded + indexed access; null INTERVAL element; INTERVAL map value with raw-bytes byte-for-byte check; null INTERVAL map value; TIME-keyed map (`getTimeKey` + decoded `getKey` + raw `getRawKey`); DECIMAL-keyed map; RowReader `getValue` / `getRawValue` agreement on a no-logical-type column.
- [x] Existing core suite: 902 pass (+2 over the prior 900 baseline thanks to the null-case additions).
- [x] CLI suite (skipping S3/Docker-gated tests): 315 pass.
- [x] Manual: `java -jar cli/target/quarkus-app/quarkus-run.jar dive -f core/src/test/resources/filter_pushdown_nested_ts.parquet` — `event` struct cell now renders timestamps in logical mode; pressing `t` flips them to raw long micros (was: no effect on nested cells before).
- [x] Public API docs updated under `docs/content/usage.md` covering decoded-vs-raw semantics and the new typed accessors.


iifawzi:
Hey @gunnarmorling, this PR resolves #445 as per the discussions we had. Finally, it isn't a lot to land in the same PR, let me know your thoughts

gunnarmorling:
Thx, @iifawzi! I've rebased this one to current main to resolve the conflicts and added one more commit to sort out a few more related issues. Also logged #460 as a follow-up. Gonna close this one in favor of #461 with both your and my commits.

Issue #440, opened 2026-05-08
Replace BitSet null bitmaps with raw long[] on the read hot path


iifawzi:
## Background

`FlatRowReader` reads null bitmaps via `BitSet[] flatNulls`, and every primitive accessor (`getInt`, `getLong`, `getDouble`, …) plus `isNull` calls `BitSet.get(rowIndex)` once per row. `BitSet.get(int)` performs a bounds check and word-index computation that the JIT cannot always strip, adding overhead to one of the hottest paths in the reader.

`BitSet` is already backed by a `long[]` internally — so the layout is unchanged; the proposal is to drop the wrapper and operate on the raw `long[]` directly.

## Proposal

Switch the null bitmap representation from `BitSet` to `long[]` end-to-end:

- `BatchExchange.Batch.nulls`: `BitSet` → `long[]` (plus a length, if not implied by `recordCount`).
- Definition-level decoding / column workers: emit `long[]` directly instead of populating a `BitSet`.
- `FlatRowReader.flatNulls`: `BitSet[]` → `long[][]`. Replace `flatNulls[col].get(row)` with the inlined form:

  ```java
  (flatNulls[col][row >>> 6] & (1L << row)) != 0L
  ```

- Replace the `NO_NULLS` sentinel (`new BitSet(0)`) with a shared empty `long[]` (or `null` plus an explicit branch), whichever benchmarks better.

The win only materializes if the producer side is changed too — wrapping at the boundary with `BitSet.toLongArray()` would add a per-batch copy with no payoff.

## Acceptance

- All `RowReader` accessors and `isNull` use direct `long[]` indexing.
- Microbenchmark on a wide, mostly-non-null flat schema shows no regression and (ideally) a measurable improvement on `getX`/`isNull`.
- No behavior change visible to public API consumers.

gunnarmorling:
Probably worth doing this only once #450 has been merged due to the affected parts of the code base.

Issue #470, opened 2026-05-12
Make PqList typed accessors return List/PqIntList instead of Iterable


gunnarmorling:
## Problem

`PqList`'s typed accessors all return `Iterable<T>`:

```java
Iterable<Integer> ints();
Iterable<LocalDate> dates();
Iterable<PqStruct> structs();
Iterable<PqIntList> intLists();
// ...etc
```

Two issues with this shape:

1. **Ergonomics for reference-type accessors.** `Iterable<T>` is the minimal contract — no `size()`, no `get(int)`, no `.stream()`, can't be passed to APIs that take `List<T>`. The underlying raw values are already materialised in memory, so the laziness `Iterable` preserves buys very little; users routinely want `List` ergonomics. Compare:

   ```java
   // today
   List<LocalDate> dates = new ArrayList<>();
   row.getList("dates").dates().forEach(dates::add);

   // proposed
   List<LocalDate> dates = row.getList("dates").dates();
   ```

2. **Boxing on primitive accessors.** `ints()` / `longs()` / `doubles()` return `Iterable<Integer>` / `Iterable<Long>` / `Iterable<Double>` — they box per element. The project already has specialized [PqIntList](../blob/main/core/src/main/java/dev/hardwood/row/PqIntList.java) / `PqLongList` / `PqDoubleList` types reachable from `RowReader.getListOfInts()` and `PqList.intLists()`. They expose `PrimitiveIterator.OfInt` / `OfLong` / `OfDouble`, `int get(int)`, `int[] toArray()`, etc. — strictly better than the boxed form.

   A user calling `getList("scores").ints()` on a top-level `list<int>` boxes per element, with no way to get the primitive view short of switching to `getListOfInts("scores")` — but if you already hold a generic `PqList` (e.g. as a struct field, or projected dynamically), there's no escape hatch.

## Proposed shape

Two coordinated changes:

**a) Return `List<T>` instead of `Iterable<T>` for reference-type accessors:**

| accessor | before | after |
|---|---|---|
| `values()` | `Iterable<Object>` | `List<Object>` |
| `rawValues()` | `Iterable<Object>` | `List<Object>` |
| `strings()` | `Iterable<String>` | `List<String>` |
| `binaries()` | `Iterable<byte[]>` | `List<byte[]>` |
| `dates()` | `Iterable<LocalDate>` | `List<LocalDate>` |
| `times()` | `Iterable<LocalTime>` | `List<LocalTime>` |
[...]

Issue #469, opened 2026-05-12
Hoist typed-iterator validation on PqList


gunnarmorling:
## Problem

`PqList`'s typed iterators (`ints` / `longs` / `floats` / `doubles` / `booleans` / `strings` / `binaries` / `dates` / `times` / `timestamps` / `decimals` / `uuids` / `intervals`) currently revalidate the element schema on every iteration step inside the per-element lambda:

```java
public Iterable<LocalDate> dates() {
    return () -> new LeafIterator<>(raw -> ValueConverter.convertToDate(raw, elementSchema));
    //                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    //                                       validates `elementSchema` is a DateType on every element
}
```

The element schema is fixed for the lifetime of the list, so the per-element validation is wasted work. Two practical effects:

1. **Per-element overhead** in the hot loop — `validateLogicalType` / `validatePhysicalType` runs N times for an N-element list, every iteration.
2. **Lazy failure on type mismatch** — a wrong-type call like `pqList.dates()` on a `List<INTERVAL>` only throws when the first element is decoded. For an empty list the type mismatch is silently masked.

## Proposed shape

Hoist the validation to the iterator factory:

```java
public Iterable<LocalDate> dates() {
    ValueConverter.validateLogicalType(elementSchema, LogicalType.DateType.class);   // once
    return () -> new LeafIterator<>(raw -> raw == null
            ? null : ValueConverter.convertLogicalType(raw, elementSchema, LocalDate.class));
}
```

Per-element work shrinks to the cast / decode that always had to happen. Wrong-type access throws at the `dates()` call rather than from the iterator.

Requires promoting `ValueConverter.convertLogicalType` from private to package-private so iterator factories can reach the unchecked decode path after their one-time validation.

## Out of scope

- Flat-side validation alignment on `FlatRowReader`'s typed-decode accessors — separate concern, gated on benchmarks, tracked in #460.
- `PqStruct` / `PqMap.Entry` per-call validation — one-shot per call, not a loop; already field-named via `ValueConverter`.

Issue #466, opened 2026-05-12
Dial back PqMap.Entry typed key accessors to the high-frequency types


gunnarmorling:
## Problem

`PqMap.Entry` exposes nine typed key accessors: `getStringKey` / `getIntKey` / `getLongKey` / `getBinaryKey` / `getDateKey` / `getTimeKey` / `getTimestampKey` / `getDecimalKey` / `getUuidKey`. Real-world Parquet map usage is wildly skewed toward a handful of those — almost all maps in practice are `String`-keyed (Avro maps are string-keyed by spec), with occasional `Int` / `Long` for ID-keyed maps and rare `Binary` for hash digests.

The other five — `Date` / `Time` / `Timestamp` / `Decimal` / `UUID` keyed maps — exist as schemas you might encounter once a year. They occupy method-slot real estate in the public API for a hit rate near zero. Users who do hit them still have `getKey()` (decoded) and `getRawKey()` (raw) as escape hatches.

The asymmetry between key and value accessors is itself a real-world fact: `PqMap.Entry`'s value-side accessors legitimately span every column type because any column can be a value, but keys cluster on identifiers. The dial-back reflects that.

## Proposed shape

Keep on `PqMap.Entry`:
- `getStringKey()`
- `getIntKey()`
- `getLongKey()`
- `getBinaryKey()`

Remove from `PqMap.Entry`:
- `getDateKey()`
- `getTimeKey()`
- `getTimestampKey()`
- `getDecimalKey()`
- `getUuidKey()`

Update the `PqMap.Entry` JavaDoc to explain the asymmetry — value accessors cover every Parquet type, key accessors cover the common-in-practice subset; `getKey()` / `getRawKey()` cover the long tail.

## Out of scope

- Reader-level validation of unsupported key types — separate from API surface trimming.
- `getFloatKey` / `getDoubleKey` / `getBooleanKey` were never added (NaN-equality and boolean-domain pathologies make those poor map keys); not changing.
- `getIntervalKey` was never added; `INTERVAL` is conceivable as a key but follows the same long-tail logic — fall back to `getKey()`.

PR #471, opened 2026-05-12
#470 Switch PqList typed accessors to List/PqIntList

gunnarmorling:
Bundle of row-reader API improvements landing after the holistic review pass.

- **#464** — Variant accessors in repeated contexts (`PqList.variants()`, `PqMap.Entry.getVariantValue()`, `RowReader.getVariant(int)`); unshredded only in lists/maps.
- **#469** — `PqList` typed iterators validate at construction, not on first element decode.
- **#470 (a)** — `PqList` typed accessors return `List<T>` (lazy) instead of `Iterable<T>`.
- **#470 (b)** — `ints()` / `longs()` / `doubles()` return `PqIntList` / `PqLongList` / `PqDoubleList` (no boxing); drop the redundant `RowReader.getListOfInts` etc.
- **#466** — `PqMap.Entry` typed key accessors trimmed to `String` / `int` / `long` / `byte[]`; long-tail keys via `getKey()`.
- **#463** — `PqMap.getValue(key)` / `getRawValue(key)` / `containsKey(key)` for the four high-frequency key types.
- **#472** — `toString` overrides on the flyweight types; `byte[]` via `Arrays.toString`.
- **#473** — By-index field access on `PqStruct` (and the rest of `StructAccessor`); `PqVariantObject` deliberately stays by-name only.


gunnarmorling:
Hey @rionmonster, @iifawzi, this PR irons out a few omissions/inconsistencies around nested types in the row reader API. It is a breaking change, but I think we're good to do these at this point. Better now than later. In regards to the timeline, I was hoping to do 1.0 Final next week, but I feel that may be rushing things; so I'm now leaning towards having CR1 ready for next week (Current conference where I'm having a talk about Hardwood).

core/src/test/java/dev/hardwood/PqMapLookupTest.java
```diff
@@ -0,0 +1,179 @@
...
+    void wrongKeyTypeForColumnThrows() throws Exception {
+        try (ParquetFileReader f = ParquetFileReader.open(InputFile.of(SIMPLE));
+             RowReader r = f.rowReader()) {
+            r.next();
+            PqMap m = r.getMap("attributes"); // map<string, int32>
+            // Map has String keys; lookup as int casts the byte[][] column to int[].
+            assertThatThrownBy(() -> m.getValue(42))
+                    .isInstanceOf(ClassCastException.class);
+        }
+    }
```
iifawzi, line 178: Could null be more expected instead of throwing? I'm not sure, just feeling ClassCastException is an internal detail

iifawzi (approved):

core/src/test/java/dev/hardwood/PqMapLookupTest.java
iifawzi: Or perhaps it should be IllegalArgumentException, so the behavior is consistent, given i see now we throw IllegalArgumentException in multiple places for type mismatches, but not sure as well, just thinking loud

iifawzi:
> Hey @rionmonster, @iifawzi, this PR irons out a few omissions/inconsistencies around nested types in the row reader API. It is a breaking change, but I think we're good to do these at this point. Better now than later. In regards to the timeline, I was hoping to do 1.0 Final next week, but I feel that may be rushing things; so I'm now leaning towards having CR1 ready for next week (Current conference where I'm having a talk about Hardwood).

Hey, took a quick look, looks good, and way much cleaner than before to me 

On the other side of the timeline, what are you looking for as a must(nice?)-to-have features before CR1/Final, just to have a view if I can tackle some parts before releasing, aside from the review of the public api and testing against java-parquet

You mentioned #105 already for the final. I also think #196 is nice to have as well as these would close the gaps in predicates pushdowns. Anything else you have in your mind? In the meantime, I have in mind to at least try to close #440 and #457, which seems possible for sure for CR1, if by next week you mean Monday/Sunday. 
edit: and https://github.com/hardwood-hq/hardwood/issues/444 to have logical types fully covered

core/src/test/java/dev/hardwood/PqMapLookupTest.java
gunnarmorling: Yeah, `IllegalArgumentException` would make sense. We're probably a bit inconsistent atm. In general, I tried to avoid the manual parameter type checking on hot paths for the sake of better perf.

gunnarmorling:
> Hey, took a quick look, looks good, and way much cleaner than before to me

Awesome, thanks a lot for checking it out. Also feeling much better about this part of the API now.

On the other side of the timeline, what are you looking for as a must(nice?)-to-have features before CR1/Final, just to have a view if I can tackle some parts before releasing, aside from the review of the public api and testing against java-parquet

Good question. Here's my current thinking; I think these can wait:

* #105 would be nice, but no blocker; might be too much effort for the remaining time and we don't need to rush it
* #196 same
* #440 would be nice as it seems like a quick perf win
* #457 can definitely wait for 1.1

Whereas these are more critical/blockers:

* #444 would definitely be very good to have (essentially a blocker as it affects switch on logical types)
* #75 no need to support > 2 GB files, but we should fail early and clearly rather than silently
* #256 I don't think it actually is an issue for S3 files but it would be for `ByteBufferInputFile` users
* #460 might be worth a pass if we can agree on something; otherwise, I'm ok with adjusting this later on; I don't consider specific runtime exception types to be part of the contract

> if by next week you mean Monday/Sunday.

Yeah, I think I'd try to cut the CR1 release on Monday. So I have something to talk about at Current, and we can then ship the 1.0 Final (with more bug fixes, as needed) a week or two later. Getting the final out before the end of the Month would be great, if we can.

Issue #444, opened 2026-05-09
Recognize the NullType (case 11) logical type annotation


gunnarmorling:
## Background

Parquet's `LogicalType` thrift union has a `NullType` variant at union member id `11`. It marks a column whose every value is null — e.g. PyArrow emits this for `pa.null()` columns. Today Hardwood's `LogicalTypeReader` skips this case, so a `NullType`-annotated column lands with `column.logicalType() == null`. Reads still work (every value is null and short-circuits via `isNull`), but the schema-introspection signal is lost.

## Proposed exposure

Pure metadata addition — no new accessor:

- New `LogicalType.NullType` record (added to the sealed-permits list).
- `LogicalTypeReader` switch gains `case 11 -> { skipField; yield new NullType(); }`.
- `ColumnSchema.logicalType()` now returns `NullType` for these columns, which is what callers need to render schemas faithfully (e.g. `hardwood dive`'s schema view, downstream DDL generators).
- No `getNull(name)` accessor — values stay null and existing typed getters keep working.
- `LogicalTypeConverter.convert(...)` has nothing to do (value is already null) — `NullType` falls through the existing `physicalValue == null` short-circuit.
- Avro mapping: emit `Schema.create(Schema.Type.NULL)` from `AvroSchemaConverter`.

## Acceptance criteria

- A fixture with a `pa.null()` column round-trips schema-wise: `column.logicalType()` is `LogicalType.NullType`, every row reads as null through both flat and nested accessors.
- Avro conversion of the column produces an Avro `null` schema.
- `hardwood dive`'s schema panel shows `NULL` (or equivalent) for the column. (Optional follow-up if it falls out cleanly from the existing logical-type rendering.)

## Out of scope

- No new public accessor.
- No predicate/filter changes — predicates against an all-null column are already handled by the existing null-semantics path.

PR #481, opened 2026-05-14
#444 Recognize the NullType (case 11) logical type annotation

iifawzi:
## Summary

- Recognize Parquet's `NullType` logical-type union member (id 11) in `LogicalTypeReader`. Adds `LogicalType.NullType` to the sealed permits list; `column.logicalType()` now returns `NullType` for `pa.null()`-style columns instead of `null`.
- Map `NullType` to a bare Avro `null` field in `AvroSchemaConverter`. Skip the usual `[null, T]` nullable wrap for OPTIONAL fields / list elements / map values when `T` is already NULL — Avro forbids unions with duplicate branches and `Schema.createUnion` would have thrown on any file with a `pa.null()` column.
- `LogicalTypeConverter.convert` gets a `NullType` arm that throws `IllegalStateException` to keep the exhaustive switch compiling. The top-of-method `physicalValue == null` short-circuit makes the arm unreachable for well-formed files — every row's def-level is 0 on a NULL-typed column, so the reader never produces a non-null physical value to convert.
- Fixture `null_logical_type_test.parquet` added via `tools/simple-datagen.py`.

## Test plan

- [x] `NullLogicalTypeTest` — `column.logicalType() instanceof NullType` and every row reads as null through `isNull("nothing")`.
- [x] `AvroSchemaConverterTest` — bare-NULL field schema (not `union [null, null]`), plus preventive coverage for `list<null>` and `map<string, null>`.
- [x] `AvroRowReaderTest.readNullLogicalTypeColumn` — end-to-end through `AvroReaders`: schema field is bare NULL and every materialized `GenericRecord` has `nothing == null`.
- [x] `./mvnw verify -DskipITs -pl '!s3' -Dtest='!*S3*'` passes across core / avro / cli / parquet-java-compat / integration-test / parquet-testing-runner.

## Notes

- Docs updated: `docs/content/usage.md` (one-line note on NullType recognition under the logical-type section) and `docs/content/avro.md` (paragraph on the bare-Avro-NULL mapping and list/map collapse).
- `parquet-java-compat`'s `SchemaConverter.toOriginalType` is unchanged: there's no NULL in the legacy `OriginalType` enum, so it falls through to the existing `return null` terminal — correct behavior.


gunnarmorling:
Awesome, thanks a lot, @iifawzi! Rebased (generally preferring this over merges to keep the commit history linear) and applied.

PR #484, opened 2026-05-15
#440 Hot-path FlatRowReader optimizations

iifawzi:
## Summary

Four changes targeting per-row accessor cost on the flat read path. All sit in `FlatRowReader` and its surrounding pipeline (`FlatColumnWorker`, `BatchExchange.Batch`) — no public API change.

1. **One-slot name → index cache** via reference equality in `resolveIndex(String)`. Intern'd `String` literals at the call site let the typical `isNull("X")` + `getX("X")` pair skip the hash + probe on the second call.
2. **`ALL_PRESENT` sentinel → `null`** for `flatValidity[col]`. REQUIRED-column accessors short-circuit on a single null compare instead of touching a pre-filled `BitSet`'s word array on every row.
3. **One-slot `isNull` cache** keyed on `(column, row)`. Accessors reached via an explicit `isNull` check no longer re-evaluate the bitmap inside `getX(int)`.
4. **Pre-allocated `validityBuffer` on `Batch`** eliminates the `currentValidity.clone()` per published batch in `FlatColumnWorker`. The worker writes directly into the recyclable buffer and flips `Batch.validity` between the buffer and `null` depending on whether any absent was observed.

## Benchmark numbers

`RecordFilterBenchmarkTest`, 10M rows × 5 cols (4 REQUIRED + 1 OPTIONAL `score`), 5 runs each, **median of runs 2–5** (dropping the first-run JIT outlier). Times in ms (lower is better).

### Per-row API hot path

| Scenario | Upstream | Branch | Δ |
|---|---:|---:|---:|
| Scan nullable, 1% null | 92.7 | 69.2 | **−25%** |
| Scan nullable, 50% null | 150.6 | 122.3 | **−19%** |
| Scan nullable, 90% null | 85.5 | 67.7 | **−21%** |
| IS NOT NULL, 1% null | 54.0 | 43.0 | **−21%** |
| IS NOT NULL, 50% null | 65.0 | 55.5 | **−15%** |
| IS NOT NULL, 90% null | 32.8 | 28.5 | **−13%** |

### Filter scenarios

| Scenario | Upstream | Branch | Δ |
|---|---:|---:|---:|
| Range+value | 9.2 | 6.7 | **−27%** |
| intIn | 45.0 | 35.8 | **−20%** |
| Compound mid 50% | 70.8 | 60.8 | **−14%** |
| Sorted cluster 50% | 21.6 | 18.1 | **−16%** |
| OR fallback | 160.7 | 144.4 | **−10%** |
| And4 ~50% | 59.6 | 54.8 | **−8%** |

### Baseline scenarios (no validity / accessor work to optimize)

| Scenario | Upstream | Branch | Δ |
|---|---:|---:|---:|
| No filter | 42.6 | 43.5 | +2% |
| Match-all (id>=0) | 41.2 | 43.6 | +6% |
| And3 match-all | 44.8 | 45.1 | 0% |

All baseline deltas sit inside the run-to-run noise band on this hardware.

Bench config: `./mvnw test -Pperformance-test -pl performance-testing/end-to-end -Dtest=RecordFilterBenchmarkTest -Dperf.runs=5`. `RecordFilterBenchmarkTest` is parameterized over `{1%, 50%, 90%}` null densities for the three nullable scenarios.

## Why each change is beneficial

| Change | Targets | Per-row saving |
|---|---|---|
| #1 Name cache | name → index hash on by-name accessors | ~5–10 ns when the user reuses a literal |
| #2 Null sentinel | REQUIRED-column bitmap touch | one fewer memory load + branch-predictor-friendly |
| #3 `isNull` cache | redundant bitmap check inside `getX` when paired with explicit `isNull` | ~3–5 ns |
| #4 No clone | per-publish `BitSet.clone()` in the producer | amortized GC pressure (~5.6 MB / 10M-row file × nullable columns) |

The biggest single contributor on the `Scan nullable column` shape is the name-cache (the bench loop hits two by-name calls per row with the same literal); #2 dominates the REQUIRED-column wins.

## Test plan

- [ ] `./mvnw verify -pl '!s3'` (1078 core tests pass on this branch)
- [ ] `./mvnw test -Pperformance-test -pl performance-testing/end-to-end -Dtest=RecordFilterBenchmarkTest -Dperf.runs=5`
- [ ] No public API change; `FieldAccessor` / `RowReader` surfaces unchanged
- [ ] `ColumnWorkerTest.flatPipelineTracksNulls` updated to pre-allocate `validityBuffer` in its factory (the production factories in `FlatRowReader.create` and `ColumnReader.createFromIterator` do this for nullable columns)

### DIFF
--- a/core/src/main/java/dev/hardwood/internal/reader/BatchExchange.java
+++ b/core/src/main/java/dev/hardwood/internal/reader/BatchExchange.java
@@ -52,7 +52,18 @@ public class BatchExchange<B> {
     /// types.
     public static final class Batch {
         public Object values;
+        /// Public per-batch validity bitmap. Set-bit-= -present polarity; `null` is
+        /// the sparse "every leaf present" form. When non-null, references
+        /// [#validityBuffer]; consumers must treat it as read-only.
         public BitSet validity;
+        /// Pre-allocated scratch BitSet used by the drain to record per-row
+        /// presence without per-batch allocation. Allocated once by the
+        /// [BatchExchange] factory for nullable columns; `null` for REQUIRED
+        /// columns. The drain calls `validityBuffer.clear()` between batches and
+        /// switches it on lazily when the first absent is observed. At publish
+        /// time, [#validity] is assigned to this buffer iff at least one absent
+        /// was recorded; otherwise [#validity] is set to `null`.
+        public BitSet validityBuffer;
         public int recordCount;
         public String fileName;
         /// Per-batch matches mask, populated by [dev.hardwood.internal.predicate.ColumnBatchMatcher]
--- a/core/src/main/java/dev/hardwood/internal/reader/FlatColumnWorker.java
+++ b/core/src/main/java/dev/hardwood/internal/reader/FlatColumnWorker.java
@@ -21,7 +21,6 @@ import dev.hardwood.schema.ColumnSchema;
 /// and null tracking via [BitSet].
 public class FlatColumnWorker extends ColumnWorker<BatchExchange.Batch> {
 
-    private BitSet currentValidity;
     private final ColumnBatchMatcher columnFilter;
     /// Tracks whether any absent (null) leaf has been seen in the current
     /// batch; cleared by [#publishCurrentBatch]. When still false at publish
@@ -54,7 +53,8 @@ public class FlatColumnWorker extends ColumnWorker<BatchExchange.Batch> {
 
     @Override
     void initDrainState() {
-        currentValidity = maxDefinitionLevel > 0 ? new BitSet(batchCapacity) : null;
+        // currentBatch.validityBuffer is pre-allocated by the BatchExchange factory
+        // for nullable columns; the worker only resets the hasAbsents flag here.
         currentBatchHasAbsents = false;
     }
 
@@ -123,9 +123,11 @@ public class FlatColumnWorker extends ColumnWorker<BatchExchange.Batch> {
             return;
         }
         currentBatch.recordCount = rowsInCurrentBatch;
-        currentBatch.validity = (currentValidity != null && currentBatchHasAbsents)
-                ? (BitSet) currentValidity.clone()
-                : null;
+        // Publish the pre-allocated validityBuffer directly when any absent was
+        // recorded; otherwise publish `null` so matchers and the consumer skip the
+        // validity-handling path entirely. No clone — the buffer belongs to this
+        // Batch and is recycled with it.
+        currentBatch.validity = currentBatchHasAbsents ? currentBatch.validityBuffer : null;
         currentBatch.fileName = currentBatchFileName;
 
         if (columnFilter != null) {
@@ -155,8 +157,8 @@ public class FlatColumnWorker extends ColumnWorker<BatchExchange.Batch> {
         batchesPublished++;
 
         rowsInCurrentBatch = 0;
-        if (currentValidity != null) {
-            currentValidity.clear();
+        if (currentBatch.validityBuffer != null) {
+            currentBatch.validityBuffer.clear();
         }
         currentBatchHasAbsents = false;
     }
@@ -222,12 +224,13 @@ public class FlatColumnWorker extends ColumnWorker<BatchExchange.Batch> {
     /// bitmap if it was already switched on by an earlier absent in the
     /// batch.
     private void markNulls(int[] defLevels, int srcPos, int destPos, int length) {
-        if (currentValidity == null) {
+        BitSet validity = currentBatch.validityBuffer;
+        if (validity == null) {
             return;
         }
         if (defLevels == null) {
             if (currentBatchHasAbsents) {
-                currentValidity.set(destPos, destPos + length);
+                validity.set(destPos, destPos + length);
             }
             return;
         }
@@ -235,11 +238,11 @@ public class FlatColumnWorker extends ColumnWorker<BatchExchange.Batch> {
             if (defLevels[srcPos + i] < maxDefinitionLevel) {
                 if (!currentBatchHasAbsents) {
                     currentBatchHasAbsents = true;
-                    currentValidity.set(0, destPos + i);
+                    validity.set(0, destPos + i);
                 }
             }
             else if (currentBatchHasAbsents) {
-                currentValidity.set(destPos + i);
+                validity.set(destPos + i);
             }
         }
     }
--- a/core/src/main/java/dev/hardwood/internal/reader/FlatRowReader.java
+++ b/core/src/main/java/dev/hardwood/internal/reader/FlatRowReader.java
@@ -141,11 +141,13 @@ public final class FlatRowReader implements RowReader {
             columnSchemas[i] = col;
         }
 
-        // Sized to next-power-of-2 ≥ 2 × columnCount (min 4) so the direct-mapped
-        // cache stays sparse (< 50% load) — keeps collisions rare for typical
+        // Sized to next-power-of-2 ≥ 2 × columnCount so the direct-mapped cache
+        // stays sparse (< 50% load) — keeps collisions rare for typical
         // projections while costing only a few dozen bytes for wide schemas.
-        int desired = Math.max(columnCount * 2, 4);
-        int slots = Integer.highestOneBit(desired - 1) << 1;
+        // The `columnCount == 0` branch keeps slot arithmetic well-defined for
+        // empty projections (no names will ever be resolved, but the field
+        // initializers must not divide by zero).
+        int slots = columnCount == 0 ? 1 : Integer.highestOneBit(columnCount * 2 - 1) << 1;
         this.nameCacheSlotMask = slots - 1;
         this.nameCacheNames = new String[slots];
         this.nameCacheIndices = new int[slots];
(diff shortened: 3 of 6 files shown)


iifawzi:
Hey @gunnarmorling, this's another great improvement e2e that I realized after experimenting with #440, switching to long[] everywhere didn't yield any significant e2e speedups, which makes me realize bitSet was never the issue itself. 

This PR mainly addresses the multiple name-to-index conversions on the name accessors by caching the latest index, so subsequent usage of retrieve index on the same column doesn't do two lookups. 

The other caching point is in the null check, as today the null check is done twice (we need to think if we can improve that), so for instance everywhere [in the compiler](https://github.com/hardwood-hq/hardwood/blob/e9b9afba43b9fc0eec551cf0b9b97acbf5f95a78/core/src/main/java/dev/hardwood/internal/predicate/RecordFilterCompiler.java#L315-L370) there's an isNull check, and then in getX it is called again. This cache eliminates the duplicate check and returns the latest result. 

The other two are the first one, improving the null sentinel to avoid checking the bitset each time, and using direct null instead, and the last one is to decrease the overhead of cloning per batch, not a speedup improvement but a memory suggestion that claude suggested and seemed useful

iifawzi:
Another nice suggestion from Claude is to increase the cache size, showing another significant win against the 1-slot cache

CLAUDE EXPLANATION: 
---
Pushed an upgrade to change #1 (commit bd328a3): the name cache is now a direct-mapped table sized to next-pow-2 ≥ 2 × `columnCount` (min 4 slots), keyed on `System.identityHashCode(name) & slotMask`.

**Why:** the previous one-slot cache hit the `isNull("X") + getX("X")` pattern but missed on multi-column row materialization where consecutive accessor calls pass different names — each call evicts the previous slot, so hit rate drops to 0% for that shape.

**Trade-off:** the multi-slot version gains on multi-column accessor patterns at a small cost to the one-name pattern. Per-call overhead is one `identityHashCode` call (header read, cached on the `String`) + one array load + one reference compare, vs the previous single field load + compare.

### Updated end-to-end numbers (median runs 2–5 vs upstream/main)

| Scenario | Upstream | 1-slot (prev) | Multi-slot (new) |
|---|---:|---:|---:|
| **And3 match-all (id+value+tag)** | 44.8 | 45.1 (+1%) | **35.0 (−22%)** |
| **Match-all (id>=0)** | 41.2 | 43.6 (+6%) | **38.2 (−7%)** |
| **And4 ~50%** | 59.6 | 54.8 (−8%) | **52.2 (−12%)** |
| Sorted cluster 50% | 21.6 | 18.1 (−16%) | 16.9 (−22%) |
| Range+value | 9.2 | 6.7 (−27%) | 6.3 (−31%) |
| OR fallback | 160.7 | 144.4 (−10%) | 141.3 (−12%) |
| Scan nullable, 1% null | 92.7 | **69.2 (−25%)** | 75.8 (−18%) |
| Scan nullable, 50% null | 150.6 | **122.3 (−19%)** | 137.1 (−9%) |
| Scan nullable, 90% null | 85.5 | 67.7 (−21%) | 65.9 (−23%) |
| NOT NULL, 90% null | 32.8 | **28.5 (−13%)** | 32.1 (−2%) |

### Reading

- **Multi-column accessor patterns** (And3, Match-all, And4): multi-slot wins decisively because the 1-slot version was evicting on every call.
- **isNull-then-get patterns** (Scan nullable, NOT NULL): the 1-slot version was already near-optimal for this shape (50% hit rate within a column); the multi-slot adds the identity-hash overhead without buying more hits, so it's slightly worse but still well below upstream.
- Net result: every scenario stays below upstream, and the previously-flat multi-column scenarios now show real wins.

### Note on the `isNull` cache (change #3)

Considered broadening that one to multi-slot too, but the dominant traffic for `isNull` is the immediately-followed-by-`getX` pattern within a column — a one-slot cache already hits 50% which is the ceiling for that shape. Multi-slot would only help interleaved patterns (`isNull("a") isNull("b") isNull("c") getX("a") getX("b") getX("c")`), which is rare. The per-call overhead would leak into the dominant shape (same lesson the name-cache trade-off shows). Keeping `isNull` as one slot.

gunnarmorling:
Ah, that's really cool! Will check it out over the weekend. Could you share numbers from `FlatPerformanceTest`, Hardwood named contender, before and after this change?

iifawzi:
> Ah, that's really cool! Will check it out over the weekend. Could you share numbers from `FlatPerformanceTest`, Hardwood named contender, before and after this change?

Nice to know about that, and in fact, it's showing a huge regression 

 Branch │ Avg time │ Records/sec │ MB/sec 
 main     │ 0.78 s     │ 52.6 M/s        │ 844 
 484      │ 1.16 s      │ 35.6 M/s        │ 572 
 
will take a look

gunnarmorling:
Yeah, `FlatPerformanceTest` and `NestedPerformanceTest` are currently the two end-to-end benchmarks we have, and we should make sure there are no regressions (there's also an issue to set up Apache Otava for continuous tracking). It's just a very simple baseline though, e.g. not exercising filtering or the S3 path.

iifawzi:
> Yeah, `FlatPerformanceTest` and `NestedPerformanceTest` are currently the two end-to-end benchmarks we have, and we should make sure there are no regressions (there's also an issue to set up Apache Otava for continuous tracking). It's just a very simple baseline though, e.g. not exercising filtering or the S3 path.

That's super nice, I was looking for a full comparison against java-parquet, seeing the 5~6x speedup is impressive! 

On the other side, I've taken a look at, the overhead of the cache, which is significant when the hit ratio is 0, hence why the regression on the flat test. so optimizing both scenarios with one shot might not be realistic, I will check if we can move the cache to the filtered path / or if we can omit the duplicated null/index checks on that path 

drafting for now, until i can come with better proposal

iifawzi:
I was now curious to run the flat benchmark against the initial attempt to convert bitSet to long[] and it has shwon an improvement on flat, while a regression on filtering (recordFilterBenchmarkTest), so I will try to find the best of both worlds 

Using words ([commit](https://github.com/hardwood-hq/hardwood/commit/32eb629593f6ec439125740b6b1d7b93d83c59ca)) 
```
PERFORMANCE TEST RESULTS

Environment:
  CPU cores:       10
  Java version:    25.0.3

Data:
  Files processed: 119
  Total rows:      651,209,003
  Total size:      9,241.1 MB
  Runs per contender: 10

Performance (all runs):
  Hardwood (row reader named) [1]        11.95      54,494,477          5,449,448        773.3
  Hardwood (row reader named) [2]        11.94      54,535,550          5,453,555        773.9
  Hardwood (row reader named) [3]        11.93      54,581,259          5,458,126        774.5
  Hardwood (row reader named) [4]        11.96      54,439,810          5,443,981        772.5
  Hardwood (row reader named) [5]        11.93      54,581,259          5,458,126        774.5
  Hardwood (row reader named) [6]        11.93      54,590,410          5,459,041        774.7
  Hardwood (row reader named) [7]        12.14      53,654,857          5,365,486        761.4
  Hardwood (row reader named) [8]        12.10      53,818,926          5,381,893        763.7
  Hardwood (row reader named) [9]        12.08      53,916,957          5,391,696        765.1
  Hardwood (row reader named) [10]        11.99      54,299,091          5,429,909        770.5
  Hardwood (row reader named) [AVG]        12.00      54,290,038          5,429,004        770.4
                                   min: 11.93s, max: 12.14s, spread: 0.21s

[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 156.2 s -- in dev.hardwood.perf.FlatPerformanceTest
```

Main: 
```
PERFORMANCE TEST RESULTS

Environment:
  CPU cores:       10
  Java version:    25.0.3

Data:
  Files processed: 119
  Total rows:      651,209,003
  Total size:      9,241.1 MB
  Runs per contender: 10

Performance (all runs):
  Hardwood (row reader named) [1]        12.31      52,887,924          5,288,792        750.5
  Hardwood (row reader named) [2]        12.37      52,627,202          5,262,720        746.8
  Hardwood (row reader named) [3]        12.30      52,926,609          5,292,661        751.1
  Hardwood (row reader named) [4]        12.32      52,879,334          5,287,933        750.4
  Hardwood (row reader named) [5]        12.63      51,580,911          5,158,091        732.0
  Hardwood (row reader named) [6]        12.29      52,973,969          5,297,397        751.7
  Hardwood (row reader named) [7]        12.29      52,999,837          5,299,984        752.1
  Hardwood (row reader named) [8]        12.36      52,699,604          5,269,960        747.8
  Hardwood (row reader named) [9]        12.29      52,969,660          5,296,966        751.7
  Hardwood (row reader named) [10]        12.31      52,892,219          5,289,222        750.6
  Hardwood (row reader named) [AVG]        12.35      52,742,286          5,274,229        748.4
                                   min: 12.29s, max: 12.63s, spread: 0.34s

[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 161.2 s -- in dev.hardwood.perf.FlatPerformanceTest

```

iifawzi:
Closing this PR in favor of switching BitSet to long, I rushed decisions in this PR and haven't verified claude claims, which appear to be all false.  doing other iterations on top of the validity-to-long [commit](https://github.com/hardwood-hq/hardwood/commit/32eb629593f6ec439125740b6b1d7b93d83c59ca), is showing better (and safe) performance gains, compared to the caching mechanism being applied here, where if the cache hit ratio is low, it breaks all performance gains massively. 

I will create another PR shortly with the bitSet to long changes. Apologies for the noise brought by this PR @gunnarmorling , and thanks for pointing to the performance test, this made me realize a lot of false assumptions I had, after verifying everything manually

PR #489, opened 2026-05-17
#438 To enforce some rules from CLAUDE.md

### LINKED ISSUE #438: feat(build): enforce no-var rule mechanically
## Motivation

CLAUDE.md (project coding rules) prohibits Java `var`:

> Never use `var` syntax.

This is enforced today by reviewer attention only. It slipped past me twice on the same branch (commits `287fbb3` and `a3ee415`), caught only on a follow-up review pass and corrected in `b633372`. A mechanical check would have flagged both at write time.

## Proposal

Add a build-time check that fails the build (or at least produces a hard warning) when `var` appears in `*.java` under `*/src/{main,test}/java/`. A few low-cost shapes:

- **A grep-style enforcer plugin step** — quickest, no new tooling. The `formatter-maven-plugin` already runs over the source tree; an extra `exec-maven-plugin` execution that greps for `\bvar\b\s+[a-zA-Z_]` across `**/src/{main,test}/java/**/*.java` and fails on hits is ~15 lines of POM config. Caveat: false positives if `var` ever appears as part of an identifier or in a string literal — manageable with a slightly more careful regex (`(?<![a-zA-Z0-9_$])var\s+[a-zA-Z_]`).
- **An ErrorProne or NullAway-style static analyzer with a custom check** — heavier setup, but lets us enforce other CLAUDE.md rules at the same time (no fully-qualified class names within code, no unsafe downcasts, etc.).
- **A Checkstyle rule** — Checkstyle has a `LocalVariableName`/regex-based check. Lightest of the proper static-analysis options.

I'd start with the grep enforcer — fastest path to closing the gap, easy to remove if a richer tool replaces it later.

## Related rules worth enforcing the same way

If we go beyond just `var`, CLAUDE.md has several other mechanically-checkable rules:

- "Avoid fully-qualified class names within the code, always add imports" — grep for `\b(?:[a-z]+\.){2,}[A-Z]` outside import statements.
- JavaDoc must use `///` Markdown syntax, not `/** */` — grep for `/\*\*` in source files (excluding Javadoc-aware tools, but it's a block comment, so easy).
- "GitHub Actions should always be referenced by SHA" — grep `.github/workflows/*.yml` for `uses:.*@v\d` and similar non-SHA refs.

These are bonus; the no-var check is the immediate motivation.

## Out of scope

Pre-existing `var` usages in tests (the ones in `MapSchemaTest.java` that pre-date the recent #435 work, for instance). The check should fail on *new* uses but a one-time cleanup of existing sites can happen separately or as part of the rule-introduction commit — caller's choice.

## Acceptance

- A build step that fails when `var` is used in production or test Java code.
- The failure message points at the offending file:line and the CLAUDE.md rule.
- `./mvnw verify` continues to succeed on the current main (after any pre-existing-var cleanup).

polo7:
Closes #438 

## Checklist

- [x] The commit message is prefixed with the issue key ("#438 To enforce some rules from CLAUDE.md")

I implemented:
- enforcement through ErrorProne. "no-var" and "/// javadoc" checks are in place. 
- clean up of existing `var`

I didn't implement:
- "Avoid fully-qualified class names" because I have doubts since two similar classes are possible and one of them may use fully qualified name
- "GitHub Actions should always be referenced by SHA" because YML files are out of scope of ErrorProne (I don't have extensive experience with ErrorProne, correct me if I am wrong).

### DIFF
--- a/.github/workflows/cli-early-access.yml
+++ b/.github/workflows/cli-early-access.yml
@@ -70,7 +70,7 @@ jobs:
           path: ~/.m2/repository/dev/hardwood/
 
       - name: 'Build Native Image'
-        run: ./mvnw -ntp -B -Dnative package -pl cli -DskipTests
+        run: ./mvnw -ntp -B -Dnative package -pl :hardwood-error-prone-checks,:hardwood-cli -DskipTests
 
       - name: 'Upload build artifact (tar.gz)'
         if: ${{ runner.os != 'Windows' }}
--- a/.github/workflows/main-build.yml
+++ b/.github/workflows/main-build.yml
@@ -51,7 +51,7 @@ jobs:
           fi
 
       - name: 'Build and install BOMs, test-support, core and avro'
-        run: ./mvnw -B clean install -pl :hardwood-bom,:hardwood-test-bom,:hardwood-test-support,:hardwood-core,:hardwood-avro
+        run: ./mvnw -B clean install -pl :hardwood-error-prone-checks,:hardwood-bom,:hardwood-test-bom,:hardwood-test-support,:hardwood-core,:hardwood-avro
 
       - name: 'Upload built artifacts'
         uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f  # v7 — https://github.com/actions/upload-artifact/releases/tag/v7
@@ -108,7 +108,7 @@ jobs:
           path: ~/.m2/repository/dev/hardwood/
 
       - name: 'Build and install s3 and aws-auth'
-        run: ./mvnw -B clean install -pl :hardwood-s3,:hardwood-aws-auth
+        run: ./mvnw -B clean install -pl :hardwood-error-prone-checks,:hardwood-s3,:hardwood-aws-auth
 
       - name: 'Upload built artifacts'
         uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f  # v7 — https://github.com/actions/upload-artifact/releases/tag/v7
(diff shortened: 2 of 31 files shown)


polo7:
I faced a problem during CI/CD process:
```
Warning:  The POM for dev.hardwood:hardwood-error-prone-checks:jar:1.0.0-SNAPSHOT is missing, no dependency information available
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.14.1:compile (default-compile) on project hardwood-test-support: Resolution of annotationProcessorPath dependencies failed: The following artifacts could not be resolved: dev.hardwood:hardwood-error-prone-checks:jar:1.0.0-SNAPSHOT (absent): Could not find artifact dev.hardwood:hardwood-error-prone-checks:jar:1.0.0-SNAPSHOT -> [Help 1]
Error:  
Error:  To see the full stack trace of the errors, re-run Maven with the -e switch.
Error:  Re-run Maven using the -X switch to enable full debug logging.
Error:  
Error:  For more information about the errors and possible solutions, please read the following articles:
Error:  [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
Error:  
Error:  After correcting the problems, you can resume the build with the command
Error:    mvn <args> -rf :hardwood-test-support
Error: Process completed with exit code 1.
```

I added workaround but I am not sure if you are OK with it. Please let me know if we can keep it or better switch to grep based solution.

gunnarmorling:
Thanks a lot, @polo7! Sorry for the delay, I was out for Current London last week, and then a few other things came up. Looking into this PR right now; already pushed one more commit to adjust a few things.

gunnarmorling:
Thanks again, this LGTM overall. I'll take it from here. Already merged a commit with your `var` clean-up (keeping your authorship), and am working on a few follow-ups via #499. Closing this one here in favor of the follow-up.

PR #490, opened 2026-05-18
#440 word shape validity

### LINKED ISSUE #440: Replace BitSet null bitmaps with raw long[] on the read hot path

iifawzi:
Hello, this PR resolves #440, replacing bitSet with words. There are no major gains here, so I'm opening the PR to spin the discussions about whether you find it better or not. In my opinion, given it's not showing a regression, it looks good to me, as it's showing a win for columns with a lot of nulls, as it unlocks the usage of numberOfTrailingZeros in iterations to efficiently find the next set bit

```
  int wordCount = (count + 63) >>> 6;
            for (int w = 0; w < wordCount; w++) {
                long present = words[w];
                while (present != 0L) {
                    int bit = Long.numberOfTrailingZeros(present);
                    sum += values[(w << 6) + bit];
                    present &= present - 1L;
                }
            }
```
Verifying that with the newly added benchmark `ValidityIterationBenchmark`  showing 8x speed on such columns, yet 2x slowdown for columns with 5% nulls, which is expected given the null sparse columns we will pay the overhead of numberOfTrailingZeros several times.

 However, with this PR, the per-item iteration mechanism is still available to use, and benchmarks show it's even 25-40% faster, so it depends on consumers' awareness of their columns; if they have a high rate of nulls, they can go with the word-backed, if not, they can use the per-item check 

```

  Three knobs: row count (1024 vs 65536) and null density (0% / 5% / 95%). Score is avg ns/op, lower is better.

  ┌───────────────┬─────────────┬───────┬────────┐
  │    Density    │  Strategy   │ 1024  │ 65536  │
  ├───────────────┼─────────────┼───────┼────────┤
  │ 0% (NO_NULLS) │ perItem     │ 303   │ 21,023 │
  ├───────────────┼─────────────┼───────┼────────┤
  │ 0%            │ wordIterate │ 309   │ 20,920 │
  ├───────────────┼─────────────┼───────┼────────┤
  │ 5% nulls      │ perItem     │ 555   │ 36,152 │
  ├───────────────┼─────────────┼───────┼────────┤
  │ 5% nulls      │ wordIterate │ 1,246 │ 80,036 │
  ├───────────────┼─────────────┼───────┼────────┤
  │ 95% nulls     │ perItem     │ 514   │ 33,367 │
  ├───────────────┼─────────────┼───────┼────────┤
  │ 95% nulls     │ wordIterate │ 44    │ 4,088  │
  └───────────────┴─────────────┴───────┴────────┘

```

Comparing per-item check on main vs. words-backed validity

```
┌───────┬─────────────┬───────────────┬─────────────────────┬─────────┐
  │ count │ nullDensity │ main (BitSet) │ branch #440 (words) │ speedup │
  ├───────┼─────────────┼───────────────┼─────────────────────┼─────────┤
  │ 1024  │ 0.00        │ 307           │ 308                 │ ~same   │
  ├───────┼─────────────┼───────────────┼─────────────────────┼─────────┤
  │ 1024  │ 0.05        │ 692           │ 551                 │ ~1.26×  │
  ├───────┼─────────────┼───────────────┼─────────────────────┼─────────┤
  │ 1024  │ 0.95        │ 704           │ 510                 │ ~1.38×  │
  ├───────┼─────────────┼───────────────┼─────────────────────┼─────────┤
  │ 65536 │ 0.00        │ 21198         │ 21045               │ ~same   │
  ├───────┼─────────────┼───────────────┼─────────────────────┼─────────┤
  │ 65536 │ 0.05        │ 50628         │ 36628               │ ~1.38×  │
  ├───────┼─────────────┼───────────────┼─────────────────────┼─────────┤
  │ 65536 │ 0.95        │ 47814         │ 33664               │ ~1.42×  │
  └───────┴─────────────┴───────────────┴─────────────────────┴─────────┘

```

# Benchmarks: 

E2E Benchmark differences fall within noise, no speedup or slowdowns noticed across several runs, sub-second differences that are sometimes in favor of one main branch, and sometimes are for this branch. 

Before: 

```
PERFORMANCE TEST RESULTS

Environment:
  CPU cores:       10
  Java version:    25.0.3

Data:
  Files processed: 119
  Total rows:      651,209,003
  Total size:      9,241.1 MB
  Runs per contender: 10

Correctness Verification:
                              passenger_count     trip_distance       fare_amount
  Hardwood (row reader indexed)       972,078,547  2,701,223,013.48  9,166,943,759.83
  Hardwood (column reader)        972,078,547  2,701,223,013.48  9,166,943,759.83

Performance (all runs):
  Hardwood (row reader indexed) [1]         3.23     201,550,295         20,155,029       2860.1
  Hardwood (row reader indexed) [2]         3.23     201,737,609         20,173,761       2862.8
  Hardwood (row reader indexed) [3]         3.38     192,950,816         19,295,082       2738.1
  Hardwood (row reader indexed) [4]         3.20     203,248,752         20,324,875       2884.2
  Hardwood (row reader indexed) [5]         3.23     201,612,694         20,161,269       2861.0
  Hardwood (row reader indexed) [6]         3.28     198,721,087         19,872,109       2820.0
  Hardwood (row reader indexed) [7]         3.29     198,116,521         19,811,652       2811.4
  Hardwood (row reader indexed) [8]         3.18     204,653,992         20,465,399       2904.2
  Hardwood (row reader indexed) [9]         3.18     204,525,441         20,452,544       2902.4
  Hardwood (row reader indexed) [10]         3.20     203,821,284         20,382,128       2892.4
  Hardwood (row reader indexed) [AVG]         3.24     201,052,486         20,105,249       2853.1
                                   min: 3.18s, max: 3.38s, spread: 0.19s

  Hardwood (column reader) [1]           2.27     287,255,846         28,725,585       4076.4
  Hardwood (column reader) [2]           2.25     289,169,184         28,916,918       4103.5
  Hardwood (column reader) [3]           2.56     254,179,939         25,417,994       3607.0
  Hardwood (column reader) [4]           2.31     281,786,674         28,178,667       3998.7
  Hardwood (column reader) [5]           2.25     289,169,184         28,916,918       4103.5
  Hardwood (column reader) [6]           2.27     286,623,681         28,662,368       4067.4
  Hardwood (column reader) [7]           2.28     286,119,949         28,611,995       4060.2
  Hardwood (column reader) [8]           2.22     293,866,879         29,386,688       4170.2
  Hardwood (column reader) [9]           2.22     292,941,522         29,294,152       4157.0
  Hardwood (column reader) [10]          2.29     284,122,602         28,412,260       4031.9
  Hardwood (column reader) [AVG]         2.29     284,122,602         28,412,260       4031.9
                                   min: 2.22s, max: 2.56s, spread: 0.35s

```

After: 
```
PERFORMANCE TEST RESULTS

Environment:
  CPU cores:       10
  Java version:    25.0.3

Data:
  Files processed: 119
  Total rows:      651,209,003
  Total size:      9,241.1 MB
  Runs per contender: 10

Correctness Verification:
                              passenger_count     trip_distance       fare_amount
  Hardwood (row reader indexed)       972,078,547  2,701,223,013.48  9,166,943,759.83
  Hardwood (column reader)        972,078,547  2,701,223,013.48  9,166,943,759.83

Performance (all runs):
  Hardwood (row reader indexed) [1]         3.66     177,877,357         17,787,736       2524.2
  Hardwood (row reader indexed) [2]         3.87     168,184,143         16,818,414       2386.6
  Hardwood (row reader indexed) [3]         3.35     194,506,871         19,450,687       2760.2
  Hardwood (row reader indexed) [4]         3.27     198,903,177         19,890,318       2822.6
  Hardwood (row reader indexed) [5]         3.31     196,739,880         19,673,988       2791.9
  Hardwood (row reader indexed) [6]         3.30     197,216,536         19,721,654       2798.6
  Hardwood (row reader indexed) [7]         3.30     197,635,509         19,763,551       2804.6
  Hardwood (row reader indexed) [8]         3.33     195,617,003         19,561,700       2775.9
  Hardwood (row reader indexed) [9]         3.50     186,112,890         18,611,289       2641.1
  Hardwood (row reader indexed) [10]         3.29     197,815,615         19,781,561       2807.1
  Hardwood (row reader indexed) [AVG]         3.42     190,523,406         19,052,341       2703.7
                                   min: 3.27s, max: 3.87s, spread: 0.60s

  Hardwood (column reader) [1]           2.37     274,887,718         27,488,772       3900.8
  Hardwood (column reader) [2]           2.33     279,129,448         27,912,945       3961.0
  Hardwood (column reader) [3]           2.59     251,432,047         25,143,205       3568.0
  Hardwood (column reader) [4]           2.43     268,097,572         26,809,757       3804.5
  Hardwood (column reader) [5]           2.21     294,798,100         29,479,810       4183.4
  Hardwood (column reader) [6]           2.24     291,108,182         29,110,818       4131.0
  Hardwood (column reader) [7]           2.26     288,273,131         28,827,313       4090.8
  Hardwood (column reader) [8]           2.26     288,400,798         28,840,080       4092.6
  Hardwood (column reader) [9]           2.23     292,415,358         29,241,536       4149.6
  Hardwood (column reader) [10]          2.25     289,812,640         28,981,264       4112.6
  Hardwood (column reader) [AVG]         2.32     281,299,785         28,129,979       3991.8
                                   min: 2.21s, max: 2.59s, spread: 0.38s

```

## Checklist


iifawzi:
Hey @gunnarmorling @rionmonster, let me know what you think about that when you have time. In my opinion, it's a win, although e2e is not showing great speedups, but this is due to the fact that the null check isn't a significant contribution to the latency on those tests, it seems like, where on the other side, micro benchmarks showed 25% - 40% speedup on per item check, and 8x on the word mechanism 

Having the ability to use the word backed mechanism is nice to have in my opinion, but we would be paying the price of having more binary operations (but well, it's a parser :d) 

Your thoughts are more than welcome, even if you want me to explore deeper into a specific direction related to that. I have been discovering this for a couple of days, so I might have overlooked something

gunnarmorling:
Looking now. Thanks a lot for keep pushing on everything performance, @iifawzi!

core/src/main/java/dev/hardwood/reader/Validity.java
```diff
@@ -83,6 +87,13 @@ static Validity of(BitSet bits) {
     /// length.
     int nextNotNull(int from, int count);
 
+    /// The word array (set-bit = present polarity). Returns `null` for
+    /// [NoNulls]. For [Backed], returns the backing array directly — no
+    /// copy. Callers must not mutate it; mirroring the inbound contract
+    /// on [#of(long[])], the `Validity` owns the bitmap once handed in.
+    /// Bits at indices `>= count` are undefined and must not be read.
+    long[] words();
```
gunnarmorling, line 95: Instead of exposing `long[]` here, should we expose `BitSet`? I feel using `long[]` internally makes sense, but at the public API boundary, `BitSet` might be more convenient for external users? No strong preference either way though, curious what you think.

gunnarmorling: Claude, doesn't seem to like it :)

> In the public Validity API, would it make sense to expose BitSet there?
>
> Probably not, and the PR itself is the strongest argument against it.
>
> **Why `long[]` over `BitSet` here:**
>
> 1. **It would re-introduce the copy the PR removes.** The internal carrier is now `long[]` end-to-end. Returning a `BitSet` from `getLeafValidity()` / `getLayerValidity()` means a `BitSet.valueOf(long[])` wrap on every call — a per-batch allocation + copy, which is exactly what the PR's "single representation, all the way through" property is built to eliminate.
>
> 2. **`BitSet.get(int)` isn't free.** Bounds check + word-index recompute that the JIT can't always strip. The whole point of `words()` is to let perf-sensitive callers skip both. Wrapping in `BitSet` and asking them to call `bs.toLongArray()` to recover that path round-trips a copy.
>
> 3. **`BitSet` is mutable, with no read-only view.** The PR's contract is "Validity owns the bitmap, don't mutate." `long[]` makes mutation obvious (array indexing) but at least doesn't hand callers a fluent `clear()` / `andNot()` / `flip()` surface that invites it.
>
> 4. **Polarity confusion.** With set-bit-=-present storage, `bits.get(i)` reads as "is present" — the opposite of "is null". The design doc explicitly calls out that prior code paid this tax (`!bits.get(i)` everywhere). The dedicated `isNull(i)` / `isNotNull(i)` predicates on `Validity` make the polarity unambiguous; a raw `BitSet` puts it back in callers' hands.
>
> 5. **Length semantics mismatch.** `BitSet.length()` is "highest set bit + 1", not the logical count. The PR's contract is "bits past `count` are undefined" — a `long[]` carrying that contract is honest; a `BitSet` masks it and invites callers to trust `length()`.
>
> The `Validity` interface already gives ergonomic high-level ops (`nullCount`, `nextNull`, `nextNotNull`); `words()` is the escape hatch when those don't fit. That covers both ends without a `BitSet` in the middle.
>
> Where a `BitSet` *would* be reasonable is purely cold paths (debug pretty-printing, one-shot introspection). For the per-batch reader path it's a regression in both directions — slower and more error-prone.

gunnarmorling:
Squashed and applied. Added one more commit for some minor adjustments and doc updates, too. Thanks a lot, @iifawzi!

core/src/main/java/dev/hardwood/reader/Validity.java
iifawzi: I believe claude is highly influenced by the PR description pushing for long benefits, but I personally agree with you that bitSet is providing better dev experience for users looking for high level usage (i wouldn't love to play with binary bits myself), however claude point related to the visibility class makes sense, for high level usage, the helper methods should answer all questions, isNull, isNotNull, etc.. this's already giving all answers needed 

the word backed solution main advantage in my opinion is that users can use `numberOfTrailingZeros` to effectively find next set bits on sparse columns, something that's not possible with only bitSet being exposed


on the other side, this change isn't showing any regression or e2e improvements, at least with current usage in tests today, however I believe it's an overall win, having it long packed could open the room for more improvements using SIMD for instance, so i'd say we can go with it given the benefit of faster iterations using `numberOfTrailingZeros` unless if dev x is our main concern and we think the helper methods on validity aren't enough

gunnarmorling: Yeah, I think the (somewhat unconscious) decision we've made is this:

- The `RowReader` API is geared towards ergonomics, accepting perf penalities
- The `ColumnReader` API is designed for peak perf, also if it means accepting usability quirks

I think this is a defensible and coherent design we can lean into. In that light, exposing `long[]` on the column reader API is fine as it allows for that usage of tzcnt.

PR #492, opened 2026-05-23
#453 Fix flaky S3 byte-comparison test

### LINKED ISSUE #453: Flaky S3SelectiveReadJfrTest.projectionTransfersFewerBytes
The test `S3SelectiveReadJfrTest.projectionTransfersFewerBytes` occasionally fails in CI because the one-column read transfers slightly more bytes (at the socket level) than the on-disk file size that the assertion uses as the upper bound.

## Failure

Seen in [run 25682614121, job 75398052441](https://github.com/hardwood-hq/hardwood/actions/runs/25682614121/job/75398052441) on `main` (commit 7243c757a1a7e65455d36f09a53f8c7ea453d839):

```
[ERROR] dev.hardwood.s3.S3SelectiveReadJfrTest.projectionTransfersFewerBytes -- Time elapsed: 1.036 s <<< FAILURE!
java.lang.AssertionError:
[Reading 1 of 3 columns should transfer fewer bytes than the file size]
Expecting actual:
  179092L
to be less than:
  170022L
	at dev.hardwood.s3.S3SelectiveReadJfrTest.projectionTransfersFewerBytes(S3SelectiveReadJfrTest.java:230)
```

## Analysis

The assertion at `s3/src/test/java/dev/hardwood/s3/S3SelectiveReadJfrTest.java:230` compares the sum of `jdk.SocketRead.bytesRead` events against `pageIndexFileSize` (the on-disk size of `page_index_test.parquet`, ~170 KB). The comment on the test states:

> Use the file size as the full-read baseline — a full read can't transfer more than the file itself.

That premise does not strictly hold for socket-level byte counts: `jdk.SocketRead` measures bytes pulled off the socket, which for many small HTTPS range requests includes per-request HTTP response headers (and any TLS framing) on top of payload. With a file that is only marginally above the 64 KB tail-cache threshold and column chunks fetched via several range requests, the overhead can push the socket-read total a few percent above the raw file size even when the payload is strictly smaller. Here the actual (179,092) is ~5% over the 170,022 byte file.

This is the same family of timing/transport-dependent assertion as #305 (`earlyCloseTransfersFewerBytesThanFullRead`), in the same test class.

## Suggested direction

Either:

- Replace the file-size baseline with an explicit full-read measurement via `readAndMeasureSocketBytes(PAGE_INDEX_FILE, ColumnProjection.all(), null)` and compare projection bytes against that (both measurements then include the same per-request overhead), or
- Use a larger fixture where projection savings dominate per-request overhead by a comfortable margin, or
- Subtract a per-range-request overhead allowance from the assertion.

iifawzi:
Fixes #453.

The bug is reproducible at a very low rate (<1/15,000 runs when I was able to reproduce it), and apparently it's **not** the HTTP request headers causing the inflated byte count we observed — it's extra `jdk.SocketRead` events fired from inside JDK HttpClient's NIO selector loop, specifically at `jdk.internal.net.http.SocketTube.readAvailable` on the `HttpClient-N-SelectorManager` thread.

## Investigation summary

I verified across three independent vantage points that the inflation lives below the application:

| Layer | Counter | Behavior |
|---|---|---|
| Server wire (TCP relay between `S3Source` and s3proxy) | `relayDown` | Deterministic — exactly 146,617 bytes for the one-column read every iteration. s3proxy sends what we asked for and nothing more. |
| Application (`S3InputFile`) | `networkBytesFetched` / `streamReadActualBytes` | Deterministic — exactly the requested bytes per read, no retries (`networkRequestCount` always 3). |
| JDK NIO socket reads | `jdk.SocketRead.bytesRead` | Occasionally inflated by ~15-100 KB. All extra events captured with stack traces showing only `jdk.internal.net.http.SocketTube.readAvailable` → `SequentialScheduler.runOrSchedule` → `InternalReadSubscription.signalReadable`. No Hardwood frames anywhere. |

Three plausible mechanisms remain at the JDK layer (TCP retransmits being observed as additional `SocketChannel.read()` events, eager NIO buffer fills that are discarded, or HttpClient internal subscriber-cancellation drains), but **none of them are reachable from Hardwood code** — every frame in the captured stacks is `jdk.internal.net.http.*`.

## Fix

Switched the projection test from `jdk.SocketRead` to the application-level counter `S3InputFile.networkBytesFetched()`. This counts bytes the reader actually requested (tail prefetch + every `readRange` length when it hits the network), which is what the test semantically wants to assert anyway. The assertion against `pageIndexFileSize` is kept — `networkBytesFetched` is fully deterministic at 145,758 bytes for the one-column read, giving ~24 KB of headroom against the 170,022 byte file size.

The other tests in the file still use JFR because they need socket-level signal to validate lazy-fetch / back-pressure / early-close behavior that isn't observable from the application counter.

## Test plan
- [x] `./mvnw -pl s3 verify` — all 147 tests pass
- [x] Stress test (`@RepeatedTest(10000)` locally) — 0 failures with the new assertion source

### DIFF
--- a/s3/src/test/java/dev/hardwood/s3/S3SelectiveReadJfrTest.java
+++ b/s3/src/test/java/dev/hardwood/s3/S3SelectiveReadJfrTest.java
@@ -215,18 +215,24 @@ public class S3SelectiveReadJfrTest extends AbstractJfrRecorderTest {
 
     @Test
     void projectionTransfersFewerBytes() throws Exception {
-        // page_index_test.parquet is 170 KB (> 64 KB tail cache), so column chunk
-        // reads go to the network and are observable via jdk.SocketRead. Use
-        // the file size as the full-read baseline — a full read can't transfer
-        // more than the file itself.
-        long oneColumnBytes = readAndMeasureSocketBytes(PAGE_INDEX_FILE,
-                ColumnProjection.columns("id"), null);
+        // page_index_test.parquet is 170 KB (> 64 KB tail cache), so column chunks
+        // reads go to the network.
+        S3InputFile s3File = (S3InputFile) source.inputFile("test-bucket", PAGE_INDEX_FILE);
+        try (ParquetFileReader reader = ParquetFileReader.open(s3File);
+             RowReader rows = reader.buildRowReader().projection(ColumnProjection.columns("id")).build()) {
+            while (rows.hasNext()) {
+                rows.next();
+            }
+        }
+        long oneColumnBytes = s3File.networkBytesFetched();
 
         assertThat(oneColumnBytes)
                 .as("Reading 1 of 3 columns should still transfer some bytes (file > tail cache)")
                 .isGreaterThan(0);
         assertThat(oneColumnBytes)
-                .as("Reading 1 of 3 columns should transfer fewer bytes than the file size")
+                .as("Reading 1 of 3 columns should transfer fewer bytes than the file size; "
+                        + "oneColumn=%,d bytes, fileSize=%,d bytes"
+                                .formatted(oneColumnBytes, pageIndexFileSize))
                 .isLessThan(pageIndexFileSize);
     }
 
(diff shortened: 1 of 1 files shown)


iifawzi:
Hello, I've been investigating this as I was afraid it might be something worse than HTTP headers (corrupt data, retries, etc.), but it isn't. It seems to be an additional read on the socket inside JDK HttpClient's NIO selector that, for some reason I haven't dug deep enough to fully explain, is always discarded and never reaches the application.

given we already track bytes at the application layer (S3InputFile.networkBytesFetched) and that counter is deterministic, asserting against it seemed like a better approach than measuring the full read at the socket layer.

gunnarmorling (approved):
Awesome, thanks a lot! This one was a bit annoying indeed.

PR #497, opened 2026-05-24
#475 Extend FlatRowReader to non-repeated struct columns

mghildiy:
## Checklist

- [x] Build via `./mvnw clean verify` passes
- [x] The commit message is prefixed with the issue key ("# 475 Extend FlatRowReader to non-repeated struct columns")
- [x] Code changes are covered by tests(existing tests cover the code changes in this case)
- [ ] If this PR adds or changes user-facing behaviour, `docs/` has been updated

Issue #475, opened 2026-05-12
Extend FlatRowReader to non-repeated struct columns


gunnarmorling:
## Motivation

`FlatRowReader` is currently selected only when `FileSchema.isFlatSchema()` is `true`, which requires both no top-level `GroupNode` **and** `maxRepetitionLevel == 0` for every column (`core/src/main/java/dev/hardwood/schema/FileSchema.java`). Any non-repeated struct at the top level fails the first check and routes to `NestedRowReader`, even though the on-disk layout is row-aligned (one value per row per leaf, no record offsets needed) and would not benefit from the rep-level / record-offset machinery.

Struct-of-primitives is probably the most common nested shape in real-world Parquet files, so widening `FlatRowReader`'s applicability here is worth investigating.

## Sketch

- Change the dispatch discriminator in `FileSchema.isFlatSchema()` from "no top-level `GroupNode`" to "`maxRepetitionLevel == 0` everywhere". Repeated columns and maps continue to route to `NestedRowReader`.
- Extend `FlatRowReader`'s name map so a group name resolves to a synthesized `PqStruct` flyweight over its leaf columns instead of throwing in `getStruct()`.
- For optional groups, carry one extra "group present" bitset per group derived from `defLevel >= groupDefLevel`, so leaf-null vs struct-null stays distinguishable.

## Hard requirement: no regression on the existing flat path

`FlatRowReader` is `final` and built around monomorphic call sites on primitive arrays. Adding struct support must not regress the existing flat-schema benchmarks. Any change that touches the hot loops needs to be backed by `-Pperformance-test` numbers before and after, and we should be willing to fall back to a separate reader (e.g. `FlatStructRowReader`) if the unified path measurably slows down the pure-flat case.

mghildiy:
I woud look into it.

gunnarmorling:
Awesome, thanks a lot! Just ping me if there are any questions!

mghildiy:
I plan to have a FlatPqStructImpl for this. I also plan to have a AbstractPqStruct where FieldDesc.Struct resides, and have all the child resolution related code,and other common code, which can then be reused in diff implementtions of PqStruct.
Data would be inside new class FlatBatchData, containing just value arrays and bitset for validity check.

mghildiy:
[PR](https://github.com/hardwood-hq/hardwood/pull/497) created for it.

PR #526, opened 2026-05-30
#525 Add guard test for record-side filtering on a primitive after a struct

### LINKED ISSUE #525: Add guard test: record-side filtering on a top-level primitive following a struct (field index != leaf-column index)
## Goal

Add a regression-guard test for filtering on a **top-level primitive that follows a non-repeated struct**, evaluated through the **record-side filter path** (`RecordFilterCompiler` → `FilteredRowReader`).

This test passes on `main` today. The point is to have it in place **before** any change that alters how struct-of-primitives files are read, so a specific, easy-to-introduce regression is caught by CI rather than shipped silently. In-flight work (#475 / #497, which routes these files through `FlatRowReader`) is exactly the kind of change this guards.

## Why this case is fragile

A reader that addresses columns by index has two distinct index spaces:

- **field index** — position among top-level fields (what the public `getInt(int)` / `isNull(int)` accessors take)
- **projected leaf-column index** — position among physical leaf columns (how value arrays are addressed)

For a pure-flat schema they're equal. For a schema **containing a struct** they diverge for any top-level primitive *after* the struct, because the struct contributes more than one leaf column.

Example — `id INT, point STRUCT{x, y}, score INT`:

| column | field index | leaf-column index |
|---|---|---|
| `id` | 0 | 0 |
| `point` | 1 | 1, 2 |
| `score` | **2** | **3** |

The record-side filter path compiles indexed leaves that call back into `reader.getInt(int)` / `reader.isNull(int)`. If the reader maps a filter column to a **leaf-column index** where the accessor expects a **field index** (e.g. by passing `projectedSchema::toProjectedIndex`), filtering on `score` addresses the wrong field — or throws when the leaf index runs past the field count.

## Why the current suite doesn't cover it

- Pure-flat schemas never make the two indices diverge.
- Struct schemas with filters today either filter on the first field (index 0, still coincident) or on struct children (which fall back to the name-keyed path).
- Struct-of-primitives files currently route to `NestedRowReader`, so the `FlatRowReader` record-side path is never exercised with this shape.

## What the test must do

1. **Schema:** at least one struct followed by a top-level primitive, so `fieldIndex != leafColumnIndex` for that primitive — e.g. `id INT, point STRUCT{x,y}, score INT` — plus a `tag STRING` column (see point 2).
2. **Predicate shape — must be guaranteed to fall through to the record-side compiler.** This is subtle: drain-side (`BatchFilterCompiler`) eligibility has grown — on `main` both `AND` and `OR` of column-local leaves are drain-eligible, so neither alone forces the record-side path. A predicate on a `BYTE_ARRAY`/string column is **not** drain-eligible (`leafColumnIndex` returns `-1`), and one non-eligible leaf inside an `AND` makes the whole query fall back (all-or-nothing). So use, e.g.:
   ```java
   FilterPredicate.and(
       FilterPredicate.eq("tag", "keep"),   // binary leaf -> forces record-side path
       FilterPredicate.gt("score", 55));     // int leaf on the post-struct primitive -> exercises the index mapping
   ```
   Do **not** rely on `OR` to force the record-side path — it no longer does on `main`.
3. **Assertion:** the filter returns exactly the rows whose `score` matches.

## The failure mode it guards against

If a reader confuses the two index spaces for `score` (field 2 vs leaf 3), the test surfaces it as either:

```
java.lang.IllegalArgumentException: Field at index 3 is not a primitive or struct
    at FlatRowReader.isNull(...)
    at RecordFilterCompiler.lambda$indexedIntLeaf$...(...)
    at FilteredRowReader.hasNext(...)
```

or, if the stray index lands on a different valid primitive, a silent wrong result.

## Acceptance criteria

- [ ] A fixture (via `tools/simple-datagen.py`) with a non-repeated struct followed by a top-level primitive, plus a string column.
- [ ] A test that filters on the post-struct primitive via an `AND` containing a binary-column leaf, asserting the correct matching rows.
- [ ] The test passes on `main` and is committed independently of #497, so it acts as the regression gate when struct-of-primitives reading moves to the flat path.

iifawzi:
Closes #525.

## What

Adds a guard test (`StructThenPrimitiveFilterTest`) and a fixture
(`struct_then_primitive_test.parquet`, schema `id INT, point STRUCT{x,y}, tag STRING, score INT`)
covering record-side filtering on a **top-level primitive that follows a struct**.

## Why

A reader that addresses columns by index has two index spaces:

- **field index** — what `getInt(int)` / `isNull(int)` take (top-level fields)
- **leaf-column index** — how value arrays are addressed (physical leaves)

They coincide for pure-flat schemas but diverge for a top-level primitive after a struct: in the fixture, `score` is field index **3** but leaf-column index **4** (the `point` struct adds two leaves). The record-side filter path (`RecordFilterCompiler` → `FilteredRowReader`) addresses filter columns through the reader's indexed accessors, so a reader that confuses the two spaces there returns wrong rows — or throws — with no other signal.

This case is currently uncovered, and in-flight work (#475 / #497) routes struct-of-primitives files to `FlatRowReader`, which is exactly where the confusion can be introduced. Landing the test on `main` first means such a regression fails CI instead of shipping silently.

## Notes

- The predicate is `and(eq("tag","keep"), gt("score", 55))`. The `BYTE_ARRAY` leaf on `tag` is **not** drain-side eligible, so the whole query falls through to the record-side path (where the bug would live). An `OR` would not work — `OR` of column-local leaves is drain-side eligible on `main`.
- Passes on `main` (these files route to `NestedRowReader` today, which handles the indices correctly).
- Fixture regenerated via `tools/simple-datagen.py`; no other fixtures changed (pinned `pyarrow`/`thriftpy2` versions match).


gunnarmorling:
Nice one, thanks!

Issue #527, opened 2026-05-30
Eliminate redundant subtree re-walk in FileSchema schema build


iifawzi:
## Summary

`FileSchema.buildChildren` walks the flat depth-first `SchemaElement` list to
build the `SchemaNode` tree. For every **group** child it walks that group's
subtree **twice**: once via the recursive `buildChildren` call (to build the
nodes), then again via `countDescendants` purely to advance the cursor past the
subtree. The second walk is redundant — the recursion already knows where the
subtree ends.

This is a code-cleanliness / DRY change. It has **no measurable runtime impact**
(schemas are tiny and parsed once per file open); the motivation is removing a
duplicated traversal that is a latent maintenance hazard.

## Current behaviour

`core/src/main/java/dev/hardwood/schema/FileSchema.java`

- `buildChildren` recurses into a group at **:215**, building all descendant nodes.
- It then re-scans the same subtree at **:239** to learn its size:
  ```java
  // Skip over this group and all its descendants
  currentIndex = currentIndex + 1 + countDescendants(elements, currentIndex + 1, groupNumChildren);
  ```
- `countDescendants` (**:290**) is a second traversal routine that exists only to
  produce that count.

When the recursive `buildChildren` call returns, its own local `currentIndex` is
already sitting exactly one past the subtree — but `buildChildren` returns only
`List<SchemaNode>`, so that position is discarded and recomputed.

### Cost

`countDescendants` is invoked once per group and walks all of that group's
descendants, so a leaf is re-counted once per ancestor group: roughly
`O(n · depth)` (≈ `O(n²)` for a degenerate deep chain). Again — negligible at
real schema sizes; flagged for cleanliness, not speed.

### Why it's worth fixing

Two independent traversal routines (`buildChildren` and `countDescendants`) must
agree, byte-for-byte, on how the element list is interpreted (group vs primitive,
`num_children` handling, cursor advancement). If one is changed and the other is
not, the cursor desyncs and the schema tree is **silently corrupted** — and the
`columnIndex` ⇄ `ColumnChunk`-order invariant the whole reader depends on breaks
without throwing.

## Proposed change

Make `buildChildren` expose the cursor it already computes, and delete
`countDescendants` entirely.

Preferred form (smallest diff, matches the existing idiom): thread a shared
`int[] cursor` the same way `columnIndex` is already threaded through these
methods, so the recursion advances the caller's cursor directly. Alternatively,
[...]

PR #528, opened 2026-05-30
#527 Advance a shared cursor in buildChildren, drop countDescendants

iifawzi:
Hey, I decided to leave the examples repo for tomorrow, and have been reading through the components with Claude to deepen my understanding of the architecture, especially the parts I've not touched, and we came across a spot where it flagged a duplicated traversal

buildChildren already walks each group's subtree to build it; the recursion knows where the subtree ends. Threading the read position through a shared cursor lets siblings resume directly instead of re-scanning via countDescendants, removing a duplicated traversal that had to stay in lockstep with the build walk and was a latent source of silent schema mis-building.

## Checklist

### DIFF
--- a/core/src/main/java/dev/hardwood/schema/FileSchema.java
+++ b/core/src/main/java/dev/hardwood/schema/FileSchema.java
@@ -142,7 +142,9 @@ public class FileSchema {
         List<ColumnSchema> columns = new ArrayList<>();
         int[] columnIndex = { 0 }; // Mutable counter for column indexing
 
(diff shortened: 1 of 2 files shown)


iifawzi:
Curious to also ask about the decision related to converted types @gunnarmorling @rionmonster, I didn't find any documented decision on why we don't have them wired today, wiring some of them would be as easy as extending effectiveLogicalType, or it's not worth it given been deprecated for years?

gunnarmorling:
Rebased and applied. Thanks a lot, @iifawzi!

gunnarmorling:
As for your question on legacy converted types, excellent point indeed. The intend indeed was to expose everything through `effectiveLogicalType`, but looks like we only actually did that for `INTERVAL`. Logged https://github.com/hardwood-hq/hardwood/issues/529 for tracking this.

Issue #543, opened 2026-06-02
Rename RowReaderBuilder.firstRow() to skip()


gunnarmorling:
Rename the `RowReaderBuilder.firstRow(long)` builder method to `skip(long)` before 1.0 freezes the public API.

### Why

`firstRow(n)` reads as a **physical** position — "the first row is #n" / "start at row n." That connotation is exactly wrong for the semantics the method is settling on: under a row-level filter it is a **logical** offset over the matched rows (skip the first `n` matches), per #538/#541. The physical reading is what seeded the confusion in #538's discussion.

`skip(n)` is the streaming/LINQ vocabulary (`skip`/`take`) that:

- is agnostic to physical-vs-logical — correct in both the no-filter case ("skip n rows") and the filtered case ("skip n matches");
- sits naturally alongside the `head`/`tail` pair without importing SQL's `offset`/`limit` pairing (which would reopen the question of renaming `head`→`limit` and strand `tail`, which has no SQL counterpart);
- reads cleanly when composed: `skip(150).head(20)` = "skip 150, take the first 20."

`head` and `tail` are **not** renamed: they are a deliberate symmetric pair, and the Unix-pipe reading (`filter | head` → first n matches) already matches the chosen logical semantics.

### Why now

The method is public API on the 1.0.0.Final track. Pre-1.0 is the only cheap window — post-1.0 this is a deprecation cycle. #540 (guard) and #541 (lift) both name this method, so the rename should land before or with them.

### Work

- Rename `RowReaderBuilder.firstRow(long)` → `skip(long)`. Internal field/variable names (`firstRow`, `withinRg`, etc.) may follow for consistency but are not part of the public surface.
- Update `docs/content/` (the row-limit / query-controls how-to) and the method JavaDoc.
- Update the wording and acceptance criteria in #540 and #541 to the new name (done on those issues).

### Acceptance criteria

- The builder exposes `skip(long)`; `firstRow` no longer appears in the public API.
- Docs and JavaDoc use `skip`.

Issue #542, opened 2026-06-02
Support tail combined with a filter (logical, over matched rows)


gunnarmorling:
Follow-up to #538. Completes the symmetry: when a row-level filter is present, the positioning controls operate over the *filtered* relation (SQL `WHERE` first, then the control). #538 establishes this for `head` (`LIMIT`) and #541 for `firstRow` (`OFFSET`); this issue is the `tail` counterpart.

### Current state

`tail` + filter is rejected outright at `RowReaderBuilder.build()`:

```
tail cannot be combined with a filter: the set of matching rows is not known
from row-group statistics alone
```

This is intentional for 1.0 (see the discussion on #538) — it is **not** a silently-wrong path like the `firstRow` case in #540, it is a clean rejection. This issue tracks lifting that restriction in a later version.

### Intended semantics

`tail(n)` with a filter returns the **last `n` matching rows**, in file order — the SQL equivalent of ordering by physical position and taking the final `n` rows of the filtered relation. Symmetric with `head(n)` returning the first `n` matches.

### Why it is deferred (and harder than head/firstRow)

`head` and `firstRow` stream forward and can stop as soon as the match budget is met. `tail` cannot: the position of the last `n` *matching* rows is not derivable from row-group statistics (which only bound min/max values, not match counts), so there is no skip-to-the-tail shortcut. Serving it correctly requires either a full forward scan retaining a sliding window of the last `n` matches, or reverse iteration with match counting from the end. That is a meaningfully larger piece of work than the forward-streaming controls, which is why it sits outside 1.0.

### Work

- Remove the `tail` + `FilterPredicate` rejection from `RowReaderBuilder.build()`.
- Implement last-`n`-matching-rows retrieval (sliding window or reverse scan), reusing the matched-row counting introduced for #538.
- Decide the `tail` + `RowGroupPredicate` (`byteRange`) interaction separately — that combination is also currently rejected because `byteRange` invalidates the known total row count that tail-mode row-group selection relies on; it may or may not be lifted here.

### Acceptance criteria

- `tail(n).filter(p)` returns the last `n` rows matching `p`, in file order.
- Tests cover the tail window spanning a row-group boundary and `tail` larger than the match count.

Issue #541, opened 2026-06-02
Support skip combined with a filter (logical, over matched rows)


gunnarmorling:
Lifts the limitation introduced in #540 (which rejects `skip` + filter at `build()` because the combination is currently silently wrong). Follows the matched-row `head` fix in #538. The end-state model is captured in `_designs/ROW_SELECTION_SEMANTICS.md`.

> **Naming:** the builder method `firstRow` is being renamed to `skip` (#543). This issue uses `skip`; the internal field/variable still reads `firstRow`/`withinRg` in the linked code until that rename lands.

### Goal

Support `skip(n)` together with `filter(...)` and remove the `build()` guard added in #540.

### Intended semantics

`skip` is a **logical** offset applied **after** the predicate — SQL `OFFSET`, symmetric with `head` being SQL `LIMIT` (#538). So:

```java
buildRowReader().skip(150).filter(p).head(20)
```

means *"skip the first 150 rows matching `p`, then return the next 20 matching rows"* — i.e. `OFFSET 150 LIMIT 20` over the filtered relation. It does **not** mean "skip 150 physical rows."

With no filter present, logical and physical coincide (every row matches), so this changes nothing for the plain `skip(n)` / `skip(n).head(k)` paths and leaves room for byte-level page-skip (#381) there. Physical positioning of a scan — split-aware reading — is a separate concern already served by `RowGroupPredicate.byteRange(...)`, not by `skip`.

### Work

- Apply the `skip` over **matched** rows, after the filter, rather than the current hybrid that locates the row group by physical `numRows` and then skips within it via `reader.next()` (see [`ParquetFileReader.java:286-291`](https://github.com/hardwood-hq/hardwood/blob/1966f207951558ff1b0295a9d698c4263fcf10ac/core/src/main/java/dev/hardwood/reader/ParquetFileReader.java#L286-L291)) — the incoherence behind the #540 bug.
- Reconcile with the matched-row `head` counter from #538: both `skip` and `head` now count post-filter rows, so the skip-then-limit composition must count the same way end to end (replacing the `maxRows + withinRg` pre-adjustment at [`ParquetFileReader.java:278-282`](https://github.com/hardwood-hq/hardwood/blob/1966f207951558ff1b0295a9d698c4263fcf10ac/core/src/main/java/dev/hardwood/reader/ParquetFileReader.java#L278-L282), which conflates a physical residue with the limit counter).
[...]

Issue #538, opened 2026-06-01
Potential inconsistency of the documented and actual behavior of head


iifawzi:
Hey, while working on the examples, I was testing filtration + head, and it yielded no data, which doesn't align with the file I was testing
Checking it, it's because the head() limits the maximum number of read rows, not filtered rows:

https://github.com/hardwood-hq/hardwood/blob/c5de421ec2a06982477b66b30dfaf1d171421361/core/src/main/java/dev/hardwood/internal/reader/FlatColumnWorker.java#L87-L94

However, it's different than the behavior we're documenting when it's combined with a filter
<img width="1298" height="915" alt="Image" src="https://github.com/user-attachments/assets/f3a81bc0-6b89-4e13-8b6f-1ffc5b071313" />

I believe the semantics of `head` was to match the `tail` behavior (hence the namings); in that case, it seems reasonable that it limits read rows, not matched rows, then we need to update the documentation, and the question would be whether we need to support a limit of matched rows or not @gunnarmorling @rionmonster 

I personally find it cool to have it supported.

gunnarmorling:
I think both head and tail should apply to the matched rows (if you want it to apply to read rows, don't apply a filter), just like a SQL statement's `LIMIT` clause applies to whatever rows which have passed a `WHERE` filter. If that's actually not the current behavior, we should fix it.

iifawzi:
> I think both head and tail should apply to the matched rows (if you want it to apply to read rows, don't apply a filter), just like a SQL statement's `LIMIT` clause applies to whatever rows which have passed a `WHERE` filter. If that's actually not the current behavior, we should fix it.

That reasoning makes sense, for head it's not the current behavior, will check Tail later today, and probably patch both, shouldn't be hard

gunnarmorling:
On second thought, in regards to tail, we actually currently don't support tail with filters. I think that's ok for 1.0. Making head + filter adhere to what's described in the docs would be a good fix to have for 1.0, as that's an actual bug.

gunnarmorling:
Discussing this with Claude, it pointed out a related issue: confusing semantics of firstRow + filter. So I've logged #540 to forbid this combination for now, and #541 for bringing it back with correct behavior. If we can get done #541 in time for 1.0, I'd be for it; otherwise, #540 would be the cheap option to fail early for that combo rather than silently returning weird results.

gunnarmorling:
Spun off the related concerns so this issue stays scoped to the `head` + filter bug:

[...]

PR #544, opened 2026-06-02
#543 rename firstrow to skip

gunnarmorling:
I think `skip()` is a better name for what this method does, avoiding the "physical" semantics suggested by `firstRow()`. Also adding a design doc for putting a system behind #538, #541, #542.

### DIFF
--- a/_designs/ROW_SELECTION_SEMANTICS.md
+++ b/_designs/ROW_SELECTION_SEMANTICS.md
@@ -1,0 +1,81 @@
+# Design: row selection semantics
+
+**Status: In progress.** Tracking issues: #538 (`head` + filter), #540 / #541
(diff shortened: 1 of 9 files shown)


gunnarmorling:
@iifawzi, just in case you have some time, as it's related to #538.

iifawzi:
Nicee this's way much easier to understand, will take a look shortly on it and the rest of the related tickets as well.

iifawzi (approved):

_designs/ROW_SELECTION_SEMANTICS.md
```diff
@@ -0,0 +1,81 @@
+# Design: row selection semantics
```
iifawzi, line 1: Looks good to me, during implementation it would become clearer if there's any blockers/inconsistencies

PR #549, opened 2026-06-02
#547 Add Fawzi Essam to developers in POM

gunnarmorling:


pragmaticway (approved):

gunnarmorling:
@iifawzi, just realized you still were not listed in the POM :)

iifawzi:
truly honored :)

gunnarmorling:
Of course! Thanks a ton for all your work on this project. Not only do they make a massive impact, it's also much more fun to work on this together with like-minded folks :)

iifawzi:
Thank you @gunnarmorling happy to be part of that, it's a very interesting area, and working with you is an absolute fun (full of learnings as well), happy and excited to keep moving that forward

PR #564, opened 2026-06-02
#538 Cap matched rows, not scanned rows, for head() with a filter

### LINKED ISSUE #538: Potential inconsistency of the documented and actual behavior of head

iifawzi:
This PR resolves #538, capping on matched row instead of scanned, for head + filter

With a filter present, the workers now scan unbounded, and the cap is enforced over matches: the drain-side reader counts yielded matches, and the FilteredRowReader wrapper counts returned matches on the indexed and nested paths. 

The fetch layer also drops its scanned-position bounds when a filter is active (perRgMaxRows returns 0), so every statistics-surviving page stays fetchable; row-group and page pushdown remain the only pruning. 

The no-filter path is unchanged: the worker and fetch planner keep bounding scanned rows, where scanned equals yielded.

## Checklist

### DIFF
--- a/core/src/main/java/dev/hardwood/internal/reader/ColumnWorker.java
+++ b/core/src/main/java/dev/hardwood/internal/reader/ColumnWorker.java
@@ -98,6 +98,9 @@ public abstract class ColumnWorker<B> implements AutoCloseable {
     // === In-flight decode tasks (tracked so close() can await them) ===
     private final Set<CompletableFuture<Void>> inFlightDecodes = ConcurrentHashMap.newKeySet();
 
+    /// Sentinel for the `maxRows` / row-limit contract meaning "no limit".
+    static final long UNLIMITED = 0L;
+
     // === Drain assembly state (drain thread only) ===
     final long maxRows;
     long totalRowsAssembled;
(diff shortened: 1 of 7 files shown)


iifawzi:
On the other side, for examples repo, we need a way to document new behaviors/examples during development until the release

I'm thinking we can have a branch 1.0.0CR2 so all documentation work during 1.0.0CR1 lives there and it uses a local build of hardwood, which will be switched to upstream and merged to main before releasing 

haven't thought a lot about that, you might have better ideas @gunnarmorling

gunnarmorling:
LGTM, applying. Thanks a lot, @iifawzi!

Issue #565, opened 2026-06-03
Parquet spec conformance audit: findings


gunnarmorling:
Audit of hardwood's reader implementation **and** user docs against the normative Parquet spec, across four areas (nesting/levels, logical types, encodings, statistics/page-index). Each finding was classified CONFORMANT / GAP / DOC-GAP / UNSURE; the two highest-severity gaps were verified against the code before filing.

### Genuine defects (sub-issues)

- [x] **#566** — float/double NaN pruning (Tier 1 — silent data loss; verified in code)
- [x] **#568** — `getTimestamp` ignores `isAdjustedToUTC` (Tier 1 — wrong instant; verified in code)
- [ ] **#569** — v1 BIT_PACKED level encoding misdecoded as RLE (Tier 2 — legacy)
- [ ] **#567** — LIST backward-compat rule 3 missing (Tier 2 — legacy; the rule-3-untested gap below falls out of this)
- [ ] **#571** — decoder arithmetic hardening (Tier 3 — robustness)
- [x] **#570** — docs: float total-order, timestamp local-vs-instant, legacy-list claim (Tier 4)
- [ ] **#595** — surface `ColumnOrder`; warn or reject when not `IEEE754TotalOrder` for floats (promoted from the decisions below)
- [ ] **#596** — confirm legacy `MAP_KEY_VALUE`-only annotation reads correctly (promoted)
- [ ] **#597** — add `DELTA_BYTE_ARRAY` + `FIXED_LEN_BYTE_ARRAY` test fixture (promoted)

### Settled decisions

- **Float equality in filters uses total order** (`-0.0 ≠ +0.0`, `NaN == NaN`). Decided to keep; the divergence from IEEE / SQL `eq` lets callers positively select NaN rows and is now documented in `docs/content/how-to/query-controls.md` and `FilterPredicate`'s JavaDoc (#570).
- **LIST rule-3 untested.** Subsumed by **#567**, which adds the implementation and tests together.

### Verified conformant (the good news)

[...]

PR #584, opened 2026-06-03
#568 Honor isAdjustedToUTC in TIMESTAMP accessors

### LINKED ISSUE #568: getTimestamp ignores isAdjustedToUTC: local timestamps returned as a UTC Instant
From the spec conformance audit (#565). **Verified** in code.

`ValueConverter.convertToTimestamp` (`ValueConverter.java:106-114`) returns `Instant` unconditionally and never reads the TIMESTAMP logical type's `isAdjustedToUTC`. Per spec:

- `isAdjustedToUTC = true` → the value is an **instant** (UTC-normalized) → `Instant` is correct.
- `isAdjustedToUTC = false` → a **local** (wall-clock) timestamp with no timezone → it is *not* an instant.

Returning a local timestamp as an `Instant` yields the wrong absolute time (the reader implicitly treats the local value as UTC). The Variant path honors the flag (`VariantShredReassembler`); the primary `getTimestamp` does not, and there is no local (`LocalDateTime`) accessor. `getTime` has the same gap (lower impact — `LocalTime` is zoneless).

### Decision needed
Honor the flag and either (a) add a local accessor returning `LocalDateTime` for `isAdjustedToUTC = false`, or (b) define explicit semantics. Pre-1.0 because (a) is an API addition that should land before the freeze.

### Acceptance
- A local-timestamp fixture is no longer presented as a UTC instant (failing test first).
- Defined, documented behavior for `isAdjustedToUTC = false` (pairs with the docs issue under #565).

gunnarmorling:
## Summary

Closes #568.

The Parquet TIMESTAMP logical type carries an `isAdjustedToUTC` flag that picks between two distinct semantic kinds:

- `isAdjustedToUTC = true` → an instant normalized to UTC → `Instant`.
- `isAdjustedToUTC = false` → a wall-clock timestamp with no timezone → `LocalDateTime`.

The reader returned `Instant` unconditionally, never reading the flag. A local-wall-clock value was silently reinterpreted as UTC — wrong absolute time, no signal to the caller. `VariantValueDecoder.asTimestamp` had the same defect: all four Variant timestamp tags (`TIMESTAMP`, `TIMESTAMP_NTZ`, `TIMESTAMP_NANOS`, `TIMESTAMP_NTZ_NANOS`) collapsed to `Instant`. The NYC TLC sample in `YellowTripDataTest` was actually a real-world instance of this — `tpep_pickup_datetime` is NTZ, and the test was asserting the (incorrect) Instant readings.

Audit (#565) flagged the row-side as silent data loss; the Variant fix lands in the same PR because `getVariant().asTimestamp()` would otherwise leak the same bug.

## Decisions

- **Strict accessor split**: `getTimestamp` / `asTimestamp` are UTC-only (throw on the local kind); new `getLocalTimestamp` / `asLocalTimestamp` return `LocalDateTime` and throw on the UTC kind. Rejection error is `IllegalStateException` with a message naming the column and the actual `isAdjustedToUTC` value. Same model as the existing \"wrong accessor for the column kind\" mismatches.
- **Why strict, not lenient**: a lenient default-zone reinterpretation hides exactly the distinction that put the issue on the audit. parquet-avro — the only parquet-java binding that distinguishes the kinds at all — does it via two distinct return types on a single untyped getter. Hardwood is typed-accessor-first, so the analogue is two named methods.
- **`getValue` (generic accessor)**: returns `LocalDateTime` for local columns, `Instant` otherwise — schema-driven, like parquet-avro's `GenericRecord.get`.
- **INT96 stays as-is**: no `isAdjustedToUTC` field, treated as UTC-adjusted (Spark / Hive convention), read via `getTimestamp`.
- **`getTime` left untouched**: same flag exists but `LocalTime` is zoneless, so there is no data-corruption equivalent — flag is informational, documented next to `getTime`.

Full rationale (including the parquet-avro / protobuf / thrift survey and rejected alternatives) lives in [`_designs/LOCAL_TIMESTAMP_ACCESSOR.md`](https://github.com/hardwood-hq/hardwood/blob/568-local-timestamp-accessor/_designs/LOCAL_TIMESTAMP_ACCESSOR.md).

## Public API additions

| Interface | Added |
|---|---|
| `FieldAccessor` | `LocalDateTime getLocalTimestamp(String name)` |
| `StructAccessor` | `LocalDateTime getLocalTimestamp(int fieldIndex)` |
| `PqMap.Entry` | `LocalDateTime getLocalTimestampValue()` |
| `PqList` | `List<LocalDateTime> localTimestamps()` |
| `PqVariant` | `LocalDateTime asLocalTimestamp()` |
| `PqVariantObject` | inherits `getLocalTimestamp` from `FieldAccessor` |

`ColumnReader` is untouched (perf-first path, no typed timestamp accessor exists today).

## Test plan

- [x] Failing test first: `LocalTimestampTest` against a generated `local_timestamp_test.parquet` (3 columns: NTZ MILLIS, NTZ MICROS, UTC MICROS, 2 rows). Tests: `getLocalTimestamp` returns wall-clock; `getTimestamp` on a local column throws naming the column and `isAdjustedToUTC=false`; `getLocalTimestamp` on a UTC column throws naming the column and `isAdjustedToUTC=true`; `getTimestamp` on the UTC column still returns the right `Instant`; generic `getValue` returns `LocalDateTime` for local and `Instant` for UTC.
- [x] Variant: `VariantValueDecoderTest.primitiveTimestampNtz` / `primitiveTimestampNtzNanos` updated to use `asLocalTimestamp` against `LocalDateTime` expectations.
- [x] Real-world regression catch: `YellowTripDataTest` (core + integration-test) now correctly reads `tpep_pickup_datetime` / `tpep_dropoff_datetime` via `getLocalTimestamp` — previously asserting Instant values that happened to share the same wall-clock minute as the UTC misreading.
- [x] `./mvnw -pl core,avro,cli,s3,parquet-java-compat,integration-test verify` — green.
- [ ] CI on this PR.

## Out of scope

- `getTime` symmetric pair (D8 in the design doc — `LocalTime` is zoneless, no data loss).
- Lenient zone-aware `getTimestampAt(name, ZoneId)` (can be added later without breaking the strict pair).
- `ColumnReader` typed timestamp accessor (no such accessor exists today; perf-first contract is unchanged).


gunnarmorling:
Addressed review feedback (force-pushed `83472ef8` → `30b46ada`):

- **DRY + INT96 fix (one change)**: added `TimestampAccessorKind.require(name, lt, wantUtcAdjusted)` that handles the `instanceof TimestampType` + boolean dispatch *and* the INT96 fallthrough (`lt == null` → reject from `getLocalTimestamp` with a clean `IllegalStateException` naming the column; accept from `getTimestamp`). The 4 per-impl `requireXxxTimestampKind` wrappers across `NestedBatchDataView` / `PqStructImpl` / `PqListImpl` / `PqMapImpl` are now one-liners against the central helper; `FlatRowReader.getLocalTimestamp` routes through the same helper, fixing the INT96 NPE.
- **Switch exhaustiveness**: `LogicalTypeConverter.convertToLocalTimestamp` is now a switch expression (no `default`). Both CLI sites (`IndexValueFormatter`, `RowValueFormatter.formatRawTimestampLong`) delegate to `LogicalTypeConverter.convertToTimestamp` / `convertToLocalTimestamp` and `.toString()` the result, eliminating the two duplicated switch statements entirely.
- **Tests**: fixture extended with `nested` struct, `local_ts_list`, `local_ts_map` — `LocalTimestampTest` now covers `PqStruct.getLocalTimestamp`, `PqList.localTimestamps()`, `PqMap.Entry.getLocalTimestampValue()` plus wrong-kind rejection through each. Added `int96_timestamp_test.parquet` and pinned the INT96 fallthrough (accepted by `getTimestamp` as `Instant`, rejected by `getLocalTimestamp`). `PqVariantObject.getLocalTimestamp` left to the directly-tested `VariantValueDecoder.asLocalTimestamp` since the existing Variant fixture's `timestamp_field` is encoded as STRING rather than TIMESTAMP_NTZ.

`./mvnw -pl core,avro,cli,s3,parquet-java-compat,integration-test verify` — green.

PR #587, opened 2026-06-03
#541 Support skip combined with a filter as a logical OFFSET

### LINKED ISSUE #541: Support skip combined with a filter (logical, over matched rows)

iifawzi:
Heyy, this PR resolves #541 

This makes skip a logical OFFSET over the matched relation, symmetric with head() as LIMIT (#538): when a filter is present the reader scans the full row-group-filtered relation, caps at skip + head matches, and discards the first skip matches, so skip(n).filter(p).head(k) reads "OFFSET n LIMIT k" over the rows matching p. Without a filter, skip stays the physical O(1-row-group) seek.

- A skip past the end now yields an empty reader in both modes rather than throwing in the no-filter case — a SQL OFFSET overshoot returns nothing, and this aligns the code with the ROW_BASED_SEEK design, which already specified empty for skip >= totalRows
- The two residue walks (physical and logical) are unified in discardLeadingRows, which closes the partially built reader if iteration fails so its worker pipeline cannot leak. 
- The DuckDB differential harness gains the skip + filter and overshoot cases (its pending #541 marker is cleared), confirming hardwood agrees with SQL OFFSET semantics, including that OFFSET past the end returns empty without error. 
- The reader-facing model is documented in the new concepts/row-selection.md.

## Checklist

### DIFF
--- a/_designs/ROW_BASED_SEEK.md
+++ b/_designs/ROW_BASED_SEEK.md
@@ -19,11 +19,14 @@ from row 0 and fetching everything in between.
 A new method on the existing `RowReaderBuilder`:
 
 ```java
-/// Begin reading from the given absolute row index. Earlier row groups
-/// are not opened — their pages are not fetched or decoded — so this
-/// is an O(1 RG) seek on remote backends.
+/// Skip rows before reading — SQL `OFFSET`. **Without a filter**, `skip(n)`
+/// is a physical absolute row index: earlier row groups are not opened —
+/// their pages are not fetched or decoded — so this is an O(1 RG) seek on
+/// remote backends. **With a filter**, `skip(n)` is a logical offset over
+/// the matched rows (see ROW_SELECTION_SEMANTICS.md / #541), and the O(1)
+/// seek does not apply — earlier groups are decoded to count matches.
 ///
-/// `skip == 0` is the no-op default. `skip == totalRows`
+/// `skip == 0` is the no-op default. `skip >= totalRows`
 /// produces an empty reader. Indexes into the *first* file's rows for
 /// multi-file readers; cross-file `skip` is out of scope. Mutually
 /// exclusive with [#tail]; composes with [#head] for a bounded
@@ -54,7 +57,12 @@ file).
 
 ## Implementation
 
-In `ParquetFileReader.buildRowReader(projection, filter, maxRows, skip)`:
+The physical seek below is the **no-filter** path. When a `FilterPredicate` is
+present, `skip` is instead a logical offset over the matched rows — earlier
+groups must be decoded to count matches — handled separately per
+[ROW_SELECTION_SEMANTICS.md](ROW_SELECTION_SEMANTICS.md) (#541).
+
+In `ParquetFileReader.buildRowReader(projection, filter, maxRows, skip)`, with no filter:
 
 1. If `skip == 0`, delegate to the existing path (no behaviour change).
 2. Otherwise, locate the target row group by walking cumulative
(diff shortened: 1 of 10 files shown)


iifawzi:
When you can @gunnarmorling, i'm also curious about your thoughts about seek as a high-level functionality similar to byteRange https://github.com/hardwood-hq/hardwood/issues/541#issuecomment-4607572951

gunnarmorling:
LGTM, thanks a lot, @iifawzi! Amended for some minor adjustments, merging.

gunnarmorling:
> When you can @gunnarmorling, i'm also curious about your thoughts about seek as a high-level functionality similar to byteRange https://github.com/hardwood-hq/hardwood/issues/541#issuecomment-4607572951

Logged https://github.com/hardwood-hq/hardwood/issues/593 for discussing that one; will reply over there.

Issue #315, opened 2026-04-23
Defensive reader check: non-object shredded levels must not have both value and typed_value populated


gunnarmorling:
## Context

The Variant shredding spec describes a per-level invariant for how `value` and `typed_value` siblings may be populated:

| Level shape | `value` populated | `typed_value` populated | Meaning |
|---|---|---|---|
| Any | no | no | Variant NULL (distinct from SQL NULL) |
| Non-object (primitive / array) | **yes** | no | Unshredded — caller reads the opaque blob |
| Non-object (primitive / array) | no | **yes** | Shredded — caller reads the typed column |
| Object | yes | yes | **Allowed** — partial shredding: `typed_value` carries the known fields, `value` carries any remaining fields |
| Object | yes | no | Fully unshredded object |
| Object | no | yes | Fully shredded object |

For **non-object** levels, having both `value` and `typed_value` non-null for the same row is a writer invariant violation: the shredded value and the fallback blob would both claim to represent the same logical value, with no defined merge semantics. The spec treats this as "must not happen"; it's a writer's job to honour it.

Hardwood's reader currently does not check this invariant. If a malformed writer produces a row where both sides are populated at, say, a shredded INT64 level, `VariantShredReassembler.writePrimitiveLevel` will take the `typed_value` branch and silently drop the `value` bytes — producing plausible-looking output that misrepresents the file.

## Principle

This conflicts with Hardwood's "correctness-first, fail-early" stance (CLAUDE.md):

> Adhere to the "fail early" principle: validate inputs, check types, and throw on misuse rather than silently producing wrong results. Silent failures are never an option.

A defensive reader check is the right response: detect the invariant violation and throw, rather than silently prefer one side.

## Proposed fix

In `VariantShredReassembler.writePrimitiveLevel` and `writeArrayLevel` (the two non-object level shapes), check after resolving def-levels whether **both** `value` and `typed_value` are present for the current row. If so, throw a clear exception naming the column path and row. The check is one `&&` per non-object level — negligible cost; fires only on malformed input.

Object levels are untouched: their path already handles the both-present case via `mergeUnshreddedObject`.

## Scope

- `writePrimitiveLevel` in `VariantShredReassembler` (primitive shredded scalars).
- `writeArrayLevel` in `VariantShredReassembler` (shredded arrays, at the level itself — not at array element levels, which are themselves scalars/arrays/objects and recurse through the normal checks).
[...]

Issue #506, opened 2026-05-29
dive: statistics-validity and sortedness audit


gunnarmorling:
## Problem

Parquet files carry per-column min/max statistics that query engines trust for
range-pruning. Two real-world failure modes are currently invisible in `dive`:

1. **Lying stats.** Reported min/max don't actually bound the data — e.g. the
   long-standing parquet-mr signed-vs-unsigned min/max bugs. Engines that prune on these
   stats then return wrong results or skip the wrong row groups.
2. **Unsorted data.** "Why doesn't my range filter prune anything?" is usually because
   the column isn't actually sorted within its row groups, so every RG's [min,max]
   overlaps the predicate.

## Proposed feature

An audit that decodes each column and reports:

- whether the reported min/max actually bound the decoded values (per RG / per page),
- whether values are sorted within each row group (and the sort direction).

Surface mismatches prominently — this is a correctness diagnostic, not a cosmetic one.

## Notes

- The decode path needed for this already exists in core.
- `parquet-cli check-stats` validates min/max but does not report sortedness or per-page
  drift; combining both with `dive` navigation is the differentiated angle.

PR #602, opened 2026-06-04
#599 Reject shredded Variant objects repeating a field across typed_value and value

### LINKED ISSUE #599: Defensive reader check: shredded Variant object must not share a field name between typed_value and value
## Context

The Variant shredding spec allows an **object** level to have both `typed_value` (the shredded struct) and `value` (an unshredded object blob) populated for the same row — this is the *partial shredding* case: `typed_value` carries the known fields, `value` carries any remaining fields. See the level-shape table in #315.

That partial-shredding case is only valid under one invariant: **the field-name sets of the shredded struct and the unshredded `value` object must be disjoint.** The spec states the same field must never appear on both sides; if it does, the file is malformed (the two sides would each claim to define the field, with no defined merge semantics). The community's stated resolution for this case is "never trust the fields that are unshredded."

`VariantShredReassembler.mergeUnshreddedObject` currently *trusts* this invariant and appends every unshredded field with no deduplication — as its own comment states:

> Spec guarantees these names are disjoint from the shredded set, so we can simply append — no deduplication needed.

`emitObject` and `VariantValueEncoder.writeObject` likewise do not deduplicate. So if a malformed writer emits a row where a field appears in **both** the shredded struct and the unshredded `value`, Hardwood silently reassembles a Variant object containing **duplicate field ids** — itself invalid Variant binary — with no error anywhere in the read path.

This is distinct from #315, which covers *non-object* (primitive/array) levels and explicitly leaves object levels untouched ("their path already handles the both-present case via `mergeUnshreddedObject`"). This issue addresses that object path: the both-*sides*-present case is handled, but the both-*claim-the-same-field* sub-case is not.

## Principle

Same fail-early stance as #315 (CLAUDE.md):

> Adhere to the "fail early" principle: validate inputs, check types, and throw on misuse rather than silently producing wrong results. Silent failures are never an option.

Producing a corrupt Variant object (duplicate field ids) from malformed input is precisely the silent-wrong-result this rule forbids.

## Proposed fix

In `VariantShredReassembler.mergeUnshreddedObject`, track the shredded field names already emitted for the current object and, when an unshredded field collides with one of them, throw a clear exception naming the colliding field. A `HashSet<String>` (or a small scan, given objects are typically tiny) seeded with the shredded names is sufficient; the check fires only on malformed input.

Applies to both object reassembly entry points: `writeObjectLevel` (top-level / row) and `writeElementObject` (array-of-objects).

## Acceptance criteria

- A shredded-object row whose unshredded `value` repeats a shredded field name throws with a message identifying the field.
- Valid partial-shredding (disjoint field sets) continues to reassemble unchanged.

Related: #315 (non-object both-present check).

gunnarmorling:
Closes #599.

## Problem

The Variant shredding spec lets an object be *partially shredded* — `typed_value` carries the known fields and `value` carries the rest — but only when the two sides name **disjoint** fields. `VariantShredReassembler.mergeUnshreddedObject` trusted that invariant and appended the unshredded fields without checking. A malformed file naming the same field on both sides was therefore silently reassembled into a Variant object with **duplicate field ids** — corrupt output, no error. That is precisely the silent-wrong-result the project's fail-early stance forbids.

This is the object path that #315 explicitly left open ("their path already handles the both-present case via `mergeUnshreddedObject`"); #315 covers the distinct non-object case.

## Change

- Detect the collision during the merge and throw `IllegalStateException` naming the offending field, rather than emit a corrupt Variant. The check lives in `mergeUnshreddedObject`, so it covers both reassembly entry points — `writeObjectLevel` (row) and `writeElementObject` (array-of-objects).
- Objects are typically tiny, so the check is a linear scan over the already-emitted shredded names — no set allocation on the read path.

## Tests

- New `VariantShredReassemblerTest` constructs a malformed shredded object (same field on both sides) and asserts reassembly throws. Written test-first; confirmed it failed (silent corruption) before the fix.
- Valid partial-shredding is unaffected: the end-to-end shredding tests (`VariantLogicalTypeTest`, `VariantInRepeatedAccessorsTest`, which read real fixtures) still pass.
- Full `core` suite: 1290 tests, 0 failures, 0 errors.

## Acceptance criteria (from #599)

- [x] A shredded-object row whose unshredded `value` repeats a shredded field name throws with a message identifying the field.
- [x] Valid partial-shredding (disjoint field sets) continues to reassemble unchanged.

### DIFF
--- a/core/src/main/java/dev/hardwood/internal/reader/VariantShredReassembler.java
+++ b/core/src/main/java/dev/hardwood/internal/reader/VariantShredReassembler.java
@@ -269,16 +269,20 @@ public final class VariantShredReassembler {
 
     /// Decode the partial-unshredded object bytes in `raw`, appending each
     /// (name, valueBytes) pair into `names` / `values` starting at `count`.
-    /// Spec guarantees these names are disjoint from the shredded set, so we
-    /// can simply append — no deduplication needed. Field ids index into the
-    /// top-level metadata dictionary (same for all nesting depths within a
+    /// The spec requires these names to be disjoint from the shredded set
+    /// (`names[0..count)`); a collision is a malformed file that would yield a
+    /// Variant object with duplicate field ids, so we reject it via
+    /// [#rejectFieldCollision] rather than append blindly. Field ids index into
+    /// the top-level metadata dictionary (same for all nesting depths within a
     /// variant), resolved via `currentMetadata`. Returns the new count.
     private int mergeUnshreddedObject(byte[] raw, String[] names, byte[][] values, int count) {
+        int shreddedCount = count;
         ObjectLayout layout = VariantValueDecoder.parseObject(raw, 0);
         int n = layout.numElements();
         for (int i = 0; i < n; i++) {
             int fieldId = VariantValueDecoder.objectFieldId(raw, layout, i);
             String fieldName = currentMetadata.getField(fieldId);
+            rejectFieldCollision(fieldName, names, shreddedCount);
             int valueStart = VariantValueDecoder.objectValueOffset(raw, layout, i);
             // The object's offset array has `numElements + 1` entries, so
             // offset[i+1] is always valid and points at the byte past element i.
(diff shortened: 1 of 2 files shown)


gunnarmorling:
Follow-up verification against the in-progress parquet-format 2.13.0 change **GH-506 / PR #565** ("Clarify field ordering for partially shredded object union"): it clarifies that a reconstructed object's `field IDs and offsets must be in the order of the corresponding field names, sorted lexicographically (unsigned byte ordering for UTF-8)`. The reassembler here already does exactly that — `emitObject` sorts via `insertionSortByName` using `Arrays.compareUnsigned` on the UTF-8 field-name bytes before resolving ids — so this merge is consistent with the clarified spec. No change needed.

PR #603, opened 2026-06-04
#600 Fail gracefully on encrypted Parquet files

### LINKED ISSUE #600: Fail gracefully on encrypted Parquet files
Hardwood does not support Parquet Modular Encryption (PME), and currently fails in a confusing way for both on-disk encryption shapes:

**Encrypted-footer mode** (the default) replaces the `PAR1` magic with `PARE` at the start and end of the file. Hardwood reports:

```
Error reading file: [file.parquet] Not a Parquet file (invalid magic number at start)
```

This is misleading — the file *is* a Parquet file, it is just encrypted.

**Plaintext-footer mode** keeps the `PAR1` magic and a readable footer, but sets `FileMetaData.encryption_algorithm` (thrift field 8); the column data is encrypted. Hardwood silently ignores the field, so `info`/`schema`/`footer` appear to succeed, and only reading data fails — with a raw, unattributable stack trace:

```
java.io.UncheckedIOException: [file.parquet] Failed to scan pages for column 'value'
	at ...SequentialPageIterator.hasNext(SequentialFetchPlan.java:314)
```

We don't support decryption yet, but we should detect encryption up front and fail with a clear, file-attributed error in both modes. Detection points:

- `PARE` magic (start/end) → encrypted footer
- `encryption_algorithm` present in the footer → plaintext footer

Fixtures for both modes can be generated via `simple-datagen.py` using `pyarrow.parquet.encryption`. Note AES-GCM uses a random nonce per write, so the generated files are not byte-stable across reruns.

gunnarmorling:
Closes #600.

Hardwood does not support Parquet Modular Encryption. Until decryption lands (#128), an encrypted file should be rejected up front with a clear, file-attributed error rather than the two confusing failures it produced before:

- **Encrypted-footer mode** (the default): magic is `PARE`, not `PAR1`, so the reader reported a misleading `Not a Parquet file (invalid magic number at start)`.
- **Plaintext-footer mode**: the footer parses fine (`PAR1` magic) but carries `encryption_algorithm` and the column data is encrypted, so `info`/`schema` appeared to succeed and only a later data read crashed with an unattributable `Failed to scan pages for column 'value'` stack trace.

## Approach

Detection happens during metadata reading, so every entry point — the reader API, the CLI commands, and the `dive` TUI (all of which read the footer up front) — fails the same way before any data is touched:

| Change | What |
|---|---|
| `EncryptedParquetException` (new, internal) | `IOException` subtype carrying the canonical message |
| `ParquetMetadataReader` | detects the `PARE` start/end magic; enriches the plaintext-footer case with file context |
| `FileMetaDataReader` | throws on the `encryption_algorithm` footer field (thrift field 8) |
| `FooterCommand` (CLI) | the low-level `footer` command reports `PARE` clearly instead of "trailing magic bytes not found" |
| `simple-datagen.py` | generates `encrypted_footer.parquet` + `encrypted_plaintext_footer.parquet` |
| `error-handling.md` | documents the limitation under the `IOException` row |

All paths now surface: `[file] Encrypted Parquet files are not supported (Parquet Modular Encryption)`.

## Tests

- `EncryptedFileTest` — both modes via `ParquetFileReader.open`
- `FooterCommandTest` — new case for the `PARE` low-level path
- Verified end-to-end that `info`/`schema`/`print`/`footer`/`dive` all fail gracefully; full `core` suite green.

## Note on fixtures

These are the first fixtures in `simple-datagen.py` that are **not byte-stable** across reruns — AES-GCM draws a fresh nonce per write. That is fine for a graceful-failure assertion (it checks the error, not the bytes), and is called out in the script header so a regeneration diff isn't mistaken for a regression.


gunnarmorling:
Hey @mghildiy, just for reference, put this in place for 1.0 to gracefully reject encrypted files rather than raising some confusing exception.

Issue #607, opened 2026-06-04
Parse SizeStatistics and NaN counts (unencoded byte size, level histograms, nan_count)


gunnarmorling:
## Context

Parquet's `SizeStatistics` (in `ColumnMetaData` and per-page in the `ColumnIndex`) carries:

- `unencoded_byte_array_data_bytes` — total unencoded size of BYTE_ARRAY data
- `repetition_level_histogram` / `definition_level_histogram`

Hardwood currently does not parse these. They are stable in the format (parquet-format ≥ 2.10; arrow-cpp enabled writing them by default, discussed Jan 2025 on dev@parquet).

The same wave of format additions also brought NaN counts, which Hardwood likewise skips: `Statistics.nan_count` (field 9) and `ColumnIndex.nan_counts` (field 8) — the latter sitting directly after the two histogram fields in the same reader.

## Why it's useful

- **`hardwood dive`**: show true unencoded/uncompressed size for BYTE_ARRAY columns without decoding the data.
- **Read planning / selectivity**: level histograms let the reader estimate null/empty counts and nesting cardinality per page or chunk, useful for projection and predicate planning.
- **NaN counts**: visibility in `hardwood dive`, and better filter work.

## Scope

- Add a `SizeStatistics` record under `metadata`, parsed in `ColumnMetaDataReader` (field 16) and the `ColumnIndex` reader (field 6).
- Support NaN counts in both `Statistics` and `ColumnIndex`.
- Expose it on the corresponding public metadata types where the existing statistics are exposed.
- Usage docs under `docs/content/` for any new public surface.

Read-only / additive; no behavioural change to existing paths.

Good first issue candidate.

nitrogen404:
@gunnarmorling I'd like to work on this issue.

nitrogen404:
@gunnarmorling Does #607 include reading the NaN counts (Statistics field 9, ColumnIndex field 8), or is that a separate issue?

iifawzi:
Hey @nitrogen404, perhaps I can shime in here. First of all thanks for your interest, and you can for sure work on it! 
I have also updated the issue, I believe NaN counts could be included as part of this yes, thanks for pointing out to that

nitrogen404:
@iifawzi thank you for the clarification! I'll open a PR shortly.

PR #613, opened 2026-06-04
#608 Treat ColumnIndex field 7 as definition_level_histograms, not geospatial stats

### LINKED ISSUE #608: ColumnIndexReader misinterprets field 7 (definition_level_histograms) as geospatial statistics
## What

`ColumnIndexReader` maps `ColumnIndex` field id **7** to `list<GeospatialStatistics>`:

https://github.com/hardwood-hq/hardwood/blob/main/core/src/main/java/dev/hardwood/internal/thrift/ColumnIndexReader.java (`case 7` plus the class-level Javadoc, which states `7: geospatial_stats`).

But the canonical [`parquet.thrift`](https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift) defines:

```thrift
struct ColumnIndex {
  1: required list<bool> null_pages
  2: required list<binary> min_values
  3: required list<binary> max_values
  4: required BoundaryOrder boundary_order
  5: optional list<i64> null_counts
  6: optional list<i64> repetition_level_histograms
  7: optional list<i64> definition_level_histograms
  8: optional list<i64> nan_counts
}
```

Field 7 is `definition_level_histograms` (`list<i64>`). `GeospatialStatistics` is referenced **only** by `ColumnMetaData` field 17 — `ColumnIndex` has no geospatial field at all.

## Impact

Not currently exploitable for data corruption: the reader guards the geospatial branch with `listHeader.elementType() == 0x0C` (STRUCT), and a real `list<i64>` carries compact element type `0x06`, so genuine def-level histograms fall through to the skip path. So there is no crash today — but:

- The field mapping and the reader's class Javadoc are wrong and misleading.
- `definition_level_histograms` is silently dropped rather than deliberately skipped.
- The dead struct-decode branch would misfire on a malformed/adversarial file that places struct-typed elements at field 7.

## Fix

Remove the geospatial branch from `ColumnIndexReader` and treat fields 6/7/8 as `list<i64>` (parse as part of histogram support, or skip them cleanly until then). Correct the class-level Javadoc.

## Related

- #607 — parsing the rep/def level histograms + unencoded byte size belongs there; this issue is the narrower correctness fix for the mis-numbered field.

gunnarmorling:
Fixes #608.

`ColumnIndexReader` mapped `ColumnIndex` Thrift field **7** to `list<GeospatialStatistics>`, but in `parquet.thrift` field 7 is `definition_level_histograms` (`list<i64>`). `ColumnIndex` has **no** geospatial field at all — `GeospatialStatistics` is referenced solely by `ColumnMetaData` (field 17).

The mis-numbered field had grown a whole dead, public layer: a `ColumnIndex.geospatialStatistics` record component and a `PageFilterEvaluator` page-level spatial-pruning path. Because a real `list<i64>` carries compact element type `0x06` (never the struct `0x0C` the branch needed), that component was always `null` from any real file and the page path pruned nothing in production — it only ever "worked" in in-memory unit tests.

## Changes
- **`ColumnIndexReader`** — removed the field-7 geospatial branch; fields 6/7/8 (`repetition_level_histograms`, `definition_level_histograms`, `nan_counts`) now skip cleanly. Corrected the class Javadoc to the real spec field names. (Parsing the histograms is left to #607.)
- **`ColumnIndex`** — removed the unreachable public `geospatialStatistics` component.
- **`PageFilterEvaluator`** — `GeospatialPredicate` now returns `RowRanges.all` (Parquet has no per-page geospatial stats to prune on); deleted `evaluateSpatialIntersectionPages` / `evaluateSpatialPages`.
- **Tests** — added `ColumnIndexReaderTest` (the reader had none): pins that fields 6/7/8 are skipped without disturbing 1–5, plus the adversarial struct-typed-field-7 case. Removed the in-memory-only page-geospatial tests from `PageFilterEvaluatorTest`.
- **Docs** — `how-to/geospatial.md` no longer claims page-level geospatial pushdown or a per-page `ColumnIndex.geospatialStatistics()`.

**Row-group** geospatial filtering (via `ColumnMetaData.geospatialStatistics()`, field 17) is unaffected and still works.

## Note on test-first
As the issue states, this defect is not exploitable — the old struct-decode and a clean skip consume identical bytes, and the buggy output was funneled into the now-removed `geospatialStatistics` field, so there is no observable red→green. The new reader test instead pins the corrected contract and exercises the path that replaced the dead branch.

### DIFF
--- a/FORMAT_COVERAGE.md
+++ b/FORMAT_COVERAGE.md
@@ -95,7 +95,7 @@ All fields (column_idx, descending, nulls_first) ❌ — struct not read.
 | 14 | bloom_filter_offset | 🟡 | shown in dive; no bloom-filter decode (#105) |
 | 15 | bloom_filter_length | 🟡 | #105 |
 | 16 | size_statistics | ❌ | #607 |
-| 17 | geospatial_statistics | ✅ | row-group / page filter evaluators |
+| 17 | geospatial_statistics | ✅ | row-group filter evaluator (no per-page geospatial stats exist) |
 
 ### PageEncodingStats
 All fields (page_type, encoding, count) ❌ — struct not read.
@@ -176,7 +176,7 @@ Empty struct; not read (see `PageHeader.index_page_header`).
 | 4 | boundary_order | ✅ | |
 | 5 | null_counts | ✅ | |
 | 6 | repetition_level_histograms | ❌ | #607 |
-| 7 | definition_level_histograms | ❌ | #607; currently mis-read as geospatial (#608) |
+| 7 | definition_level_histograms | ❌ | skipped; surfacing tracked by #607 |
 | 8 | nan_counts | ❌ | #607 |
 
 ### OffsetIndex
@@ -220,7 +220,7 @@ The ❌ rows cluster into a handful of capabilities, cross-referenced to ROADMAP
 
 - **Modular encryption** — entire feature stubbed to fail-fast. #128 (ROADMAP has no phase yet).
 - **Bloom-filter pushdown** — ROADMAP 9.3 / 9.4; #105.
-- **Size statistics & level histograms** — ROADMAP 9.x; #607, #608.
+- **Size statistics & level histograms** — ROADMAP 9.x; #607.
 - **Statistics completeness** — distinct_count, exactness flags, nan_count; #483, #607.
 - **Declared sort order** — `sorting_columns`, `is_sorted`; ROADMAP 4.2.
 - **Column orders** — float total-order vs type-defined; #483.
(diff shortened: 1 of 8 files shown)


gunnarmorling:
Hey @mghildiy, this fixes an issue I encountered while mapping our coverage of the different fields in the Parquet Thrift spec. Specifically, the initial geospatial PR read geospatial statistics at the page level, they don't exist there actually though (i.e. we read an unrelated field).

A bit bummed I didn't notice this during the original review, but it is what it is. Curious though how that came to be in the first place if you remember? Why did we read that field at a level where it doesn't exist? Also, any thoughts on this PR welcome. Thanks!

mghildiy:
> Hey @mghildiy, this fixes an issue I encountered while mapping our coverage of the different fields in the Parquet Thrift spec. Specifically, the initial geospatial PR read geospatial statistics at the page level, they don't exist there actually though (i.e. we read an unrelated field).
> 
> A bit bummed I didn't notice this during the original review, but it is what it is. Curious though how that came to be in the first place if you remember? Why did we read that field at a level where it doesn't exist? Also, any thoughts on this PR welcome. Thanks!

I can't remember exactly I would dig into original work to understand it in detail.

PR #632, opened 2026-06-06
#624 Remaining row filtering for column readers

### LINKED ISSUE #624: Column-reader `.filter()` must return exact results — no client-side residual (distinct ColumnReaders and grouped ColumnReaders)
## Summary

The column-reader filter (`buildColumnReader(...).filter(pred)` and `buildColumnReaders(...).filter(pred)`) returns a **superset** of the matching rows: it prunes row groups/pages whose statistics prove no match, but still yields *every* row of the surviving pages. The caller is therefore silently required to re-apply the predicate ("client-side residual filtering") to get correct results. Aggregating directly over the filtered output produces **wrong answers with no error**.

A method called `filter(lt("event_time", 25_000))` that returns rows where `event_time >= 25_000` is a footgun, and silently returning wrong results contradicts the project's "fail early / no silent failures" stance. **Client-side residual filtering must not be required.**

This is a **major blocker for 1.0.0.CR2.**

## Current behaviour

`ColumnReaderBuilder.filter(FilterPredicate)` is documented as:

> Apply a column-statistics filter — row groups whose statistics prove no row matches are skipped.

In practice it also skips at page level via the column index, but the result is still page-granular, not row-exact. The grouped `ColumnReadersBuilder.filter(...)` carries the same contract. Only the **row reader** (`buildRowReader().filter()` → `FilteredRowReader`) filters exactly today.

## Reproduction

From `hardwood-benchmarks` (a generated file with monotonic `event_time`, ~40k rows/page):

```java
ColumnReader et  = reader.buildColumnReader("event_time").filter(FilterPredicate.lt("event_time", 25_000)).build();
ColumnReader amt = reader.buildColumnReader("amount").filter(FilterPredicate.lt("event_time", 25_000)).build();
// iterate, sum amount
```

- Expected: 25,000 rows.
- Actual: **40,000 rows** (the whole surviving page region), and the `amount` sum is correspondingly wrong.

The error is invisible when the threshold happens to land on a page boundary (e.g. at 50M rows a 2.5M threshold returned exactly 2.5M), which makes it especially dangerous — it passes casual testing and fails on arbitrary predicates.

## Requirement

After this change, a `ColumnReader` configured with `.filter(pred)` must yield **only** the rows satisfying `pred` — exact, byte-for-byte equivalent to the row reader's filtered result — with **no client-side residual**. This must hold:

1. **For grouped `ColumnReaders`** — `buildColumnReaders(columns(...)).filter(pred)`; every `getColumnReader(name)` in the group returns only matching rows, row-aligned across columns.
2. **For distinct single `ColumnReader`s** — independently built `buildColumnReader(a).filter(pred)` and `buildColumnReader(b).filter(pred)` each return only matching rows, and their per-batch counts agree.
3. Regardless of whether `pred` references the reader's own column, another projected column, or a column not otherwise read.

## Design considerations

- **Self-column predicate** (read `x`, filter on `x`): trivial — drop non-matching values from the decoded batch.
- **Cross-column predicate** (read `amount`, filter on `event_time`): requires decoding the predicate column, evaluating it exactly, building a **selection** (row mask / index list), and applying that selection to the payload columns — i.e. **late materialization**. This is the architectural fork tracked in #500; the `ColumnBatchMatcher`/`MergePlan` machinery already used by `FilteredRowReader` is reusable as the predicate-evaluation step.
- **Distinct readers** are the hard case: two independently built readers cannot coordinate, so each would have to decode and evaluate the predicate column itself — redundant decode of the predicate column once per reader. To avoid that, the computed selection (per row group, per predicate) should be **shared/cached** on the owning `ParquetFileReader` so sibling readers reuse it — see #74 (composable predicate cache) / #70 (predicate scan cache). The grouped `ColumnReaders` can compute the selection once and share it directly.
- Correctness is the bar here; the redundant-decode optimization for distinct readers can follow as long as results are exact.

## Acceptance criteria

- [ ] `buildColumnReader(payload).filter(pred-on-other-column)` returns only matching rows; a direct aggregate over the output is correct with no residual.
- [ ] Same for `buildColumnReaders(...).filter(...)` across all columns in the projection.
- [ ] Distinct readers and grouped readers return identical row counts/sums for the same file + predicate, and match the row reader's filtered result.
- [ ] Tests cover a **non-page-aligned** threshold (the case that exposed this), multiple row groups, a predicate column **not** in the projection, and self-column vs cross-column predicates.
- [ ] `ColumnReaderBuilder.filter` / `ColumnReadersBuilder.filter` Javadoc updated to state exact semantics (replacing the "row groups … are skipped" wording).

## Related

- #500 — drain-side vs late-materialization architecture fork (the mechanism this needs).
- #74, #70 — predicate caches (to make the distinct-reader case avoid redundant predicate-column decode).

gunnarmorling:


core/src/main/java/dev/hardwood/reader/ColumnReader.java
```diff
@@ -478,6 +522,183 @@ public void close() {
...
+        }
+        recordCount = count;
+        // Invalidate per-batch caches so accessors recompute over the compacted batch.
+        realViewComputed = false;
+        currentRealView = null;
+        cachedRealValues = null;
+        cachedRealBinaryBytes = null;
+        cachedRealBinaryOffsets = null;
+        cachedBinaries = null;
+        cachedStrings = null;
```
iifawzi: nit: we cn move to shared method to use it in the other place

iifawzi (approved):
looks good to me

core/src/main/java/dev/hardwood/reader/ColumnReader.java
gunnarmorling: Good idea, done.

gunnarmorling:
Thanks a lot for reviewing, @iifawzi! For sure glad we found this one before the 1.0 Final. Noticed this one, and also #634, while doing some benchmarking for the announcement.

iifawzi:
> Thanks a lot for reviewing, @iifawzi! For sure glad we found this one before the 1.0 Final. Noticed this one, and also #634, while doing some benchmarking for the announcement.

Yes pretty good catch, with this fix now it opens the door for skip/head support on columnar side as well if we want

PR #638, opened 2026-06-06
#596 Recognize legacy MAP_KEY_VALUE-only annotation as a map

### LINKED ISSUE #596: Confirm legacy MAP_KEY_VALUE-only annotation reads correctly
Surfaced by the spec conformance audit (#565), "Decisions for the maintainer".

## Gap

Parquet's MAP encoding has a legacy form in which the *inner* repeated group carries the converted-type annotation `MAP_KEY_VALUE` (in addition to, or instead of, `MAP` on the outer group). Modern writers annotate only the outer group as `MAP`. Older parquet-mr / Hive / Impala files may emit `MAP_KEY_VALUE` on the inner group with no annotation on the outer.

The conformance audit could not determine whether Hardwood reads this case correctly because **no fixture exercises it**. The code path may already work, may degrade gracefully to a list-of-structs view, or may misclassify.

## Acceptance

- Add (or locate in the `parquet-testing` corpus) a fixture where a MAP column is annotated `MAP_KEY_VALUE` on the inner group and **not** annotated `MAP` on the outer group.
- Test asserts `getColumn(...).logicalType()` reports `MapType`, `rowReader.getMap(name)` returns a `PqMap` with the expected key/value entries, and `SchemaNode.GroupNode.isMap()` returns `true`.
- If reads currently fail, fix to honour the legacy annotation. If reads already work, the test pins the behaviour against future regression.

iifawzi:
Fixes #596.

## Problem

Parquet's MAP encoding has a legacy form in which only the inner repeated `key_value` group carries the `MAP_KEY_VALUE` converted type, with **no** `MAP` annotation on the outer group. Older parquet-mr / Hive / Impala writers emit this form.

Hardwood keyed map recognition entirely off the outer group's `MAP` / `MapType` annotation, so these files fell through to a struct view: `getColumn(...).logicalType()` reported no `MapType`, `SchemaNode.GroupNode.isMap()` returned `false`, and `rowReader.getMap(...)` failed. No fixture exercised the case, so the gap was invisible.

## Fix

`FileSchema` now synthesizes a `LogicalType.MapType` on the outer group when its sole child is a `REPEATED` group carrying `MAP_KEY_VALUE`. Because every map-detection site (the reader's `TopLevelFieldMap` / `PqListImpl` / `NestedLevelComputer`, plus the CLI `schema`/`convert`/`dive` commands) consults `isMap()` / `logicalType()`, the legacy form is now handled identically to a MAP-annotated group with no scattered changes. The inner `key_value` group (two children) is never itself mistaken for a map.

## Tests & fixtures

- New hand-crafted fixture `map_annotation_legacy_key_value_test.parquet` (outer group unannotated, inner `key_value` group = `MAP_KEY_VALUE`), generated via a new `annotate_map_as_legacy_key_value` helper in `parquet_annotators.py` (PyArrow 24 emits neither annotation on the inner group).
- `MapLogicalTypeRecognitionTest` now runs all assertions across the legacy variant alongside `both annotations` and `modern-only`, and adds a read test asserting `getMap("attrs")` returns the expected `a→1, b→2` entries plus the null row.

## Docs

`docs/content/reference/accessors.md` documents that the legacy MAP encoding is recognized.


iifawzi:
Quick wins

gunnarmorling (approved):

gunnarmorling:
Rebased and applied; thanks a lot, @iifawzi!

PR #652, opened 2026-06-08
#645 Provide clearer error message when opening a non-existent file

### LINKED ISSUE #645: Provide clearer error message when opening a non-existent file
When opening a non-existent file with the CLI, we just say "Error reading file: <path>"; so it's not clear whether that file actually couldn't be parsed or whether it doesn't exist. We should raise a clearer error message if it's the latter. Noticed with the CLI, might affect core too, though.

yashpriyadarshan:
Fix #645 — Provide clearer error message when opening a non-existent file

Added a Files.exists() check in FileMixin.toInputFile() to proactively throw a clear "File not found" ParameterException across all CLI commands.
Updated the local *CommandTest.java files to assert this new error message.

## Checklist

- [x] The commit message is prefixed with the issue key ("#645 ")

### DIFF
--- a/cli/src/test/java/dev/hardwood/cli/command/ConvertCommandContract.java
+++ b/cli/src/test/java/dev/hardwood/cli/command/ConvertCommandContract.java
@@ -111,6 +111,8 @@ interface ConvertCommandContract {
     default void failsOnNonexistentFile() {
         Cli.Result result = Cli.launch("convert", "-f", nonexistentFile(), "--format", "csv");
 
+        assertThat(result.errorOutput()).contains("File not found");
+        assertThat(result.errorOutput()).contains(nonexistentFile());
         assertThat(result.exitCode()).isNotZero();
     }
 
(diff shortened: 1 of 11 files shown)


cli/src/main/java/dev/hardwood/cli/command/FileMixin.java
```diff
@@ -64,7 +64,12 @@ InputFile toInputFile(RangeBacking rangeBacking) {
                 return null;
             }
         }
-        return InputFile.of(Path.of(file));
+        Path localPath = Path.of(file);
+        if (!Files.exists(localPath)) {
```
gunnarmorling: Instead of solving this here in the CLI module, it may be better to move this logic to the `InputFile` implementations. This would then make sure a meaningful message is then also raised for library users (if not the case yet), and it would address files on S3 too (see _s3_ module).

gunnarmorling (commented):
Thanks a lot, @yashpriyadarshan! One comment inline.

gunnarmorling:
Hey @yashpriyadarshan, are you still planning to follow up on this one?

yashpriyadarshan:
Hey @gunnarmorling, sorry for the delay. I was busy with my university exams and completely forgot to follow up on this. Thanks for the reminder. I'll work on it and update the PR soon.

gunnarmorling:
No worries at all, just wanted to check in. All the best for your exams!

yashpriyadarshan:
Hey @gunnarmorling, I've updated the PR and moved the validation logic to the `InputFile` implementations. Let me know if this looks good to you.

PR #666, opened 2026-06-13
#657 Handle Parquet MAPs without a value field (key-only)

### LINKED ISSUE #657: MAP without a value field (key-only key_value) throws IndexOutOfBoundsException
## Summary

The Parquet [LogicalTypes spec](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#backward-compatibility-rules) permits a MAP whose `key_value` group has **no `value` field**:

> If the `value` field is not present, it can be represented as a map with all null values or as a set of keys.

Hardwood does not support this. The reader assumes the `key_value` group has exactly two children, so a key-only group fails with a raw `IndexOutOfBoundsException` rather than reading the data (as a key-set / all-null-values map) or rejecting it with a clear, file-attributed error.

## Where

- `TopLevelFieldMap.buildMapDesc` (`core/src/main/java/dev/hardwood/internal/reader/TopLevelFieldMap.java`) unconditionally reads `keyValueGroup.children().get(1)`.
- `SchemaNode.mapChild` (`core/src/main/java/dev/hardwood/schema/SchemaNode.java`) returns `null` when `children().size() != 2`, so `getMapValue()` cannot express "no value".

## Expected

Either surface a key-only MAP as a map with all-null values (spec option 1), or fail early with a clear, file-attributed error. A raw `IndexOutOfBoundsException` violates the fail-early / no-confusing-errors principle.

## Repro / test fixture

PyArrow always writes a `value` field, so a fixture must be hand-crafted via footer surgery (cf. `tools/parquet_annotators.py`): write a normal `map<int,int>`, then drop the `value` schema element and fix `path_in_schema`. A failing test should assert the chosen behavior (read-as-null-values, or a typed exception with the file name).

## Context

Found while auditing the LogicalTypes backward-compatibility rules during #656 (unannotated repeated fields). Gap 1 from that audit (the legacy two-level list-of-lists element rule) is fixed in the #656 branch; this is the remaining gap.

mohamedibrahim54:
The Parquet LogicalTypes spec permits MAP fields where the `key_value` group contains only a `key` field, representing a set of keys or a map with all-null values. Hardwood previously failed to read these files with an IndexOutOfBoundsException because it unconditionally expected two children.

This change:
- Updates SchemaNode to return null for mapChild when the value field is missing.
- Modifies TopLevelFieldMap and PqMapImpl to handle missing value projections safely (surfacing them as null).
- Updates NestedBatchIndex to return true for isElementNull when no projection column exists.
- Adds a new test fixture and MapKeyOnlyTest to verify end-to-end recognition and reading of key-only maps as spec-compliant null-valued maps.

closes #657

## Checklist

- [x] The commit message is prefixed with the issue key ("#657")


gunnarmorling:
Thanks a lot, @mohamedibrahim54! Pushed one more commit to do a few doc adjustments and expand test coverage. We're at the CR phase for 1.0 atm., but I think we're good to include this one, as it's a very targeted change, strictly improving a code path which would have thrown an exception before. Merging.

Issue #577, opened 2026-06-03
Pin multi-file boundary cells in BuilderCombinationTest (skip + tail)


gunnarmorling:
Workstream of #548. The matrix in #575 enumerates pairwise `RowReaderBuilder` combinations on a single-file fixture, but the multi-file boundary cells are currently parked behind `@Disabled` in `BuilderCombinationTest.multiFileBoundaries` because there is no multi-file fixture in the corpus yet:

```java
@Test
@Disabled("multi-file boundary cells need a multi-file fixture — see session task")
void multiFileBoundaries() {
    // skip + multi-file: indexes into the first file's rows only.
    // tail + multi-file: currently UnsupportedOperationException.
}
```

Two cells to pin:

- **`skip` + multi-file** — documented to index into the first file's rows only (cross-file skip is out of scope per `RowReaderBuilder.skip` JavaDoc). Assert the rows yielded come from file 0 starting at the skip offset; later files are unaffected.
- **`tail` + multi-file** — currently throws `UnsupportedOperationException`. Assert the rejection (instance + message substring), consistent with how every other rejection cell is pinned in `BuilderCombinationTest`.

### Acceptance criteria
- A small multi-file fixture (≥ 2 files) lives under `core/src/test/resources/` or the existing `differential/` corpus, generated by `tools/simple-datagen.py`.
- `BuilderCombinationTest.multiFileBoundaries` (or replacement parameterised cells) is enabled, exercises both combinations, and the test class still runs clean.
- The TODO/`@Disabled` reference to "the session task list" in `BuilderCombinationTest` is removed once the cells are live.

### Out of scope
- Multi-file behaviour of `head`, projection, filters, or the differential harness — those compose over the existing single-file matrix; this issue is only the two `skip` / `tail` × multi-file boundary cells the matrix currently can't assert.

gunnarmorling:
Context for whoever picks this up: neither boundary cell reflects a *fundamental* limitation — both are scope cuts that trace to one fact, and it's worth knowing that before writing the asserts so the test pins behavior for the right reason.

[...]

PR #667, opened 2026-06-13
#577 Pin multi-file boundary cells in BuilderCombinationTest (skip + tail)

RayanRal:
Added minimal two-file fixture (rows 0-149 and 150-249, carrying their global position as  id ⁠) and enable no-filter skip and tail cells: `skip` resolves against the first file and then streams the rest in order, `tail` ⁠ is rejected for multi-file readers.

Pinning `skip` surfaced that the JavaDoc and how-to guide were wrong for the multi-file case: both claimed `skip >= totalRows` yields an empty reader, but a physical skip never carries across a file boundary — `skip` at or beyond the first file's row count drops file 0 and streams the remaining files in full, so a multi-file reader is never emptied this way. 
Corrected both to describe what test now locks in.

## Checklist


RayanRal:
I've addressed test-related comments. 
`tools/simple-datagen.py` - my change seems to match overall style of the file
doc comments - sound valid, but I'm not entirely confident in that - open to changing, but will wait on human verdict.

gunnarmorling:
Thanks a lot, @RayanRal! Could you please add the two new fixture files to this PR, too?

gunnarmorling:
Squashed and applied; thanks a lot, @RayanRal!

Issue #670, opened 2026-06-14
Vectorize the split-block bloom filter membership check


iifawzi:
## Goal

Provide a SIMD implementation of the split-block bloom filter probe (`SplitBlockBloomFilter.mightContain`) using the Java Vector API, behind the existing method seam, for workloads that perform many membership checks.

Depends on #669, which provides the scalar probe this builds on and optimizes.

## Context

A split-block block is 8 × 32-bit little-endian words = 256 bits, and the eight probes are independent — a natural fit for a 256-bit vector (`IntVector.SPECIES_256`, 8 lanes). The whole per-block check (compute eight masks, AND against the eight words, reduce) becomes branch-free in a single pass. The project already depends on `jdk.incubator.vector`, so this can be a native Vector API path rather than hand-rolled intrinsics.

Sketch:
- broadcast `key`, multiply by the eight salts (constant vector), `>>> 27` → eight bit positions, per-lane `1 << pos` → mask vector;
- load the block's eight words, `(words & masks) == masks`, `allTrue()` → present.

## Scope

- Add a vectorized path in `SplitBlockBloomFilter.mightContain`, selected when the species fits (8-lane / 256-bit) and the platform supports it; keep the scalar implementation as a fallback for other shapes/platforms.
- Treat the scalar implementation as the correctness oracle: a differential test asserts the vector and scalar paths agree across random hashes and bitsets.
- Add a benchmark (JMH or equivalent) comparing scalar vs. vector across realistic batch sizes (single probe; `IN`-list of N).

## Measure-first / acceptance

- Bloom filter probing happens at the **metadata level** — once per `(column chunk, predicate value)`, not per row — so the expected gain is small except for large `IN` lists across many chunks. This is explicitly a measure-first issue: land the vector path only if the benchmark shows a worthwhile win; otherwise document the finding and keep the scalar version.
- Correctness: the vector path must produce identical results to scalar for all inputs.

## Out of scope

- I/O-level optimizations (lazy / coalesced filter reads, hash-once-per-predicate) — those live with the pushdown integration in #105.
- XXH64 vectorization — per-value hashing isn't the bottleneck and doesn't vectorize cleanly per value; not pursued here.

Issue #105, opened 2026-03-02
Support predicate push-down via bloom filters


gunnarmorling:
Follow-up to #59 (row group statistics-based filtering).

Parquet files can include bloom filters per column chunk, enabling probabilistic membership testing. Unlike min/max statistics which only support range predicates, bloom filters can efficiently answer equality predicates (`EQ`) even when the value range is wide.

### Scope

- Parse bloom filter metadata and data from Parquet files
- Integrate bloom filter checks into `RowGroupFilterEvaluator` for `EQ` predicates
- Fall back gracefully when bloom filters are absent (existing statistics-based filtering still applies)

### Notes

- Bloom filters are stored separately from column chunk metadata (offset in `ColumnMetaData` field 14: `bloom_filter_offset`)
- The Parquet bloom filter spec uses Split Block Bloom Filters
- This complements rather than replaces statistics-based filtering — both can be applied together

mghildiy:
I couldn't find any entity by the name RowGroupFilterEvaluator. Has the name been changed?

PR #671, opened 2026-06-14
#669 Read and decode Parquet bloom filters

### LINKED ISSUE #669: Read and decode Parquet bloom filters + membership-check primitive
## Goal

Read and decode the bloom filter stored at a column chunk's `bloom_filter_offset`, and provide a membership-check primitive that answers *"might this value be present?"*.

This is the foundational **read path** for bloom filters. Using it during query evaluation to prune row groups / column chunks (predicate pushdown) is kept separate in #105 and is built on top of this.

The footer fields `bloom_filter_offset` / `bloom_filter_length` (`ColumnMetaData` fields 14/15) are already surfaced on `ColumnMetaData`; this issue covers reading the filter body those offsets point to.

## Scope

- **`BloomFilterHeader`** — parse the thrift `BloomFilterHeader` struct (`numBytes` plus the `algorithm` / `hash` / `compression` single-variant unions), with fail-early validation of wire types and required fields.
- **`BloomFilter` + reader** — read the header followed by the raw split-block bitset; expose the bitset as a zero-copy, read-only view of the underlying (memory-mapped) bytes, with no per-filter allocation/copy.
- **Split-block membership check (SBBF)** — given a 64-bit hash, probe the correct 32-byte block's eight little-endian words and report present/absent (no false negatives).
- **XXH64 hashing** — the hash the Parquet bloom filter spec mandates (seed 0), with allocation-free `long` / `int` primitives and a `byte[]` entry point for binary values.

## Out of scope (follow-up, #105)

- Consulting the bloom filter during query evaluation to prune column chunks for equality / `IN` predicates. That work owns the physical-type → hash dispatch, the "stats inconclusive → consult filter" check ordering, lazy/coalesced filter reads, and any user-facing wiring.
- No public API or CLI surface is added here; all types live under `dev.hardwood.internal.*`.

## Verification

End-to-end against a real PyArrow-written fixture (`bloom_filter_test.parquet`) carrying bloom filters on INT64, INT32, and variable-length STRING columns: assert no false negatives for every stored value (the STRING lengths exercise all XXH64 code paths) and that absent values are discriminated.

iifawzi:
Hello, this PR resolves #669, adding the foundation for bloom filter support: the header and bitset decoders, the
split-block membership check, and the hashing algorithm implementation. Logged #670 to follow up with SIMD improvement, and #105 should be next, after this lands for the actual pushdown support 

## Checklist


core/src/main/java/dev/hardwood/internal/thrift/BloomFilterHeaderReader.java
```diff
@@ -0,0 +1,103 @@
...
+                            + "' has wrong wire type 0x" + Integer.toHexString(type & 0xFF));
+        }
+        short saved = reader.pushFieldIdContext();
+        try {
+            ThriftCompactReader.FieldHeader variant = reader.readFieldHeader();
+            if (variant == null) {
+                throw new IllegalStateException(
+                        "Invalid BloomFilterHeader: union field '" + name + "' has no variant set");
+            }
+            reader.skipField(variant.type()); // consume the variant's value (the empty inner struct)
```
rionmonster, line 93: I'm wondering if we should also check that the variant here is a `TYPE_STRUCT` similar to the above type outer union check (e.g. `variant.type != TYPE_STRUCT`).

Thoughts?

core/src/main/java/dev/hardwood/internal/thrift/BloomFilterHeaderReader.java
```diff
@@ -0,0 +1,103 @@
...
+            throw new IllegalStateException(
+                    "Invalid BloomFilterHeader: required field '" + "numBytes"
+                            + "' has wrong wire type 0x" + Integer.toHexString(type & 0xFF));
+        }
+        return reader.readNonNegativeI32("BloomFilterHeader.numBytes");
+    }
+
+    private static short readUnionVariant(ThriftCompactReader reader, byte type, String name) throws IOException {
+        if (type != TYPE_STRUCT) {
+            throw new IllegalStateException(
```
rionmonster: I think it may be worth creating a separate helper(s) for these exceptions (e.g. `invalidBloodFilterHeader`) to avoid some of the exception preamble like "Invalid BloodFilterHeader: ..." and throwing the exceptions directly with the specific message.

core/src/main/java/dev/hardwood/internal/thrift/BloomFilterHeaderReader.java
iifawzi: agree, i believe better to fail early just in case it is corrupted, the following line is asserting on the stop byte, but worth asserting on the type here as well

core/src/main/java/dev/hardwood/internal/bloomfilter/XxHash64.java
```diff
@@ -0,0 +1,153 @@
...
+ *
+ *  Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package dev.hardwood.internal.bloomfilter;
+
+/// XXH64 hash, as used by the Parquet split-block bloom filter to turn a value's plain-encoded
+/// bytes into the 64-bit hash that [SplitBlockBloomFilter] probes. The seed is fixed at 0, the
+/// value Parquet specified for bloom filters. All arithmetic is unsigned 64-bit (Java `long`
+/// two's-complement wraparound is bit-identical); multibyte reads are little-endian.
+public final class XxHash64 {
```
rionmonster, line 14: I'm wondering if it's worth adding some additional tests for coverage for these `XxHash64` functions. If anything just as a regression test (although I don't suspect these are things we'll touch too often).

iifawzi: i was thinking about that, but noticed we're not used to have dedicated unit tests. it's not expected to be touched at all, in fact, i delegated the full implementation of this algo to claude, and relied on extending the e2e tets of the bloom filter reader and the fixture to ensure full coverage

that being said, and given the complexity of it, having a unit test wouldn't harm, given we both doubted the same, I will add it

core/src/main/java/dev/hardwood/internal/bloomfilter/SplitBlockBloomFilter.java
```diff
@@ -0,0 +1,52 @@
...
+    /// <a href="https://github.com/apache/parquet-format/blob/master/BloomFilter.md">BloomFilter.md</a>.
+    private static final int[] SALT = {
+            0x47b6137b, 0x44974d91, 0x8824ad5b, 0xa2b7289d,
+            0x705495c7, 0x2df1424b, 0x9efc4947, 0x5c6bfb31
+    };
+
+    /// Probes `bitset`, a read-only little-endian view of the filter's bytes. Reads are absolute, so
+    /// the buffer's position is neither used nor modified.
+    public static boolean mightContain(ByteBuffer bitset, long hash) {
+        int numBlocks = bitset.capacity() / BYTES_PER_BLOCK;
```
rionmonster, line 35: We may want to add a guard on capacity here, specifically to cover empty or values that aren't specifically multiples of `BYTES_PER_BLOCK`) as a safety value.

iifawzi: That's already guaranteed through the reader 
```
 if (numBytes == 0 || numBytes % 32 != 0) {
            throw new IllegalStateException(
                    "Malformed bloom filter: bitset size must be a positive multiple of 32 bytes but was "
                            + numBytes);
        }
```

do you think it's still worth it having it here?

rionmonster: Ah! I missed that one as I was looking at the filter in isolation when I made the comment. We should be good then and don't need the redundant check! 👍

rionmonster:
@iifawzi 

Great job! Overall it looks good and this will be an awesome improvement to the project. The `SplitBlockBloomFilter` is a _really_ cool bloom filter variety as well. 👍

rionmonster (approved):

iifawzi:
> @iifawzi
> 
> Great job! Overall, it looks good, and this will be an awesome improvement to the project. The `SplitBlockBloomFilter` is a _really_ cool bloom filter variety as well. 👍

Thanks! Yes, it turned out to be very interesting. I didn't know about that variant before

gunnarmorling:
Woot, woot, very cool! Gonna take a look very soon. In terms of timelines, are we on the same page that this would go into 1.1?

iifawzi:
> Woot, woot, very cool! Gonna take a look very soon. In terms of timelines, are we on the same page that this would go into 1.1?

I think this PR can be included, given it's just an internal detail that's not exposed through any interface. perhaps the predicate push down utilization of the filter is the one that can be kept for 1.1, however if we deicde that bloom filters utilization would be on 1.1, we can decide to avoid merging this for now as well if we want

gunnarmorling:
Yeah, at this point, close to the final, after two CRs, I'd prefer to minimize the changes we integrate to bug fixes, doc updates, etc. Aiming to get out 1.0 this week or latest next week.

The plan then would be to get started with 1.1 right away, with a first Alpha/Beta whenever we feel ready (including the Bloom filters fully functionally would be a great goal).

If too many concurrent changes pile up for post 1.0 until next week and the potential for merge conflicts rises, we can also create 1.0 release branch and start integrating 1.1 changes into main; but I don't think we're at that point yet.

iifawzi:
Totally aligned

core/src/main/java/dev/hardwood/internal/bloomfilter/XxHash64.java
gunnarmorling: Yeah, so in general I'm leaning towards end-to-end tests, i.e. reading/querying a file and asserting the outcome is the expected one. But having a few unit tests for components like this makes sense, in particular for edge cases.

gunnarmorling (approved):
LGTM, merging. Exciting, great to see this first step towards full Bloom filter support. Thanks a lot, @iifawzi!

PR #673, opened 2026-06-14
Design doc for FFM zero copy native decode pipeline

mghildiy:
## Checklist


gunnarmorling:
Hey @mghildiy, do we have an issue for tracking this?

gunnarmorling:
Overall, I think this needs still a fair bit of work. First, what exactly do why achieve, i.e. what are the expected benefits? Note that we already pool and re-use internal memory segments (see `LibdeflateDecompressor`). Once we have sorted that out, we should make the design doc here fully self-contained, i.e. if it salvages anything from a prior PR, then this should be added here directly.

PR #674, opened 2026-06-14
#640 Fail fast when dive has no interactive terminal

### LINKED ISSUE #640: hardwood dive should fail fast when stdout is not a TTY
## Problem

`hardwood dive` launches the JLine TUI unconditionally, even when stdout/stdin is not an interactive terminal. When there is no TTY, JLine's startup terminal-capability queries leak to whatever is on the other end of the pipe and nothing renders.

This is easy to hit via Docker without `-it`:

```
docker run --rm -v "$(pwd)":/repo -w /repo ghcr.io/hardwood-hq/hardwood:local dive -f some.parquet
```

Instead of the TUI, the user gets a stray escape sequence such as:

```
^[[?2027;4$y
```

(that is the *host* terminal answering a DECRQM query for mode 2027 that leaked out of the container, since stdin is not connected back).

## Expected

When stdout (or stdin) is not a TTY, `dive` should exit with a clear, actionable message rather than emitting terminal control sequences — e.g.:

```
Error: 'dive' requires an interactive terminal. Re-run attached to a TTY (with Docker: docker run -it ...).
```

## Notes

- Detection: check `System.console() != null` or the JLine terminal type before entering the TUI loop.
- Scope: `cli` / `dive` only; non-interactive subcommands (`info`, `print`, `schema`, …) are unaffected.
- Surfaced while testing the Docker distribution (#223 / PR #639).

rionmonster:
## Description

This PR will close #640 by addressing a bit of low-hanging fruit for the TUI. `hardwood dive` launched the JLine TUI even without an interactive terminal. When stdin or stdout is not a TTY — most easily a `docker run` without `-it` — JLine's startup queries leak to the other end of the pipe and nothing renders. These changes introduce a guard for the command and exits early with the following actionable message:

> Error: 'dive' requires an interactive terminal. Re-run attached to a TTY (with Docker: docker run -it ...).

## Key Changes
  - Guard `DiveCommand` on `System.console() != null`, before any file I/O so a non-interactive run never pays a (possibly remote) open. `System.console()` is `null` whenever either stream is redirected.
  - `--smoke-render` is exempt (it never enters raw mode), so the native-image smoke test is unaffected.
  - Documented the requirement in `docs/content/reference/cli.md`.

## Verification and Testing
  - `DiveCommandTest.failsFastWithoutTty` launches `dive -f <fixture>` without `--smoke-render`. The Surefire fork has no console, so the guard fires for real — the same path a `docker run` without `-it` hits — rather than being mocked. The existing smoke-render test doubles as proof the guard is correctly bypassed.
  - `cli`: full module suite green, 379 tests.

  ## Checklist

  - [x] User-facing behavior change documented (`docs/content/reference/cli.md`)

### DIFF
--- a/cli/src/main/java/dev/hardwood/cli/command/DiveCommand.java
+++ b/cli/src/main/java/dev/hardwood/cli/command/DiveCommand.java
@@ -73,6 +73,15 @@ public class DiveCommand implements Callable<Integer> {
 
     @Override
     public Integer call() {
+        // Without a TTY the JLine TUI can't run; checked before any file I/O.
+        // Smoke-render never enters raw mode, so it is exempt.
+        if (!smokeRender && interactiveTerminalUnavailable()) {
+            spec.commandLine().getErr().println(
+                    "Error: 'dive' requires an interactive terminal. "
+                            + "Re-run attached to a TTY (with Docker: docker run -it ...).");
+            return CommandLine.ExitCode.USAGE;
+        }
+
         // Dive re-reads the same byte ranges constantly (page-up/page-down
         // navigation, jump-to-end-then-back, the `t` toggle, etc.). Opt
         // into the sparse-tempfile range cache so those repeats hit a
@@ -107,6 +116,13 @@ public class DiveCommand implements Callable<Integer> {
                 logHandler.close();
             }
         }
+    }
+
+    /// Reports whether stdin or stdout is not attached to a terminal.
+    /// `System.console()` is `null` when either is redirected, including in
+    /// the native image.
+    private static boolean interactiveTerminalUnavailable() {
+        return System.console() == null;
     }
 
     /// Configures JUL logging for an interactive dive session.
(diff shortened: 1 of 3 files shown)


cli/src/main/java/dev/hardwood/cli/command/DiveCommand.java
```diff
@@ -109,6 +118,13 @@ public Integer call() {
         }
     }
 
+    /// Reports whether stdin and stdout are both attached to a terminal.
+    /// `System.console()` is `null` when either is redirected, including in
+    /// the native image.
+    private static boolean interactiveTerminalAvailable() {
```
iifawzi: very tiny: perhaps we can invert the condition to avoid the double negation here and in the call site

iifawzi (approved):

cli/src/main/java/dev/hardwood/cli/command/DiveCommand.java
rionmonster: Yup, I'm good with that. 👍

gunnarmorling:
Very nice; thanks a lot, @rionmonster!

Issue #595, opened 2026-06-04
Surface ColumnOrder; warn or reject when not IEEE754TotalOrder for floats


gunnarmorling:
Surfaced by the spec conformance audit (#565), "Decisions for the maintainer".

## Gap

The Parquet footer carries a `column_orders` field — a list of `ColumnOrder` per leaf — with two values: `TYPE_ORDER` (signed type-default ordering) and `IEEE754TotalOrder` (the IEEE 754 total order used for float / double statistics). Hardwood does not read this field. Float / double pruning unconditionally uses `Float.compare` / `Double.compare` (total order), which is the *intended* semantic when a writer emits `IEEE754TotalOrder` (or omits `column_orders` entirely — the implicit default for modern writers).

A file that explicitly sets `TYPE_ORDER` for a float column would have a different semantic, and Hardwood would silently apply the wrong order against its stats. Modern writers don't do this in practice, so it's safe today; but the discrepancy is **undetected** rather than handled.

## Acceptance

- Decode `column_orders` from the Thrift footer alongside the rest of the file metadata.
- For each float / double leaf where the matching `ColumnOrder` is anything other than `IEEE754TotalOrder` (and is not absent), either:
  - Log a `WARN` naming the column and the unexpected ordering (preferred — keeps reads working when stats aren't used), **or**
  - Reject the file at open time with a clear message if the maintainer prefers fail-early.
- A small synthetic fixture (generated via `tools/simple-datagen.py` or a manual Thrift-edited file) carries an explicit `TYPE_ORDER` on a float column; a test asserts the warn / reject path fires.
- No behavioural change for the common case (no `column_orders`, or `IEEE754TotalOrder` everywhere).

PR #682, opened 2026-06-21
#595 Surface ColumnOrder and apply it to float/double ±0 pruning

iifawzi:
Hey, I've been looking into #595, checking pyarrow, it's still not following the recommendation to use `IEEE_754_TOTAL_ORDER`, it writes `TYPE_ORDER`, yet it does follow [the spec](https://github.com/apache/parquet-format/blob/325b9f4a70663c88485d4d59fdc1a121d7497943/src/main/thrift/parquet.thrift#L1126-L1169) correctly, writing `+0.0` for a zero max and `-0.0` for a zero min, hence why it was working as expected

On the other side, for writers who don't normalise those bounds, the spec identifies it as a compatibility rule to apply under the type-defined order:

> If the min is +0, the row group may contain -0 values as well.
> If the max is -0, the row group may contain +0 values as well.

Since it's being mentioned as a compatibility rule, I'd rather honor it than drop accuracy with a warning, so in this PR, I chose to widen the bounds to `[-0.0, +0.0]` for the type-defined order before pruning float/double statistics. columns with `IEEE_754_TOTAL_ORDER` order prune exactly with no widening

No strong opinion, though, happy to revert to the warning approach if you feel strongly about it, but since the spec frames it as a reader compatibility rule, following it keeps old/non-compliant files correct rather than just flagging them.

I'm not sure, but if we agree on shipping it in this way, perhaps it should land on 1.0 instead of 1.1? 

## Checklist

### DIFF
--- a/FORMAT_COVERAGE.md
+++ b/FORMAT_COVERAGE.md
@@ -34,7 +34,7 @@ is read and displayed yet feeds no filtering). The read-vs-skipped axis matches 
 | 4 | row_groups | ✅ | |
 | 5 | key_value_metadata | ✅ | exposed on `FileMetaData` |
 | 6 | created_by | ✅ | displayed in CLI / dive |
-| 7 | column_orders | ❌ | stats sort order assumed type-defined; #483 |
+| 7 | column_orders | ✅ | decoded onto `FileMetaData.columnOrders`; #595 |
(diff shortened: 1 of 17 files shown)


gunnarmorling:
Hey, thanks a lot for tackling this, agreed on all accounts. Let's get it in for 1.0.0.Final; while most writers should normalize, I don't see any reason for not doing it right no our end right now. Merging.

PR #696, opened 2026-06-24
#695 Force-include MAP key and VARIANT siblings in sub-field projections

### LINKED ISSUE #695: Projection resolver drops a MAP's mandatory key and a Variant's sibling leaves — sub-field projection into map values / variants is unreadable
While building Avro projection tests for #692 we found that Hardwood's projection
resolver under-collects columns when a projection path descends into a `MAP`
value or a `VARIANT`, because Hardwood projects by **leaf name** rather than by
**sub-schema**.

## Background: why parquet-java doesn't hit this

The two engines express projection differently:

- **parquet-java** projects with a requested **schema** (a `MessageType` subset,
  e.g. `conf.set(ReadSupport.PARQUET_READ_SCHEMA, ...)` /
  `AvroReadSupport.setRequestedProjection`). Because the projection is a
  well-formed schema tree, a `MAP`'s required `key` is necessarily present — you
  cannot write a valid map node without it.
- **Hardwood** projects with a flat list of **leaf names**
  (`ColumnProjection.columns("people.key_value.value.age")`). The resolver
  collects only the named leaf; nothing forces a special group's required sibling
  columns to come along.

So parquet-java gets the `key` "for free" from the structural invariant; Hardwood
has no such invariant and must add it explicitly.

### Case 1 — MAP value sub-field throws

`map_struct_value_test.parquet`: `id INT32, people map<string, struct{name, age}>`.

```java
try (ParquetFileReader r = ParquetFileReader.open(InputFile.of(file));
     RowReader rows = r.buildRowReader()
             .projection(ColumnProjection.columns("people.key_value.value.age"))
             .build()) {
    rows.next();
    rows.getMap("people");   // throws
}
```

```
java.lang.ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 1
```

### Case 2 — VARIANT sub-field returns a null Variant (silent wrong data)

`variant_shredded_test.parquet`: `var` is a shredded Variant; row 0 holds INT64(42).

```java
try (ParquetFileReader r = ParquetFileReader.open(InputFile.of(file));
     RowReader rows = r.buildRowReader()
             .projection(ColumnProjection.columns("var.typed_value"))
             .build()) {
    rows.next();
    rows.getVariant("var");   // returns null instead of INT64(42)
}
```

No exception — just a null Variant, which violates the fail-early principle.

## Root cause

`ProjectedSchema.create()`
(`core/src/main/java/dev/hardwood/internal/schema/ProjectedSchema.java:53`),
invoked from the row-reader path (`ParquetFileReader.java:355/389/490`), resolves
a dotted path in `resolveNestedColumn()` (`:160`) by navigating to the target
node and calling `collectColumnsFromNode()` (`:210`), which walks only the
subtree rooted at that node. It has no notion that:

- a `MAP` / `MAP_KEY_VALUE` group's `key` leaf is required whenever any part of
  the map is projected, or
- a `VARIANT` group must contribute *all* of its leaves when any one is projected
  (a Variant is read atomically).

The same gap exists on the simple-column path (`resolveSimpleColumn`, `:125`).

## Suggested fix

In the resolver, when a projected path lands on / descends into:

- a **MAP** — force-include the `key` leaf of the enclosing `key_value` group;
- a **VARIANT** group (`SchemaNode.GroupNode#isVariant()`) — collect all of its
  leaves, not just the navigated one.

`SchemaNode.GroupNode` already exposes `isMap()` / `isVariant()`, so both shapes
are detectable while walking.

## Impact

- MAP value sub-field projection: hard crash (`AIOOBE`) at read time.
- VARIANT sub-field projection: silent wrong data (null Variant).
- Affects the core `RowReader` directly, and therefore the Avro reader (#692),
  CLI, and every `ColumnProjection` consumer.

## Reproduction

Runnable repro branch: **iifawzi/hardwood@`reproduce-core-projection-defect`**
(https://github.com/iifawzi/hardwood/tree/reproduce-core-projection-defect)

- `core` `ColumnProjectionTest.mapValueSubFieldProjectionShouldReadKeysAndValue`
  and `variantSubFieldProjectionShouldReassemble` — fail today; should pass once
  fixed.
- `parquet-testing-runner` `MapValueProjectionComparisonTest` — green:
  `parquetJavaReadsKeysAndNarrowsValue` shows real parquet-java reading the same
  map value sub-field *with* keys, alongside `hardwoodFailsToProjectMapValueSubField`
  pinning the current throw.

iifawzi:
Closes #695.

## Summary

`ColumnProjection` lists leaf paths, not a schema subset, so the resolver has no structural invariant to lean on the way parquet-java's requested `MessageType` does. As a result, projecting a sub-field of a `MAP` value or a `VARIANT` left the reader unable to assemble those columns:

- `columns("people.key_value.value.age")` dropped the MAP's mandatory `key` leaf → `ArrayIndexOutOfBoundsException` at `PqMapImpl.create` when reading the map.
- `columns("var.typed_value")` dropped `metadata` and `value` → `getVariant("var")` returned `null` (silent wrong data — also affects unshredded rows that carry their payload in `value`).

The resolver now walks the schema post-order and, for any `MAP`/`VARIANT` group with at least one projected descendant, pulls in the leaves it cannot be materialized without: a `MAP`'s `key` column, and every leaf of a `VARIANT` (read atomically). This is opt-in via a new `completeContainers` parameter — the row-reader paths in `ParquetFileReader` pass `true`; the columnar `ColumnReader` / `ColumnReaders` paths keep passing `false` and read leaves literally.

While in `ProjectedSchema`, also fixed a separate coherency bug in `resolveSimpleColumn`: a bare-name match landing on a nested leaf (e.g. `columns("age")` matching `people.key_value.value.age` by `leafName()`) now registers the leaf's top-level ancestor field via `fieldPath().topLevelName()`, so the projection is addressable at the row level instead of yielding `getFieldCount() == 0`.

## Test plan

Four new tests in `ColumnProjectionTest` and one in `PredicatePushDownTest`:

- [x] `mapValueSubFieldProjectionShouldReadKeysAndValue` — `columns("people.key_value.value.age")` reads keys + ages across all rows of `map_struct_value_test.parquet` (including the empty-map row).
- [x] `simpleNameMatchingNestedLeafShouldReadMap` — `columns("age")` (bare name, no dot) resolves to the same projection, with `getFieldCount() == 1` and field name `"people"`.
- [x] `variantSubFieldProjectionShouldReassemble` — `columns("var.typed_value")` reassembles correctly across all 4 rows of `variant_shredded_test.parquet`, including row 2 (unshredded `BOOLEAN_TRUE` carried in `value`) — the load-bearing case that fails if only `typed_value` is pulled in.
- [x] `mapOfVariantSubFieldProjectionShouldReadBothSiblings` — `columns("var_map.key_value.value.value")` against `variant_in_repeated_test.parquet` exercises the post-order traversal on a compound shape: variant siblings (`metadata`) and the outer MAP's `key` are both force-included in one pass.
- [x] `testFilterOnMapSubFieldIsRejected` — pins the safety net that the `SelectionEngine` filter path relies on: `FilterPredicate.eq("people.key_value.value.age", 30)` is rejected at resolution time because the column is repeated, so the literal-projection path used for predicate columns is structurally unreachable.
- [x] Full core test suite passes (`./mvnw -pl core test` → 1384 tests).

### DIFF
--- a/core/src/main/java/dev/hardwood/internal/schema/ProjectedSchema.java
+++ b/core/src/main/java/dev/hardwood/internal/schema/ProjectedSchema.java
@@ -51,6 +51,24 @@ public final class ProjectedSchema {
     /// @return a projected schema containing only the selected columns
     /// @throws IllegalArgumentException if a projected column name is not found in the schema
     public static ProjectedSchema create(FileSchema schema, ColumnProjection projection) {
+        return create(schema, projection, false);
+    }
+
+    /// Resolves a [ColumnProjection] against `schema`.
+    ///
+    /// When `completeContainers` is true, special groups are expanded to the
+    /// leaves they cannot be materialized without: a MAP's `key` column and every
+    /// leaf of a VARIANT (read atomically). Row assembly requires this; the
+    /// columnar [ColumnReader] / [dev.hardwood.reader.ColumnReaders] paths read
+    /// individual leaves and pass false to keep the projection literal.
+    ///
+    /// @param schema the file schema
+    /// @param projection the requested columns
+    /// @param completeContainers whether to pull in required sibling leaves of
+    ///        projected MAP / VARIANT groups
+    /// @return the resolved projection
+    public static ProjectedSchema create(FileSchema schema, ColumnProjection projection,
+            boolean completeContainers) {
         if (projection.projectsAll()) {
             return createAllColumnsProjection(schema);
         }
@@ -77,12 +95,27 @@ public final class ProjectedSchema {
             }
         }
 
-        // Sort by original index to maintain column order
-        includedOriginalIndices.sort(Integer::compareTo);
-        includedFieldIndices.sort(Integer::compareTo);
+        // Enforce structural invariants on special groups: a MAP cannot be
+        // assembled without its key column, and a VARIANT is read atomically.
+        // Pull in the required sibling leaves whenever any part of such a group
+        // is projected, then rebuild the leaf list in column order.
+        boolean[] includedLeaf = new boolean[originalCount];
+        for (int idx : includedOriginalIndices) {
+            includedLeaf[idx] = true;
+        }
+        if (completeContainers) {
+            enforceGroupInvariants(schema.getRootNode(), includedLeaf);
+        }
 
-        // Remove duplicates from field indices
-        includedFieldIndices = new ArrayList<>(includedFieldIndices.stream().distinct().toList());
+        includedOriginalIndices = new ArrayList<>();
+        for (int i = 0; i < originalCount; i++) {
+            if (includedLeaf[i]) {
+                includedOriginalIndices.add(i);
+            }
+        }
+
+        // Sort and de-duplicate field indices
+        includedFieldIndices = new ArrayList<>(includedFieldIndices.stream().sorted().distinct().toList());
 
         // Build projected arrays
         int projectedCount = includedOriginalIndices.size();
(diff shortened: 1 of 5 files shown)


gunnarmorling:
Thanks a lot, @iifawzi. LGTM. Added one more commit for adding another test (projecting the key of a map does _not_ pull in the value). And I've logged #700 for actually implementing projection of VARIANT sub-fields.

PR #697, opened 2026-06-24
#692 Narrow AvroRowReader's materialized schema to the projection

### LINKED ISSUE #692: AvroRowReader throws on subset column projection — materializeRow() drives off the full file schema
## Summary

`AvroReaders.buildRowReader(fileReader).projection(subset)` throws at read time for any projection that isn't "all columns." The underlying `RowReader` is correctly narrowed to the projected columns, but the Avro materialization loop iterates the **full** file schema and tries to read every column — so the first non-projected column fails.

## Repro

```java
     AvroRowReader r = AvroReaders.buildRowReader(fr)
             .projection(ColumnProjection.columns("VendorID", "trip_distance"))
             .head(1)
             .build()) {
    r.next();   // throws
}
```

```
Column not in projection: tpep_dropoff_datetime
```

Reproduced against released `1.0.0.CR2` (NYC Yellow Taxi `yellow_tripdata_2026-01.parquet`).

## Root cause

`AvroReaders.RowReaderBuilder.build()` forwards the projection to the row reader but derives the Avro schema from the **whole** file:

```java
// AvroReaders.java:114
Schema avroSchema = AvroSchemaConverter.convert(fileReader.getFileSchema());  // full schema
return new AvroRowReader(rowReader, avroSchema);                              // projected reader
```

`AvroRowReader.materializeRow()` then iterates every field of that full schema and reads each from the projected reader:

```java
// AvroRowReader.java:78
private GenericRecord materializeRow() {
    GenericRecord record = new GenericData.Record(avroSchema);
    for (Schema.Field field : avroSchema.getFields()) {   // all columns
        if (rowReader.isNull(name)) { ... }               // throws for non-projected column
        record.put(field.pos(), materializeValue(rowReader, name, field.schema()));
    }
}
```

The reader is projected (subset); the record template is the full schema — and the materializer asks the reader for columns it doesn't have.

Secondary symptom: `getSchema()` returns the full file schema even under a subset projection, since the schema is never narrowed.

## Suggested fix

Derive the Avro schema from the **projected** schema, so `materializeRow()` only iterates projected fields — this also fixes `getSchema()` to reflect the projection. (Alternatively, skip non-projected fields during materialization, but that leaves `getSchema()` wrong.)

## Test gap

`AvroRowReaderTest` has no test exercising `.projection()` on the Avro path (only `filter` + defaults), which is why this wasn't caught. A regression test should project a subset, assert `getSchema()` contains only those fields, and read a record.

## Impact

A subset projection is unusable on the Avro path today — the only projection that doesn't throw is all-columns. Filter and head/tail are unaffected.

iifawzi:
Closes #692.

Depends on #696 — once that lands, this PR will reduce to a single commit. The MAP / VARIANT sub-field projection tests here rely on the resolver fix from #696.

## Summary

`AvroReaders.buildRowReader(fileReader).projection(subset)` threw at read time because the row reader was narrowed to the projection while `AvroSchemaConverter.convert(fileReader.getFileSchema())` still produced an Avro schema covering the **whole** file — materialization then walked the unprojected columns and tripped on the first one.

The converter now takes the projection and prunes recursively:

- top-level fields: drop any field whose subtree has no projected leaf;
- structs: same rule applied to each child;
- lists: prune the element subtree, so `items.list.element.quantity` narrows the element record to just `quantity`;
- maps: prune the value subtree (the key stays — the core resolver force-includes it, see #696);
- Variants: left at the canonical `{metadata: bytes, value: bytes}` since they are read atomically — projecting any leaf pulls the whole column.

`AvroRowReader.getSchema()` therefore mirrors the partial rows the row reader actually serves, instead of advertising fields the reader can't materialize.

## Test plan

13 new tests in `AvroRowReaderTest`, all reusing core fixtures under `core/src/test/resources`:

- [x] `readSubsetProjectionTopLevel` — `columns(\"value\")` on `plain_uncompressed.parquet`: schema has only `value`, 3 rows read correctly.
- [x] `readNestedFieldProjection` — `columns(\"address.city\")`: `address` record narrows to a single `city` field.
- [x] `readHead` / `readTail` — pin the existing head/tail builder controls on the Avro path.
- [x] `readProjectionWithFilter` / `readProjectionWithFilterAndHead` — project + filter + head, ensure the three controls compose.
- [x] `readWholeNestedGroupProjection` — `columns(\"address\")` (parent group, no dot) retains all 3 children.
- [x] `readListColumnProjection` / `readMapColumnProjection` — projecting just a list / map column keeps the column wholesale.
- [x] `readListOfStructSubFieldProjection` — `columns(\"items.list.element.quantity\")`: list-element record narrows to `quantity` only; reads 3 rows including the 2-element first row.
- [x] `readMapValueSubFieldProjection` — `columns(\"people.key_value.value.age\")`: map value record narrows to `age` only, key is force-included so the map reassembles. Depends on #696.
- [x] `readVariantSubFieldProjection` — `columns(\"var.typed_value\")`: schema stays `{metadata, value}` and reassembled value bytes match `[0x18, 42, 0, 0, 0, 0, 0, 0, 0]`. Depends on #696.
- [x] Full `AvroRowReaderTest` passes (22 tests).

### DIFF
--- a/avro/src/main/java/dev/hardwood/avro/AvroReaders.java
+++ b/avro/src/main/java/dev/hardwood/avro/AvroReaders.java
@@ -111,7 +111,7 @@ public final class AvroReaders {
                 underlying.tail(tailRows);
             }
             RowReader rowReader = underlying.build();
-            Schema avroSchema = AvroSchemaConverter.convert(fileReader.getFileSchema());
+            Schema avroSchema = AvroSchemaConverter.convert(fileReader.getFileSchema(), projection);
             return new AvroRowReader(rowReader, avroSchema);
         }
     }
--- a/avro/src/main/java/dev/hardwood/avro/internal/AvroSchemaConverter.java
+++ b/avro/src/main/java/dev/hardwood/avro/internal/AvroSchemaConverter.java
@@ -13,9 +13,11 @@ import java.util.List;
 import org.apache.avro.LogicalTypes;
 import org.apache.avro.Schema;
 
+import dev.hardwood.internal.schema.ProjectedSchema;
 import dev.hardwood.metadata.LogicalType;
 import dev.hardwood.metadata.PhysicalType;
 import dev.hardwood.metadata.RepetitionType;
+import dev.hardwood.schema.ColumnProjection;
 import dev.hardwood.schema.FileSchema;
 import dev.hardwood.schema.SchemaNode;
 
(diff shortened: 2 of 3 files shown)


iifawzi:
I rebased to main, so tests here will fail until #696 lands

gunnarmorling:
LGTM, merging. Thanks, @iifawzi!

PR #698, opened 2026-06-24
#693 Dispatch UINT_32 by physical type in AvroRowReader

### LINKED ISSUE #693: AvroRowReader throws ClassCastException on UINT32 columns — materializer reads by Avro type, not physical type
While building a test suite to exercise #692 we figured out that the Avro path
also mishandles unsigned 32-bit integers, unrelated to projection.

## Summary

Reading any `UINT32` column through `AvroRowReader` throws `ClassCastException`.
`AvroSchemaConverter.convertIntType()` maps `UINT32` to Avro `LONG` (correct — an
unsigned 32-bit value doesn't fit a signed Avro `INT`). But `materializeValue()`
then dispatches on the **Avro** type and calls `reader.getLong(name)` against a
column whose **physical** type is `INT32`. The flat reader's batch is an `int[]`,
so the cast to `long[]` fails.

## Repro

`unsigned_int_test.parquet`: `id UINT32, uint32_val UINT32, uint64_val UINT64`.

```java
     AvroRowReader r = AvroReaders.rowReader(fr)) {
    r.next();   // throws
}
```

```
java.lang.ClassCastException: class [I cannot be cast to class [J
    at dev.hardwood.internal.reader.FlatRowReader.getLong(FlatRowReader.java:425)
    at dev.hardwood.avro.AvroRowReader.materializeValue(AvroRowReader.java:96)
```

## Root cause

`materializeValue()` (and `materializeStructValue()` / `materializeListElement()`
/ `materializeMapValue()`) switch on the converted Avro schema type. For a
`UINT32` column the Avro type is `LONG`, so the `LONG` branch fires `getLong()` —
but the underlying storage is `INT32`. The value must be read with `getInt()` and
widened with `& 0xFFFFFFFFL` to recover the unsigned magnitude (e.g. `0xFFFFFFFF`
→ `4294967295L`, not `-1L`).

## Impact

`UINT32` columns are unreadable on the Avro path — top-level and nested (struct,
list, map values all share the same dispatch). `UINT64` is unaffected (already
`long[]`). This is a separate defect from the #692 projection bug.

## Test gap

`AvroRowReaderTest` had no test reading unsigned integer columns.

## Suggested fix

Decide the accessor by the source column's **physical** type, not the Avro type;
for `UINT32` read the `int` and apply the unsigned-widening mask before storing
the `Long` in the record.

iifawzi:
## Summary

- Fixes #693: `AvroRowReader` threw `ClassCastException` on any `UINT_32` column because the materializer dispatched on the Avro type (`LONG`, the widened form that fits `4294967295`) and called `getLong` against an `int[]`-backed batch.
- `AvroSchemaConverter` now tags the `LONG` schema produced from `UINT_32` with a `hardwood.unsignedInt32` marker, and the four `materialize*` sites in `AvroRowReader` (top-level, struct, list, map) read the int and widen with `Integer.toUnsignedLong` when the marker is present.

## Test plan

- [x] `AvroRowReaderTest.readUnsignedIntColumns` — top-level `UINT_32` / `UINT_64` against the existing `unsigned_int_test.parquet`, including `0xFFFFFFFF → 4294967295L` (the issue's repro) and the `UINT_64` wrap to `-1L`. Also asserts the marker prop is present on `UINT_32` schemas and absent on `UINT_64`.
- [x] `AvroRowReaderTest.readNestedUnsignedIntColumns` — `UINT_32` inside a struct, list, and map<string,_> against a new `unsigned_int_nested_test.parquet` fixture, exercising the three remaining dispatch paths.
- [x] `./mvnw verify -pl avro` passes (17 tests).


gunnarmorling:
Hey @iifawzi, I've pushed another commit, as that issue also manifested for two more column types. Can you take a look? Thanks!

iifawzi:
oh, nice additional catch, looks good to me

gunnarmorling:
Rebased and applied; thx, @iifawzi!

PR #701, opened 2026-06-25
#694 Use FLBA byte length for Avro fixed size

### LINKED ISSUE #694: AvroSchemaConverter emits wrong FIXED size for decimal / FIXED_LEN_BYTE_ARRAY columns
While building a test suite to exercise #692 we figured out that the Avro schema
converter produces incorrect (and sometimes invalid) Avro `fixed` sizes,
unrelated to projection.

## Summary

For a `decimal` stored as `FIXED_LEN_BYTE_ARRAY`, `AvroSchemaConverter` declares
the Avro `fixed` size as the column's **leaf index** rather than its byte length.
For a plain `FIXED_LEN_BYTE_ARRAY` it hardcodes `12`. Both are wrong; the decimal
case throws outright when the (bogus) size can't hold the precision.

## Repro

`compat_decimal_10_2.parquet`: `id INT64, amount decimal(10,2)` stored as
`FIXED_LEN_BYTE_ARRAY` of length **5**, `amount` at leaf index **1**.

```java
     AvroRowReader r = AvroReaders.rowReader(fr)) {   // throws during build()
}
```

```
java.lang.IllegalArgumentException: fixed(1) cannot store 10 digits (max 2)
    at dev.hardwood.avro.internal.AvroSchemaConverter.convertDecimalType(AvroSchemaConverter.java:188)
```

The `fixed(1)` is the column's leaf index (1), not its byte length (5).

## Root cause

```java
// AvroSchemaConverter.java:188 — decimal on FIXED_LEN_BYTE_ARRAY
return decimal.addToSchema(Schema.createFixed(prim.name(), null, null, prim.columnIndex()));
// AvroSchemaConverter.java:212 — plain FIXED_LEN_BYTE_ARRAY
case FIXED_LEN_BYTE_ARRAY -> Schema.createFixed(prim.name(), null, null, 12);
```

`Schema.createFixed`'s fourth argument is the **size in bytes**. The real length
lives in `ColumnSchema.typeLength`, but the `SchemaNode.PrimitiveNode` the
converter walks doesn't carry it — `columnIndex` was used as a stand-in, and the
plain-FLBA branch just assumes `12`.

## Impact

- Decimal-on-FLBA columns whose leaf index < required digit capacity throw at
  schema-conversion time; others silently get a wrong `fixed` size.
- Plain `FIXED_LEN_BYTE_ARRAY` columns of any length other than 12 get a wrong
  `fixed` size (silently produces an Avro schema that misdescribes the data).

## Test gap

`AvroRowReaderTest` had no test asserting the Avro `fixed` size of a decimal /
FLBA column.

## Suggested fix

Plumb the FLBA byte length (`typeLength`) into `SchemaNode.PrimitiveNode` (or have
the converter consult `ColumnSchema`) and use it as the `fixed` size for both the
decimal and plain `FIXED_LEN_BYTE_ARRAY` branches. `INT96` legitimately stays 12.

iifawzi:
Fixes #694.

`AvroSchemaConverter` declared the Avro `fixed` size from the wrong source. For a `decimal` on `FIXED_LEN_BYTE_ARRAY` it used the column's **leaf index** (so `compat_decimal_10_2.parquet` — `amount` at leaf index 1, byte length 5 — threw `fixed(1) cannot store 10 digits`); for a plain `FIXED_LEN_BYTE_ARRAY` it hardcoded `12`, silently misdescribing any column of a different width.

## Change

The converter is now instance-based, holding the `FileSchema`, so a primitive's real `type_length` is resolved from its `ColumnSchema` by leaf index — for both the decimal and plain FLBA branches. `INT96` keeps its legitimate fixed 12-byte width.

A fixed-length column with no declared `type_length` is malformed; `fixedByteLength()` throws rather than emit a zero-width `fixed`, matching `ByteStreamSplitDecoder` / `PlainDecoder`, which already reject the same condition.

## Tests

Two fixture-backed tests in `AvroRowReaderTest`, both of which fail on the pre-fix converter (verified):
- `decimalOnFixedUsesByteLengthAsFixedSize` — `compat_decimal_10_2.parquet`, asserts `fixed` size 5 and decimal(10,2).
- `plainFixedUsesByteLengthAsFixedSize` — `delta_byte_array_flba_test.parquet`, asserts `fixed` size 4 for required and optional FLBA columns.


avro/src/main/java/dev/hardwood/avro/internal/AvroSchemaConverter.java
```diff
@@ -201,19 +207,34 @@ private static Schema convertIntType(LogicalType.IntType i) {
...
             case INT96 -> Schema.createFixed(prim.name(), null, null, 12);
         };
     }
 
+    /// Resolve the declared byte length of a [PhysicalType#FIXED_LEN_BYTE_ARRAY]
+    /// column, looked up from its [dev.hardwood.schema.ColumnSchema] by leaf index.
+    /// A fixed-length column with no `type_length` is malformed; fail early rather
+    /// than emit a bogus zero-width Avro `fixed`, matching the decoders that reject
+    /// the same condition.
+    private int fixedByteLength(SchemaNode.PrimitiveNode prim) {
```
iifawzi, line 286: Caveat for reviewers: this is now the **third** copy of the "FLBA must declare `type_length`" guard — `ByteStreamSplitDecoder` and `PlainDecoder` already throw the same way at their point of use. We may want to unify it into a single early gate in `FileSchema.fromSchemaElements`, so the invariant holds for every consumer and `ColumnSchema.typeLength` is guaranteed non-null for FLBA columns.

The trade-off is strictness: validating at schema construction rejects a malformed file at **open**, even when the bad column is projected away — whereas the current point-of-use checks only fail if that column is actually decoded/converted. Kept it local here to match the existing decoder pattern and avoid regressing the projection case; flagging the unification as a possible follow-up rather than folding it into this fix.

iifawzi:
Now it should be ready @gunnarmorling if you can take a look

gunnarmorling:
LGTM overall. Claude found and fixed one remaining corner case (list of FIXED would have triggered the same exception during serialization). Merging. Thanks a lot, @iifawzi!

I think that's the last one you had for 1.0.0.Final, correct? I.e. I could go ahead and start the release?

iifawzi:
LET'S GOOOOOOOOOO

gunnarmorling:
Woot, woot, awesome!

Issue #98, opened 2026-03-02
InputFile abstraction: decouple read pipeline from memory-mapped files


gunnarmorling:
## Summary

Introduced an `InputFile` interface that decouples the entire read pipeline from `MappedByteBuffer`, enabling alternative file backends without further internal changes.

## Changes

- New `InputFile` interface (`dev.hardwood`) with `readRange(long, int)`, `length()`, `name()`
- `MappedInputFile` implementation preserving the existing mmap backend
- Widened all internal types from `MappedByteBuffer` to `ByteBuffer` (decompressors, page reader, page scanner, etc.)
- Added `InputFile`-accepting overloads on `ParquetFileReader.open()`, `Hardwood.open()`, and `Hardwood.openAll()`

## Motivation

This is foundational for:

- **Object store support (#31)** — users can implement `InputFile` for S3, GCS, Azure Blob, etc. with range-read semantics
- **Predicate push-down** — `InputFile.readRange()` enables reading only the column chunks and pages that survive filtering, rather than mapping the entire file upfront
- **In-memory reading** — wrapping a `byte[]` or `ByteBuffer` as an `InputFile` for testing or embedded use

## Design

See `designs/INPUT_FILE_ABSTRACTION.md` for the full design document.

No behavioral changes to the existing mmap path. All existing tests pass unchanged.

PR #703, opened 2026-06-25
#702 Allow sizing the decoder thread pool via the Hardwood entry point

### LINKED ISSUE #702: Cannot size the decoder thread pool via the Hardwood multi-file entry point
`Hardwood.create(int threads)` used to let callers size the page-decode thread pool when bootstrapping the multi-file reader. It was removed in b4a60824 (#98, "Introducing InputFile abstraction"), incidental to that commit's purpose.

The sizing plumbing still exists on the public `HardwoodContext.create(int)`, and single-file reads can use it via `ParquetFileReader.open(InputFile, HardwoodContext)`. But `Hardwood` only exposes a no-arg `create()` and its constructor takes the internal `HardwoodContextImpl`, so the documented multi-file entry point has no way to control decode parallelism.

Restore the capability by adding `Hardwood.create(HardwoodContext)`, letting callers build a context (sized as they like) and share it across the `Hardwood` instance and standalone `ParquetFileReader`s. The supplied context stays caller-owned and is not closed by `Hardwood.close()`, mirroring the existing `ParquetFileReader.openAll(files, context)` contract.

gunnarmorling:
Fixes #702.

`Hardwood.create(int threads)` used to let callers size the page-decode thread pool when bootstrapping the multi-file reader, but it was dropped in #98 ("Introducing InputFile abstraction"), incidental to that commit. The sizing plumbing survived on the public `HardwoodContext.create(int)` and remained reachable for single-file reads via `ParquetFileReader.open(InputFile, HardwoodContext)`, but `Hardwood` exposed only a no-arg `create()` whose constructor took the internal `HardwoodContextImpl` — so the documented multi-file entry point had no way to control decode parallelism.

## Change

- Add `Hardwood.create(HardwoodContext)`, accepting a caller-built (and caller-sized) context.
- The supplied context stays **caller-owned**: it is not closed by `Hardwood.close()`, so one context — and its thread pool — can be reused across reads and shared with standalone `ParquetFileReader`s. This mirrors the existing ownership contract of `ParquetFileReader.openAll(files, context)`.
- Docs updated: `how-to/multi-file.md` and `concepts/concurrency-model.md`.

This goes a step beyond the old `create(int)`: rather than only sizing a private pool, it lets the context be shared across entry points.

## Tests

- A read over a supplied 2-thread context, asserting both correct output and that the context survives `Hardwood.close()` (ownership contract).
- A null-context rejection test.

All `MultiFileRowReaderTest` cases pass and the full reactor compiles.

### DIFF
--- a/core/src/main/java/dev/hardwood/Hardwood.java
+++ b/core/src/main/java/dev/hardwood/Hardwood.java
@@ -24,18 +24,47 @@ import dev.hardwood.reader.ParquetFileReader;
 /// }
 /// ```
 ///
+/// To control the decode parallelism, or to reuse one context (and its thread
+/// pool) across many reads and standalone [ParquetFileReader]s, create the
+/// [HardwoodContext] yourself and pass it in:
+/// ```java
+/// try (HardwoodContext context = HardwoodContext.create(4)) {
+///     try (Hardwood hardwood = Hardwood.create(context)) {
+///         // ...
+///     }
+///     // context is still open here for further use
+/// }
+/// ```
+///
 /// For single-file usage, [ParquetFileReader#open(InputFile)] is simpler.
 public class Hardwood implements AutoCloseable {
 
     private final HardwoodContextImpl context;
+    private final boolean ownsContext;
 
-    private Hardwood(HardwoodContextImpl context) {
+    private Hardwood(HardwoodContextImpl context, boolean ownsContext) {
         this.context = context;
+        this.ownsContext = ownsContext;
     }
 
     /// Create a new Hardwood instance with a thread pool sized to available processors.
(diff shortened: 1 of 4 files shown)


gunnarmorling:
@iifawzi, just in case you have a minute: small change to bring back an API method that was accidentally dropped a while ago. It allows to customize the decoder thread pool size for multi-file usage.

iifawzi:
nice nice, thank you! I was relying [in the reader ](https://github.com/hardwood-hq/hardwood-examples/blob/4d41d8a5faf37ab4f3c2bcc4291ea08ea332e0a9/multi-file/src/main/java/dev/hardwood/examples/multifile/Main.java#L94)in the example

PR #705, opened 2026-06-26
Avoid column index fetches for unfiltered scans

### LINKED ISSUE #683: Skip ColumnIndex fetch on unfiltered scans
## Problem

`RowGroupIndexBuffers.fetch()` reads a single contiguous byte range spanning **both** the OffsetIndex and ColumnIndex regions of a row group, unconditionally, once per row group (`core/src/main/java/dev/hardwood/internal/reader/RowGroupIndexBuffers.java:53-79`).

The `OffsetIndex` is needed regardless of whether a predicate is present — the unfiltered fast path (`IndexedFetchPlan`) uses page locations to lay out column reads. The `ColumnIndex`, however, is only ever *decoded* when a filter is present: `RowGroupIterator.getSharedMetadata` calls `PageFilterEvaluator.computeMatchingRows` only when `filterPredicate != null` (`RowGroupIterator.java:285`).

Consequently, **an unfiltered scan fetches the entire ColumnIndex region — plus the gap between it and the OffsetIndex block — and never decodes a byte of it.** In the parquet-mr layout the ColumnIndexes form one block and the OffsetIndexes another; the ColumnIndex is by far the larger of the two (per-page min/max values vs. three longs per page), so this is the bulk of the wasted read.

This is the same class of optimization DataFusion landed in [apache/datafusion#22857](https://github.com/apache/datafusion/pull/22857) — see https://ratuld.me/blog/datafusion-parquet-page-index-skip/ — which defers/skips page-index loading when it cannot affect the result.

## Proposed change

Thread a `needColumnIndex` flag (= `filterPredicate != null`) from `RowGroupIterator.getSharedMetadata` into `RowGroupIndexBuffers.fetch()`. When false, compute the fetch range from `offsetIndexOffset`/`offsetIndexLength` entries only and skip slicing the ColumnIndex buffers. This skips the ColumnIndex block *and* the inter-block gap on unfiltered scans.

Scope is local — `RowGroupIndexBuffers.fetch()` and its single caller.

## Notes / non-goals

- We already coalesce into a single read, so this saves transferred **bytes**, not round-trips. The win is on object storage with wide / many-page schemas; on local files (page cache) it is negligible.
- The second DataFusion case — a predicate that row-group statistics already prove is *fully matched*, so page pruning cannot drop anything — is **out of scope** here. It would require making `RowGroupFilterEvaluator` three-valued (definitely-no / maybe / definitely-all-match) and has a narrower payoff; track separately if desired.

## Acceptance

- Unfiltered reads of a file whose ColumnIndex and OffsetIndex are non-adjacent fetch a strictly smaller index range (assert via a test counting bytes read / read ranges).
- Filtered reads are unchanged.

arnabnandy7:
## Summary

Fixes #683 by avoiding ColumnIndex metadata fetches during unfiltered scans.

Unfiltered reads still fetch OffsetIndex metadata so page locations can be planned efficiently, but they now skip ColumnIndex statistics bytes since those are only needed for page-level predicate pushdown. Filtered reads keep the existing behavior and fetch both index types.

## Changes

- Added an offset-only mode to `RowGroupIndexBuffers.fetch(...)`
- Updated `RowGroupIterator` to request ColumnIndex buffers only when a filter predicate is present
- Extended `CountingInputFile` test helper to track fetched bytes
- Added a test proving offset-only index fetches skip ColumnIndex buffers and read fewer index bytes

## Testing

- `git diff --check` passes
- `./mvnw clean verify` 

## Checklist

### DIFF
--- a/core/src/main/java/dev/hardwood/internal/reader/RowGroupIndexBuffers.java
+++ b/core/src/main/java/dev/hardwood/internal/reader/RowGroupIndexBuffers.java
@@ -47,6 +47,20 @@ public class RowGroupIndexBuffers {
     /// @param rowGroup  the row group whose indexes to fetch
     public static RowGroupIndexBuffers fetch(InputFile inputFile,
             RowGroup rowGroup) throws IOException {
+        return fetch(inputFile, rowGroup, true);
+    }
+
+    /// Fetches offset indexes for a row group, and optionally column indexes.
+    ///
+    /// Unfiltered scans need OffsetIndex bytes to plan page reads, but do not
+    /// need ColumnIndex statistics. Filtered scans include both so page-level
+    /// predicate pushdown can evaluate min/max/null-count metadata.
+    ///
+    /// @param inputFile the file to read from
+    /// @param rowGroup the row group whose indexes to fetch
+    /// @param includeColumnIndexes whether ColumnIndex buffers should be fetched
+    public static RowGroupIndexBuffers fetch(InputFile inputFile,
+            RowGroup rowGroup, boolean includeColumnIndexes) throws IOException {
 
         List<ColumnChunk> allColumns = rowGroup.columns();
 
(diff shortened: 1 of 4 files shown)


gunnarmorling:
Nice one, applied. Thanks a lot, @arnabnandy7! Note I've prefixed the commit message with the other key, as per the repo conventions. Can you do the same for your other pending PRs? Thx!

arnabnandy7:
> Nice one, applied. Thanks a lot, @arnabnandy7! Note I've prefixed the commit message with the other key, as per the repo conventions. Can you do the same for your other pending PRs? Thx!

I'll amend it once I'm back thanks for pointing out.

arnabnandy7:
@gunnarmorling both PR has been amended, please proceed with your review

PR #707, opened 2026-06-26
Make multi-file skip a global offset

### LINKED ISSUE #672: Make physical skip() a true global OFFSET across multi-file readers
## Background

Physical `skip(N)` (no filter) on a multi-file reader currently resolves against the **first file only** — it never carries across a file boundary. A `skip` at or beyond the first file's row count drops file 0 and streams the remaining files in full, so a multi-file reader is never emptied by `skip`. This is documented behaviour (see the warning box in `docs/content/how-to/query-controls.md` and the `skip` JavaDoc in `ParquetFileReader`) and is now pinned by `BuilderCombinationTest.multiFileBoundaries` (#577 / #667).

## Why the limitation exists

`openAll()` reads exactly one footer — file 0's. Files 1..N are opened lazily by the `RowGroupIterator` as iteration reaches them; their row counts are never read up front. The physical-`skip` resolution in `ParquetFileReader.buildRowReader(...)` runs at build time, before any iteration, so the only metadata available is `firstFileMetaData.rowGroups()`. The cumulative-`numRows()` walk therefore cannot extend the offset past file 0; when `skip` overshoots, it hands the iterator an empty first-file subset and the iterator streams files 1..N in full.

The limitation falls out of two deliberate properties: lazy per-file opening (O(1) open cost) and physical `skip` as a true seek (earlier row groups are never opened).

## Proposed change — Approach A: footer-walk across files

Make physical multi-file `skip(N)` a true global OFFSET over the concatenated relation:

- At skip-build time, read footers of files 0, 1, … in order until the cumulative row count exceeds `skip`, then position into that file's row groups and let the iterator stream the rest.
- Skipped files are never *data*-decoded — only their footers are read — so the seek property is preserved across files.

Cost / trade-offs:
- One footer read (an I/O round-trip on remote backends) per *skipped* file. Bounded by the number of files preceding the target, and only footers, not data.
- Gives up the current "only file 0's metadata is ever touched" invariant. The multi-file column-reader path and `getFileMetaData()` lean on first-file-only metadata today, so the footer-walk must stay scoped to the `skip` resolution and not leak into those.

A decode-and-discard alternative (stream from file 0, discard the first N rows like the *filtered* skip path already does) is explicitly **not** proposed: it would decode every skipped row and throw away the reason physical skip exists.

## Acceptance criteria

- [ ] Physical multi-file `skip(N)` returns rows starting at global position `N` (e.g. on the `multi_file_part0/1` fixture, `skip(160)` yields ids 160..249, not 150..249).
- [ ] Skipped files are not data-decoded (assert via the existing JFR/events or a read-byte probe where practical).
- [ ] `skip >= sum(rows across all files)` yields an empty reader (true global OFFSET past the end).
- [ ] Multi-file column-reader path and `getFileMetaData()` semantics are unchanged.
- [ ] Docs (`how-to/query-controls.md`, `skip` JavaDoc) and `BuilderCombinationTest.multiFileBoundaries` updated to the new semantics.

## Compatibility

Binary- and source-compatible (no signature change), but a **behavioural/semantic change**: `skip` results shift for multi-file readers that overshoot the first file. Today's behaviour is documented as a limitation ("to skip within a non-first file, open it separately"), not an intended feature to depend on, so this is a limitation-lift rather than a contract break. Target 1.1; call it out in the release notes.

Follow-up to #577 / #667.

arnabnandy7:
## Summary

Fixes #672.

This changes physical no-filter `skip(N)` on multi-file row readers to behave as a true global offset across the concatenated input files. Hardwood now walks file footers until it finds the target file/row group, then starts reading from there. Skipped files have their footers read for row counts, but their data pages are not fetched or decoded.

## Changes

- Resolves physical multi-file `skip(N)` globally instead of relative to file 0.
- Returns an empty reader when `skip >= total rows across all files`.
- Preserves filtered `skip(N)` as logical `OFFSET` over matching rows.
- Adds regression coverage for multi-file boundary behavior and skipped-file read counts.
- Updates JavaDoc, query-control docs, and release notes.

## Verification

- Manual compile checks passed for changed production/test files with `javac --release 21`.
- JShell smoke checks passed for:
  - `skip(160)` yielding ids `160..249`
  - `skip(250)` yielding empty
  - skipped file data not being read

## Checklist

### DIFF
--- a/core/src/main/java/dev/hardwood/internal/reader/RowGroupIterator.java
+++ b/core/src/main/java/dev/hardwood/internal/reader/RowGroupIterator.java
@@ -72,6 +72,7 @@ public class RowGroupIterator {
     private final List<InputFile> inputFiles;
     private final HardwoodContextImpl context;
     private final long maxRows;
+    private final long physicalSkip;
 
     /// Number of leading rows of the first row group to skip. Non-zero only on
     /// the tail-read fast path; consumed by [#computeFetchPlans] to synthesize
@@ -80,6 +81,7 @@ public class RowGroupIterator {
     /// be promoted from the construction-time value of `0` via
     /// [#setTailSkip(long)] once the gate decision is in.
     private long tailSkip;
+    private long firstRowGroupSkip;
 
     // Set after first file
     private FileSchema referenceSchema;
@@ -171,21 +173,46 @@ public class RowGroupIterator {
     ///        cross-column alignment.
     public RowGroupIterator(List<InputFile> inputFiles, HardwoodContextImpl context,
                             long maxRows, long tailSkip) {
+        this(inputFiles, context, maxRows, tailSkip, 0);
+    }
+
+    /// Creates a RowGroupIterator with tail-skip and physical-skip budgets.
+    ///
+    /// @param inputFiles one or more input files (must not be empty)
+    /// @param context the Hardwood context
+    /// @param maxRows maximum rows to read (0 = unlimited)
+    /// @param tailSkip leading rows of the first row group to skip via per-page
+    ///        masking (0 = unused)
+    /// @param physicalSkip leading physical rows to skip while building the
+    ///        work list (0 = unused). Whole row groups are dropped; the residue
+    ///        within the first kept row group is exposed via [#firstRowGroupSkip()].
+    public RowGroupIterator(List<InputFile> inputFiles, HardwoodContextImpl context,
+                            long maxRows, long tailSkip, long physicalSkip) {
         if (inputFiles.isEmpty()) {
             throw new IllegalArgumentException("At least one file must be provided");
         }
         if (tailSkip < 0) {
             throw new IllegalArgumentException("tailSkip must be non-negative, got " + tailSkip);
         }
+        if (physicalSkip < 0) {
+            throw new IllegalArgumentException("physicalSkip must be non-negative, got " + physicalSkip);
+        }
         this.inputFiles = new ArrayList<>(inputFiles);
         this.context = context;
         this.maxRows = maxRows;
         this.tailSkip = tailSkip;
+        this.physicalSkip = physicalSkip;
     }
 
     /// Returns the maximum rows limit (0 = unlimited).
     public long maxRows() {
         return maxRows;
+    }
+
+    /// Rows to discard from the first kept row group after whole row groups
+    /// before the physical skip target have been dropped from the work list.
+    public long firstRowGroupSkip() {
+        return firstRowGroupSkip;
     }
 
     /// Sets the reference schema and pre-prepared first file, skipping [#openFirst()].
@@ -821,7 +848,8 @@ public class RowGroupIterator {
         if (maxRows <= 0 || filterPredicate != null) {
             return 0;
         }
-        return Math.max(0, maxRows - workItem.rowsConsumedBefore());
+        long leadingSkip = workItem.workItemIndex() == 0 ? firstRowGroupSkip : 0;
+        return Math.max(0, maxRows - workItem.rowsConsumedBefore() + leadingSkip);
     }
 
     /// Returns the context.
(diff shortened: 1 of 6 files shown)


gunnarmorling:
Thanks a lot, @arnabnandy7! I had Claude review this PR, could you take a look at its findings below?

---

The behaviour and tests look right, but the main point is about *where* the global offset gets resolved.

### Resolve the skip lazily in `RowGroupIterator`, not at build time

The PR resolves the offset eagerly in `ParquetFileReader.buildRowReader(...)` — `locatePhysicalSkipTarget` walks footers 0..N at build time, then `buildRowReaderFromFile` injects file N as the iterator's "first file". The thing is, `RowGroupIterator.buildWorkList()` already does almost all of this lazily as it walks files: it opens each file one at a time (with the next file async-prefetched), reads each footer, validates schema compatibility, accumulates `rowsConsumed`, and already stops early when the `maxRows` `rowBudget` is exhausted — dropping whole row groups it doesn't need.

A global physical `skip` is the mirror image of that `maxRows` budget, and there's already a precedent for the residue handling in `tailSkip`. So the cleaner shape is to thread a `skip` budget into the iterator and, in `buildWorkList`, drop every row group whose cumulative end ≤ `skip` (recording the within-row-group residue on the straddling group). That computes the same target as the iterator naturally walks files — no separate build-time footer walk — and it keeps the async file prefetch that the current serial walk gives up. It also lets us delete `locatePhysicalSkipTarget`, `readMetadataForPhysicalSkip`, `buildRowReaderFromFile`, `buildEmptyRowReader`, and the duplicated `validateSchemaCompatibility`.

Let's go with this approach.

### Other findings (largely subsumed by the above, but noting them)

- **Duplicated `validateSchemaCompatibility`:** the new method in `ParquetFileReader` is a near-verbatim copy of `RowGroupIterator.validateSchemaCompatibility` (same four checks, same messages). Two copies will drift.
- **Duplicated metadata read:** `readMetadataForPhysicalSkip` duplicates `RowGroupIterator.openAndReadMetadata`, including the `FileOpenedEvent` JFR emission.
- **Serial footer reads:** the build-time walk opens skipped files serially with no prefetch, unlike the iterator's `loadFileAsync` / `triggerPrefetch` pipeline — a real cost on remote backends with many skipped files.
- **Byte-range filter across files:** `locatePhysicalSkipTarget` applies the `rowGroupFilter` to files 1..N, but a `ByteRange` predicate is file-local (it tests a row group's byte offset), so a range derived from file 0 has no meaning against file 1's offsets. Multi-file + byte-range split is unusual, but worth confirming this can't silently mis-drop row groups in a non-first file.

### Test gap

`physicalMultiFileSkipDoesNotReadSkippedFileData` only proves file 0's data isn't read — and file 0's footer was already read by `openAll`, so it never exercises the footer-walk-then-discard path against an actually-skipped file. A 3-file fixture where the **middle** file is skipped would cover it: assert the middle file's read count is footer-only (its data pages are never fetched) while the target in position 2 is reached.

arnabnandy7:
@gunnarmorling, it'll take sometime, since weekend is over. Thanks for this review I'll certainly work on it as soon as possible.

gunnarmorling:
Sure thing, no worries. Thanks a lot for your efforts!

arnabnandy7:
> Sure thing, no worries. Thanks a lot for your efforts!

@gunnarmorling Merge conflict is now resolved in this branch, and I validated that it doesn't break or impact the original fixes.

arnabnandy7:
> Thanks a lot, @arnabnandy7! I had Claude review this PR, could you take a look at its findings below?

Implemented the suggested change from the review. Physical multi-file skip is now resolved lazily inside RowGroupIterator during the normal file/row-group walk, so we no longer do a separate eager footer scan in ParquetFileReader.

Also removed the duplicated metadata/schema validation path and added a regression test for a skipped middle file to verify it only reads footer metadata, not data pages.

**Verification:**

```
.\mvnw -pl core -Dtest=BuilderCombinationTest clean test
.\mvnw -pl core -am test
```
Both passed.

arnabnandy7:
@gunnarmorling, please review the new changes

gunnarmorling:
@arnabnandy7, nice, this is looking good now. Could you just rebase this to current main and squash everything into one commit (we generally prefer rebases over merges for the sake of a linear commit history). Thanks!

arnabnandy7:
@gunnarmorling it's done, please check.

gunnarmorling:
Thanks a lot, @arnabnandy7. Pushed one more commit for some minor adjustments, merging now.

Issue #713, opened 2026-06-26
Unguarded 32-bit overflow in Variant object/array offset arithmetic


gunnarmorling:
## Bug

`VariantValueDecoder.parseObject` / `parseArray` compute `numElements * idSize` and `(numElements + 1) * offsetSize` in plain `int`; `VariantMetadata` does the same with `(dictionarySize + 1) * offsetSize`. A hostile `numElements` / `dictionarySize` overflows the 32-bit product, yielding a wrapped (possibly negative or small-positive) offset that bypasses bounds checks and points at an arbitrary buffer location.

parquet-java explicitly guards this with widened arithmetic (the "offset arithmetic bounds check").

## How it surfaced

The `int_overflow_in_bounds_check.parquet` fixture from [apache/parquet-testing#113](https://github.com/apache/parquet-testing/pull/113) is an object with `numElements = 0x33333333` and `offset_size = 4`, designed to overflow `numElements * offsetSize`. Hardwood currently rejects it only by accident — it hits a field-id-range check first (`IndexOutOfBoundsException: Field id … 255 (size=0)`) and never evaluates the overflowing arithmetic. The vulnerability the fixture targets is unguarded.

## Fix

- Perform the offset/id/size arithmetic in `long` (or validate each factor against the remaining buffer extent before multiplying) in `parseObject`, `parseArray`, and `VariantMetadata`.
- Reject with `IllegalArgumentException` when a declared count cannot fit the surrounding buffer.
- Add a unit test asserting rejection for the overflow layout.

## Related

- Hardening test harness: #711.

steveloughran:
@gunnarmorling the test data for int overflow has just been updated, fwiw...the java code wasn't checking for overflows properly and the test code wasn't triggering it

gunnarmorling:
Thanks, @steveloughran. To be sure, are you referring to the latest commit on https://github.com/apache/parquet-testing/pull/113 (["updated int overflow in bounds check test file"](https://github.com/apache/parquet-testing/pull/113/commits/1c520459baededbe41a0f7aaa96b42ba3dec3eaf))?

@rionmonster, tagging you for awareness. It's not quite clear to me yet, it seems there's some more work left to do here. Or maybe not, as we have our own fixture?

rionmonster:
@gunnarmorling 

[...]

Issue #711, opened 2026-06-26
Enable VARIANT decoder hardening tests once parquet-testing#113 lands


gunnarmorling:
## Goal

Commit the VARIANT hardening test work once [apache/parquet-testing#113](https://github.com/apache/parquet-testing/pull/113) (malformed-variant fixtures under `bad_data/variants/`) is merged, and use it to drive the decoder-hardening fixes tracked in the related issues below.

The work already exists on the local branch `test/parquet-testing-pr113-variant-hardening` (uncommitted):

- `ParquetTestingRepoCloner` temporarily repointed at the fork branch `steveloughran/parquet-testing@pr/variant-hardening`. **Revert this to `apache/parquet-testing` master** when the PR lands — the only reason it points at the fork is so the test can run before merge.
- New `VariantHardeningTest` (in `parquet-testing-runner`): one dynamic test per fixture.

### Note on what is actually under test

These fixtures do **not** carry a Parquet `VARIANT` logical-type annotation — they store the variant `metadata` and `value` as two plain `required binary` columns, because they come from parquet-java's variant *binary decoder* hardening tests (`TestHardenedReader`). So the unit under test is Hardwood's variant decoder, not the file-read path: the test reads the two buffers, feeds them into `PqVariantImpl`, and walks the whole value tree to force every lazy decode.

### Key finding: we reject, but mostly not for the right reason

parquet-java added an **eager, shallow validator** in the `Variant` constructor that rejects malformed structure up-front with `IllegalArgumentException` (or `UnsupportedOperationException` for the version) and a descriptive message. Hardwood has no such validator — `PqVariantImpl` is a lazy flyweight, so most fixtures are rejected only *incidentally*, with whatever low-level exception fires first during navigation. Comparing against `TestHardenedReader`'s expected exceptions, two fixtures are rejected for the *wrong* reason and mask real bugs, and three reject via raw JDK exceptions with no validation (one a genuine allocation-DoS path).

## Current behaviour vs. parquet-java reference (12 fixtures)

| Fixture | parquet-java expects | Hardwood actually throws | Verdict | Issue |
|---|---|---|---|---|
| `variant_version_2_header` | `UnsupportedOperationException` …"version" | `IAE` "Unsupported … version: 2" | ✅ correct | — |
| `out_of_range_dictionary_size` | `IAE` …"dictionary" | `IAE` "truncated before offset table" | ✅ reason ✓ | (msg → #715) |
[...]

PR #716, opened 2026-06-26
#712 Read Variant metadata offset-size from the correct header bits

### LINKED ISSUE #712: Variant metadata offset-size read off by one bit (bits 6-7, not 5-6)
## Bug

`VariantBinary.METADATA_OFFSET_SIZE_SHIFT` is `5`, so the metadata header's offset-size field is read as `(header >>> 5) & 0x3` (bits 5-6). The [Variant spec](https://github.com/apache/parquet-format/blob/master/VariantEncoding.md) places `offset_size_minus_one` at **bits 6-7**. parquet-java reads it as `((header >> 6) & 0x3) + 1` (`VariantUtil`).

This is a **correctness defect, not just hardening**: any variant whose metadata dictionary needs `offset_size >= 2` — i.e. the dictionary string section exceeds 255 bytes (many/long field names) — is misparsed (wrong `offset_size`, wrong `dictionary_size`, wrong offset table), producing garbage or an incidental exception.

It is masked across the entire existing test suite (125 shredded-variant byte-for-byte cases + the comparison suite) because real variant metadata almost always uses `offset_size = 1`, where bits 5-6 and 6-7 both read `0`.

## How it surfaced

The `negative_dictionary_size.parquet` fixture from [apache/parquet-testing#113](https://github.com/apache/parquet-testing/pull/113) uses header `0xC1` (`offset_size = 4`). Hardwood reads `offset_size = 3` and `dictSize = 0xFFFFFF`, then rejects with a *truncation* error — the file is rejected, but for the wrong reason, masking this parse bug.

Note that fixing the shift alone is not enough for this fixture to reject *for the right reason*: with `offset_size = 4` the `dictSize` field reads `0xFFFFFFFF`, i.e. a negative `int`. A non-negative / within-buffer dictionary-size guard is therefore part of this fix too.

## Fix

- Change the metadata offset-size shift to bits 6-7 (`(header >> 6) & 0x3`) and correct the layout comment in `VariantBinary`.
- Add the non-negative / within-buffer `dictionary_size` guard in the `VariantMetadata` header parse, so `negative_dictionary_size` rejects with a descriptive `IllegalArgumentException` ("unsigned int" / out-of-range size) rather than an incidental truncation/`AIOOBE`. This lives in the metadata parse that already runs when `PqVariantImpl` is built — not a new value-tree validation pass.
- Add a regression test with `offset_size >= 2` metadata (a well-formed file with a >255-byte dictionary section) that round-trips correctly.

## Related

- Hardening test harness: #711.

gunnarmorling:
Fixes #712.

## Problem

The Variant metadata header's `offset_size_minus_one` field lives in **bits 6-7**, but `VariantBinary` read it from bits 5-6 (`METADATA_OFFSET_SIZE_SHIFT = 5`). Every existing fixture uses `offset_size = 1`, where both readings yield `0`, so the bug was invisible across the whole suite (125 byte-exact shredded-variant cases + the comparison sweep). Any Variant whose dictionary string section exceeds 255 bytes — needing `offset_size >= 2` — was misparsed into garbage.

It surfaced via the `negative_dictionary_size` fixture from apache/parquet-testing#113, which uses `offset_size = 4`.

## Fix

- `METADATA_OFFSET_SIZE_SHIFT` 5 → 6 (+ corrected the layout comment).
- Guard against a 4-byte `dictionary_size` that reads back as a negative `int`: reject with a clear "not a valid unsigned int" message rather than letting the bogus size drive later arithmetic. (The `(dictSize+1)*offsetSize` overflow widening is intentionally left to #713 to avoid overlap.)

## Tests

- **Real end-to-end fixture** via `tools/simple-datagen.py`: `variant_metadata_offset_size2.parquet` — a VARIANT object whose metadata dictionary is 320 bytes, forcing `offset_size = 2`. `VariantLogicalTypeTest.readsVariantWhoseMetadataUsesOffsetSizeTwo` reads it through the VARIANT row API and resolves both long-named fields (`'a'*160` → INT8 5, `'b'*160` → BOOLEAN_TRUE).
- `VariantMetadataTest.negativeDictionarySizeRejected` for the guard.
- Both new tests **fail against the pre-fix shift** (verified) and pass with the fix.
- No regressions: the 125 shredded-variant byte-exact tests plus the variant decoder/encoder, row-API, and conversion tests are green.

## Notes

- Change is confined to `internal` packages — no public API or config change, so no usage-docs update.
- Release notes are curated at release time off the **1.0.1.Final** milestone (where #712 lives), so none added here.


core/src/main/java/dev/hardwood/internal/variant/VariantBinary.java
```diff
@@ -21,13 +21,13 @@ private VariantBinary() {}
     // ==================== Metadata header byte ====================
     //   bit 0-3: version (must be 1)
     //   bit 4:   sorted_strings flag
-    //   bit 5-6: offset_size_minus_one (0..3 → 1..4 bytes)
-    //   bit 7:   unused
+    //   bit 5:   unused
```
iifawzi, line 24: noice noice, i was confused as it's not specified in the docs, but found the correction here https://github.com/apache/parquet-format/pull/574

iifawzi (approved):
Looks good to me, nice catch, seems like it was gunnar vs variant today, seeing a couple of bugs!

gunnarmorling:
Hehe, yeah; #711 has the overview. This was triggered by a mail on parquet-dev where I learned about a pending addition to parquet-testing for a range of malformed VARIANT files ([context](https://lists.apache.org/thread/lpv7xqwfk8zlcxsmlw7qzb7408lh7ppt)). One of the fixtures revealed that parsing bug here.

PR #719, opened 2026-06-27
#718 Evaluate row-group statistics once and split row-group filter JFR events

### LINKED ISSUE #718: First file row group statistics are evaluated twice on column-readers path
We're evaluating the first file statistics twice when filters are applied, first eagerly in the reader https://github.com/hardwood-hq/hardwood/blob/5056f18d564a8eb531553bf371ca41f849a582e0/core/src/main/java/dev/hardwood/reader/ParquetFileReader.java#L552 second in the iterator https://github.com/hardwood-hq/hardwood/blob/5056f18d564a8eb531553bf371ca41f849a582e0/core/src/main/java/dev/hardwood/internal/reader/RowGroupIterator.java#L984

surfaced while looking to columnar jfr events, where second one seems useless given it has been already filtered on the first run: 
Image

iifawzi:
Fixes #718.

## Problem

On the column-readers path, the **first file's** row groups had their column statistics evaluated twice: once eagerly in `ParquetFileReader.filterRowGroups` and again in `RowGroupIterator`. The second pass re-ran `RowGroupFilterEvaluator.canDropRowGroup` over an already-pruned list — pure redundant work.

It was also observable in JFR: the first file emitted **two** `dev.hardwood.RowGroupFilter` events, and the second one reported the already-pruned count as its `totalRowGroups` rather than the file's real row-group count, so the record was self-inconsistent (every other file emitted one).

## Fix

- Statistics push-down now lives **solely in `RowGroupIterator`** — the single component that evaluates every file uniformly. `ParquetFileReader.filterRowGroups` is reduced to the cheap byte-range `RowGroupPredicate` pre-filter.
- The two filtering stages now emit **distinct** events, each self-describing and fired exactly once:
  - `dev.hardwood.RowGroupByteRangeFilter` — byte-range split selection (`ParquetFileReader`)
  - `dev.hardwood.RowGroupFilter` — statistics predicate push-down (`RowGroupIterator`)

## Behavior changes (intended)

- A fully unfiltered `columnReaders(projection)` read no longer emits a row-group filter event — consistent with the `RowReader` path, which already emitted none.
- Byte-range split pruning is now reported under `RowGroupByteRangeFilter` instead of `RowGroupFilter`. Documented in `reference/configuration.md`.

## Testing

- New `RowGroupFilterEventTest` (written failing-first) asserts each stage emits exactly one event and that `totalRowGroups` is the file's real count — guards the regression directly.
- `RowGroupFilterTest`, `PredicatePushDownTest`, `ColumnReadersTest`, `ColumnReaderExactFilterTest` (118 tests) all pass; single-file split correctness (`columnReaderByteRangeKeepsTwoComplementaryHalvesExactly`) is unaffected.

### DIFF
--- a/core/src/main/java/dev/hardwood/jfr/RowGroupByteRangeFilterEvent.java
+++ b/core/src/main/java/dev/hardwood/jfr/RowGroupByteRangeFilterEvent.java
@@ -1,0 +1,46 @@
+/*
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Copyright The original authors
+ *
+ *  Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package dev.hardwood.jfr;
+
+import jdk.jfr.Category;
+import jdk.jfr.Description;
+import jdk.jfr.Event;
+import jdk.jfr.Label;
+import jdk.jfr.Name;
+import jdk.jfr.StackTrace;
+
+/// JFR event emitted when row groups are selected by a byte-range row-group predicate
+/// (split-aware reading).
+///
+/// This is distinct from [RowGroupFilterEvent], which reports row groups dropped by
+/// statistics predicate push-down. Byte-range selection keeps a row group when its
+/// midpoint falls inside the requested split, and runs before push-down: the row groups
+/// reported as kept here are the input to the push-down evaluation.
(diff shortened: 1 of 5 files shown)


gunnarmorling (approved):
Good catch, merging. Thanks, @iifawzi!

PR #720, opened 2026-06-27
#105 Support predicate push-down via bloom filters   

### LINKED ISSUE #105: Support predicate push-down via bloom filters

iifawzi:
Keeping the momentum high, this PR resolves #105 wiring bloom filters for predicate push down, It's based on #719 for now, as the branch is in my repo, I couldn't use it as a base, but ideally, we should merge it after #719 lands. 

I was checking what parquet-java supports and it seems to support [float/double](https://github.com/apache/parquet-java/blob/0b1da61cc705dc4539a765b4a08a7d00d315fc97/parquet-column/src/main/java/org/apache/parquet/column/values/bloomfilter/BlockSplitBloomFilter.java#L379-L389), so i extended xxhash to have it supported as well, the rest is just using the foundation that has been introduced earlier 

## Checklist


_designs/BLOOM_FILTER_PUSHDOWN.md
```diff
@@ -0,0 +1,83 @@
...
+### `internal/predicate/RowGroupBloomFilterSource.java`
+
+The I/O-backed implementation over one `(InputFile, RowGroup)` pair. It reads a column's filter lazily — only when the evaluator probes it — and caches the result (including absence) for the lifetime of one row group's evaluation, so an `IN` list probing the same column reads the filter once. Evaluation of a row group is single-threaded (a sequential `stream().filter(...)`), so a plain `HashMap` cache suffices.
+
+Reading a filter:
+
+- `bloomFilterLength` present (the common case; written by current parquet-mr and PyArrow): one `readRange(offset, length)`, then `BloomFilterReader.read`.
+- `bloomFilterLength` absent (legacy writers): read a small fixed probe window to parse the header, derive the total length as `headerBytes + numBytes`, then `readRange(offset, total)` and `BloomFilterReader.read`.
+
+An `IOException` while reading a filter the footer declared is treated as corruption and surfaced as an `UncheckedIOException` carrying the file name (fail-early), consistent with `RowGroupIndexBuffers.fetch`.
```
iifawzi: Another suggestion by mr claude is to just log, given bloom filters pruning is nice-to-have, so we can decide to avoid throwing, I have no strong opinion honestely

gunnarmorling:
Noice, so cool to see! Will check it out asap. One thought: how about adding a test case for an equality filter benefitting from Bloom filters to the hardwood-benchmarks repo? It would be interesting to see how our implementation compares to parquet-java in that regard, and also what's the effect over Hardwood without Bloom filters. WDYT?

iifawzi:
> Noice, so cool to see! Will check it out asap. One thought: how about adding a test case for an equality filter benefitting from Bloom filters to the hardwood-benchmarks repo? It would be interesting to see how our implementation compares to parquet-java in that regard, and also what's the effect over Hardwood without Bloom filters. WDYT?

That would be very cool to have, it would even show us the limitations in our implementation if any, will work on it

gunnarmorling:
Merged #719; could you rebase to address the conflicts?

_designs/BLOOM_FILTER_PUSHDOWN.md
```diff
@@ -0,0 +1,83 @@
...
+
+Pruning is automatic whenever a filter predicate is supplied; there is no public API or CLI surface, and all new types live in `dev.hardwood.internal.*`.
+
+## Scope
+
+Bloom-filter pruning applies to a leaf predicate when **all** of the following hold:
+
+- The operator is `EQ`, or the predicate is an `IN` list.
+- The column's physical type is one of `INT32`, `INT64`, `FLOAT`, `DOUBLE`, `BYTE_ARRAY`, or `FIXED_LEN_BYTE_ARRAY` — the types for which the resolved predicate value maps unambiguously to the value's plain encoding, which is what the writer hashed. This corresponds to the resolved leaf types `IntPredicate` / `IntInPredicate` (→ `XxHash64.hash(int)`), `LongPredicate` / `LongInPredicate` (→ `XxHash64.hash(long)`), `FloatPredicate` (→ `XxHash64.hash(float)`), `DoublePredicate` (→ `XxHash64.hash(double)`), and `BinaryPredicate` / `BinaryInPredicate` (→ `XxHash64.hash(byte[])`). `FLOAT` / `DOUBLE` are `EQ`-only — there is no floating-point `IN` leaf.
```
gunnarmorling, line 18: I think the answer is yes, so just to double check: for logical predicate types, we'd convert to the physical value and then also apply Bloom filters to them, correct? For instance, are we converting the value of a `InstantColumnPredicate` to a long so the filter works?

_designs/BLOOM_FILTER_PUSHDOWN.md
gunnarmorling: I think we should raise the exception if the filter is corrupt. The general stance is fail-early, and if the filter isn't correct, chances are there other invalid blocks in the file to.

We may decide to add a kind of lenient mode at some point which would ignore corrupt yet optional part if it's possible. This would be a neat feature in particular for the TUI, allowing you to inspect a broken file as much as possible. But I'd handle that separately.

core/src/main/java/dev/hardwood/internal/predicate/BloomFilterSupport.java
```diff
@@ -0,0 +1,118 @@
...
+/// sufficient to drop the row group. Every check resolves the column's filter before hashing, so
+/// the statistics-only path (no [BloomFilterSource]) never pays for the probe hash.
+final class BloomFilterSupport {
+
+    private BloomFilterSupport() {
+    }
+
+    /// Whether the column's bloom filter proves the `INT32` `value` is absent. Returns `false`
+    /// (cannot prove absence) when no source is supplied or the column carries no filter.
+    static boolean absent(BloomFilterSource bloomFilters, int columnIndex, int value) {
```
gunnarmorling: maybe `valueAbsent()`? Feels a bit more descriptive.

core/src/main/java/dev/hardwood/internal/predicate/RowGroupBloomFilterSource.java
```diff
@@ -0,0 +1,119 @@
...
+    private BloomFilter readFilter(long offset, Integer length) throws IOException {
+        if (length != null) {
+            ByteBuffer buffer = inputFile.readRange(offset, length);
+            return BloomFilterReader.read(new ThriftCompactReader(buffer));
+        }
+        // Legacy writers omit bloom_filter_length. Over-read a fixed window — large enough to
+        // hold the header (a fixed-shape struct: an i32 plus three single-variant unions, ~19
+        // bytes at most), clamped so it never runs past EOF — and parse just the header to learn
+        // the region's total length.
+        int probe = (int) Math.min(fileLength() - offset, HEADER_PROBE_BYTES);
```
gunnarmorling: `intExact()`?

core/src/test/java/dev/hardwood/internal/predicate/BloomFilterPushDownTest.java
```diff
@@ -0,0 +1,234 @@
...
+
+/// Bloom-filter row-group pruning against `bloom_filter_test.parquet` (one row group, 64 rows;
+/// bloom filters on `id` INT64 `0..63`, `code` INT32 `0,3,…,189`, `name` STRING `""`…`"x"*63`,
+/// `price` FLOAT `0,2,…,126`, `ratio` DOUBLE `0,0.5,…,31.5`; `value` INT64 `0,10,…,630` carries
+/// none).
+///
+/// The discriminating cases use values that fall *inside* the column's statistics min/max range —
+/// so statistics alone keep the row group — but were never written, so only the bloom filter can
+/// prove their absence.
+class BloomFilterPushDownTest {
```
gunnarmorling, line 55: Do we somewhere have an end-to-end test, i.e. at the reader-level, too?

gunnarmorling (commented):
Woot, woot, really cool! A few comments inline. Plus, some minor nits via Claude:

## Documentation
- [x] `ROADMAP.md` ("…`eq` / `in` on INT32, INT64, and binary columns") and `FORMAT_COVERAGE.md` ("drives `eq`/`in` row-group pruning for INT32, INT64, and binary columns") both omit FLOAT/DOUBLE, yet the code prunes `eq` on `FloatPredicate` / `DoublePredicate` and the design doc + `query-controls.md` list them — add FLOAT/DOUBLE (eq-only) so the four doc surfaces agree on the supported set.
- [x] `_designs/BLOOM_FILTER_PUSHDOWN.md` "### Callers" claims *both* `RowGroupIterator.filterRowGroups` and `ParquetFileReader.filterRowGroups` construct a `RowGroupBloomFilterSource` and pass it to `canDropRowGroup`, but this PR removes all evaluator use from `ParquetFileReader.filterRowGroups` (now byte-range only; its `RowGroupFilterEvaluator` import is deleted) — bloom pruning is wired solely in `RowGroupIterator`, with the single-file path reaching it through the iterator. Reword so the doc doesn't send a reader hunting for a source construction that isn't there.

## Tests
- [x] The `absentAll(long[])` (INT64 `in`) and `absentAll(byte[][])` (binary `in`) paths are never exercised — only `in("code", …)` (INT32) is. INT64 `eq` bloom-drop is also untested because the `id` column is contiguous `0..63` (no in-range-absent probe exists). Add a binary `in` case and, if a long-bloom drop is wanted, a non-contiguous INT64 bloom column.
- [x] The design doc lists `FIXED_LEN_BYTE_ARRAY` as eligible, but the fixture has no FLBA column, so the FLBA `eq`/`in` path is asserted only by the shared `BinaryPredicate` code path, never with real bytes — consider an FLBA column in `bloom_filter_test.parquet` to cover the claim directly.

## Nits
- [x] `_designs/BLOOM_FILTER_PUSHDOWN.md` says the per-column cache is "a plain `HashMap`", but `RowGroupBloomFilterSource` actually uses parallel `BloomFilter[]` / `boolean[]` arrays indexed by column position (as its own class JavaDoc correctly states) — align the design-doc wording with the array implementation.

_designs/BLOOM_FILTER_PUSHDOWN.md
iifawzi: Yupp, because bloom filters are being applied on `ResolvedPredicate`, so the conversion is already taken care of in the resolver layer, if there's no test case covering that, will add it

core/src/test/java/dev/hardwood/internal/predicate/BloomFilterPushDownTest.java
iifawzi: We have one on the columnar reader; in this test class, I will add a couple more, and probably in a separate class

iifawzi:
Thanks for checking @gunnarmorling, changes have been applied, rebasing to main, doing the renamings of the methods, moving e2e tests to a separate class, and finally adding an additional oracle against java-parquet to ensure identical behavior when it comes to the hashing and filter parsing. I will work on the benchmark tomorrow to see how that behaves

gunnarmorling:
Hey @iifawzi, awesome work! Added two more commits for a few minor adjustments and logged #735 as a follow-up for reducing the number of GET requests for fetching Bloom filters from remote backends. This is ready to merge from my POV. Thanks a lot, really nice that we've checked this box now 🥳!

Issue #636, opened 2026-06-06
Row reader: reuse materialized Strings for dictionary-encoded columns instead of allocating per value


gunnarmorling:
## Context

From the #606 nested benchmark work: the row reader's record reconstruction is **materialization-bound**. On Overture Maps places (4.79 M rows, 628 MB), a full record scan is ~6.8 s, of which decode into batches is only ~1.05 s (~15 %) and per-value materialization on the consumer thread is ~5.8 s (~85 %) — overwhelmingly `new String(bytes, UTF-8)` allocations.

Overture's string columns (`names.primary`, `addresses[].*`, `sources[].*`, …) are `RLE_DICTIONARY`-encoded: a column chunk holds ~15–20 k distinct dictionary entries but millions of values referencing them. Today the row reader allocates a fresh `String` per value, so it re-materializes the same dictionary entries millions of times.

## Proposal

When a column chunk/page is dictionary-encoded, materialize each **dictionary entry's `String` once** (per chunk) and hand out the cached reference for every value referencing it, instead of `new String` per value. Transparent to the consumer — no API change.

This is distinct from #513: that issue *exposes* the dictionary block + per-row IDs to the consumer (a columnar API for engines like Flink to keep data dictionary-encoded through their pipeline). This issue is an internal allocation optimization on the existing materialized-`String` row path; consumers see ordinary `String`s, just fewer allocations.

## Expected impact

For a dict-encoded string column of cardinality C over N values, `String` allocations drop from N to ~C — on Overture's heavy text columns, roughly ~100× fewer. Since materialization (and its GC pressure) is the dominant cost and the limiter on parallel scaling, this should both shorten the single-thread scan and raise the parallel ceiling.

## Considerations

- Dictionary encoding is per-page; a column chunk may mix dictionary and plain pages — fall back to per-value materialization on plain pages and across the boundary.
- Lifetime/mutation: shared `String`s are immutable (safe), but the cache lifetime must be tied to the batch/chunk; existing flyweight contracts (values not safe to hold across `next()`) still apply to nested containers.
- Applies to other dict-encoded materialized leaves too (e.g. `byte[]`), though `String` is the dominant one.

Related: #513 (dictionary-encoded batches on the column-reader path — complementary, different layer).

gunnarmorling:
[...]

Issue #513, opened 2026-05-29
Expose dictionary-encoded batches on the read path


gunnarmorling:
On the read path, `ColumnReader` always materializes values into typed arrays, even when the underlying column chunk is dictionary-encoded. Engines with native dictionary column vectors cannot keep data dictionary-encoded through their pipeline.

Flink's vectorized reader has first-class dictionary support (`ParquetDictionary` + dictionary mode on its heap vectors) that saves CPU and memory on low-cardinality columns by deferring materialization. Because Hardwood eagerly decodes, that optimization is unavailable to a Hardwood-backed reader.

### Proposal
When a batch/page is dictionary-encoded, optionally expose:
- the dictionary block (the distinct values), and
- per-row dictionary IDs,

so a consumer can populate a dictionary-backed vector instead of materialized values. Consumers that don't opt in keep today's materialized arrays.

### Considerations
- Dictionary encoding is per-page; a column chunk may mix dictionary and plain pages. The API must express "this batch is dictionary-encoded with dictionary D" vs. "plain", or fall back to materialized values across a page/batch boundary.
- Distinct from #196 (dictionary-based filter pushdown) — this is about handing dictionary structure to the consumer, not evaluating predicates.

_Context: surfaced while integrating the column-reader API into Apache Flink's `flink-parquet`._

gunnarmorling:
## A concrete API shape

The obvious sketch is to let `getInts()` return dictionary ids and add `getIntDictionary()` / `getLongDictionary()` / `getStringDictionary()` alongside it. That overload is the one thing to avoid: for an `INT32` column both meanings return `int[]`, so a caller that forgets to check gets ids typed identically to values — plausible-looking, silently wrong numbers, exactly the failure mode the "fail early / no silent failures" rule targets. Ids should get their own name and `getInts()` should keep meaning "the values".

```java
// Build time — this changes what the decode workers do, so it can't be flipped per batch.
ColumnReaders readers = file.buildColumnReaders(projection)
        .withDictionaryOutput(DictionaryOutput.PREFERRED)   // NEVER | PREFERRED | REQUIRED
        .build();

while (readers.nextBatch()) {
    ColumnReader r = readers.get("city");
    if (r.isDictionaryEncoded()) {                 // per batch
        int[] ids = r.getDictionaryIds();          // length == getValueCount()
        String[] dict = r.getStringDictionary();   // length == getDictionarySize()
    }
    else {
[...]

PR #723, opened 2026-06-28
#636 Reuse materialized Strings for dictionary-encoded columns on the row reader

rionmonster:
## Description
This pull request addresses #636 by materializing each dictionary entry's `String` once per column chunk and handing that cached reference to every value referencing the entry, instead of calling `new String(bytes, UTF-8)` per value. On dictionary-encoded `UTF8` / `JSON` columns — the common case for low-cardinality text — the same entry was previously decoded into a fresh `String` for each of its (often millions of) references. The row reader's record reconstruction on nested, string-heavy data is materialization-bound rather than decode-bound, so that allocation dominated.

The change is transparent: callers still receive ordinary `String`s, with no public API change and no opt-in.

## Key Changes
  The end-state design is documented under `_designs/DICTIONARY_STRING_REUSE.md`; at a high level:

  - **Per-chunk interned `String` cache**
      - `Dictionary.ByteArrayDictionary` (record → final class) holds a lazily-allocated `String[]` parallel to its entry bytes; `internedString(index)` decodes each entry once and reuses it. One `ByteArrayDictionary` lives per column chunk, so the cache is chunk-scoped and released with the chunk. It is written and read only on the consumer thread (via `stringAt`), so it needs no synchronization — the decode threads only ever touch the immutable entry bytes.
  - **Carrying the dictionary entry index to the batch**
      - `RleBitPackingHybridDecoder.readDictionaryByteArrays` now also emits the per-value entry index (`-1` at null positions); `Page.ByteArrayPage` carries the source `ByteArrayDictionary` and the indices, both `null` for `PLAIN` / placeholder pages.
  - **A single per-batch dictionary slot in `BinaryBatchValues`**
      - The value holder carries one `dictionary` reference plus a lazily-allocated `int[] dictIndices`, gated to string columns via `internStrings`. `stringAt` resolves a dictionary-encoded value through the cache and falls back to the packed bytes otherwise. The array is allocated on the first dictionary page (`ensureDictionary`, which backfills the plain prefix with `-1`), so a plain string column — or any non-string column — allocates nothing.
      - A batch flushes at file boundaries, so it can straddle two chunks (two dictionaries). It keeps the first chunk's dictionary; the second chunk's values record `-1` and decode through the packed bytes — correct, merely un-deduplicated for that one straddling batch.
  - **Interning across every string accessor**
      - One `ValueConverter.isStringLeaf` predicate is the single source of truth for the string-leaf gate (the recording side and the consumer side both resolve through it). It routes the decoded-string paths — typed `getString`, generic `getValue`, `PqList.get` / `strings()` / `values()`, and `PqMap` generic/typed key + value — through the interned `getString` instead of `convertValue` / `new String`. The packed-byte representation is retained, so `getBinary` / raw-byte access keep their existing contract.

## Verification and Testing
Coverage exercises the change through real reads, not mocks; an independent review of the branch found no correctness blockers, and its findings were addressed in this PR:
  - `DictionaryStringReuseTest` — repeated values reuse one instance; cross-**batch** reuse over a 600k-row fixture; cross-**chunk** straddle over a two-row-group fixture (`dict_cross_chunk.parquet`) where one batch spans two dictionaries and chunk 1's values must resolve against their own dictionary; the generic / list accessors return the same interned instance; generic + typed `PqMap` key/value interning (`dict_string_map.parquet`); null dictionary values read back null; and `getBinary` still returns the raw bytes.
  - `ByteArrayDictionaryInternTest` — the per-chunk cache decodes each entry once; an empty-string entry is cached and reused; and the widened `decodePage` still decodes a non-string (`INT96` / `FLBA`) dictionary correctly.
  - Four committed fixtures (`dict_cross_batch`, `dict_cross_chunk`, `dict_nested_repeats`, `dict_string_map`) generated by `tools/simple-datagen.py`.
  - Full `core` suite green: `Tests run: 1454, Failures: 0, Errors: 0`.

## Performance and Throughput
Base-vs-branch sweep on a 14-core Apple Silicon host (allocation is deterministic; the flat throughput is the in-repo multifile test and very stable; the nested timing is single-shot JMH and indicative — a Linux host would tighten it). The numbers are the **#636 delta** and are orthogonal to the rebase onto current `main`:

  | Workload / metric (row reader, indexed) | main | Current | Δ |
  |---|---|---|---|
  | Overture nested scan — alloc/op | 52.4 GB | **36.7 GB** | **−29.9%** (−15.7 GB) |
  | Overture nested scan — GC pauses | 73 | 38 | **−48%** |
  | Overture nested scan — throughput | 0.92 M rows/s | 1.42 M rows/s | **~1.55×** _(indicative)_ |
  | NYC taxi flat scan — row reader / core | 20.16 M rec/s | 20.67 M rec/s | +2.5% _(no regression)_ |
  | NYC taxi flat scan — column reader / core _(control)_ | 39.67 M rec/s | 39.79 M rec/s | +0.3% _(no regression)_ |

The win lands on the materialization-bound nested string workload (Overture Maps places). The numeric flat scan is the regression floor: the row reader is flat-to-slightly-up, and the columnar path — untouched by this change — is unchanged, confirming the comparison is free of machine drift.

A follow-up (#724) tracks gating the dictionary-index recording off the column-reader path, which currently pays the bookkeeping without using it; correctness is unaffected and it's noted in the design doc's deferred section.

## Checklist
  - [x] If this PR adds or changes user-facing behavior, `docs/` has been updated _(no user-facing API change)_


rionmonster:
@gunnarmorling / @iifawzi 

This is the first pass at dictionary String reuse to start chipping away at the allocation waste detailed in #636. Overall — for dictionary-encoded UTF8/JSON columns, the row reader now materializes each entry's String once per column chunk and hands back the cached reference, dropping allocations from one object per value to one per entry. **That's effectively a ~95–100% cut in string-related allocation for those columns** depending on the payload:

image

Our real world nested Overture scan resulted in roughly **30% less total allocations (52 to 37 GB/op), roughly half the GC pauses, and a throughput bump on the nested path (~1.5x on my local MacBook Pro)**:

dict-string-reuse-636-allocation

I've also carved out a follow-up to add a similar impact to the column reader which I'm tracking in #724, which I'm hoping will be roughly as fruitful.

_designs/DICTIONARY_STRING_REUSE.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Scope
+
+- `UTF8` / `JSON` `BYTE_ARRAY` columns on the row-reader path (flat and nested), across every
+  string accessor — typed `getString`, generic `getValue` / `toString`, and the list and map
+  collection views all resolve through the per-chunk cache. The per-value `dictIndices` array is
+  allocated lazily and only for these columns; all other columns are unaffected.
+- The packed-byte representation is retained, so `getBinary` / raw-byte access keep their
+  existing contract.
+- Out of scope: non-string dictionary leaves, and the column reader's own `getStrings`
```
gunnarmorling, line 75: What's the rationale for not having the same fix for the column reader?

core/src/main/java/dev/hardwood/internal/reader/Dictionary.java
```diff
@@ -126,18 +127,57 @@ public Page decodePage(RleBitPackingHybridDecoder indexDecoder, int numValues,
...
         }
 
+        /// Returns dictionary entry `index` as a `String`, decoding it once per chunk
+        /// and caching it. Repeated values across the chunk share this one instance.
+        /// `index` must be a valid entry index: the row reader reaches this only via
+        /// [BinaryBatchValues#stringAt] for a non-null dictionary value, so a wiring
+        /// bug surfaces immediately as an out-of-bounds access here.
+        String internedString(int index) {
+            String[] cache = interned;
+            if (cache == null) {
```
gunnarmorling, line 159: Is there any benefit to creating the cache array lazily? Won't it be required in any case, so we're just slowing down each access?

core/src/main/java/dev/hardwood/internal/reader/BinaryBatchValues.java
```diff
@@ -42,9 +62,19 @@ public byte[] byteArrayAt(int idx) {
...
 
-    /// Materialise value `idx` as a UTF-8 decoded `String`. Allocates one
-    /// `String` per call.
+    /// Materialise value `idx` as a UTF-8 decoded `String`. A dictionary-encoded
+    /// value (when [#dictionary] is set and `dictIndices[idx]` is non-negative)
+    /// reuses the chunk dictionary's per-entry interned cache, so repeated values
+    /// are decoded once per chunk; any other value allocates one `String` per
+    /// call from [#bytes].
     public String stringAt(int idx) {
+        Dictionary.ByteArrayDictionary dict = dictionary;
```
gunnarmorling, line 71: Why do we copy here?

core/src/main/java/dev/hardwood/internal/reader/PqMapImpl.java
```diff
@@ -279,6 +279,9 @@ private Object valueAt(int valueIdx) {
         if (isValueNullAt(valueIdx)) {
             return null;
         }
+        if (ValueConverter.isStringLeaf(valueSchema)) {
```
gunnarmorling: There's a bit of duplication here, could we extract a helper? In Claude's words:

> The string-vs-convert branch this PR adds — `ValueConverter.isStringLeaf(schema) ? batch.getString(projCol, idx) : ValueConverter.convertValue(batch.getValue(projCol, idx), schema)` — is duplicated across `PqListImpl.getLeafValue` (plus the `LeafList` `internString` flag behind `values()` / `strings()`), `PqMapImpl.getValueAt` + `getKey`, `PqStructImpl` field decode, `NestedBatchDataView` leaf decode, and `FlatRowReader.getValue`. Consolidate into one `decodeLeaf(projCol, idx, schema)` on `NestedBatchIndex` (the four nested flyweights share it) plus a static sibling for the flat path, so the gate lives in one place.

gunnarmorling (commented):
Great work, @rionmonster! A few first comments inline. Gonna take another look later on, might have more feedback after that. Thanks a lot for driving this!

_designs/DICTIONARY_STRING_REUSE.md
gunnarmorling: If it is intentional to leave it out from this PR, could you log a follow-up issue for this? Note there is #513 which is related, but I think separate.

rionmonster: Yup, I created #724 for the column reader specifically just to avoid the scope of this PR growing too wildly. The plan is to carve it out after this one lands.

core/src/main/java/dev/hardwood/internal/reader/Dictionary.java
rionmonster: Not really on the speed front. 

The laziness is really so the non-interning dictionaries involved don't allocate a `String[]` that they'll never use/touch. Since we have a `ByteArrayDictionary` that backs all the dictionaries, but `internedString` here is _only_ called for UTF8/JSON, we don't want to allocate that for all of the ones that don't (e.g., FBLA decimals, INT96 timestamps, binary). It may be worth adjusting the comment to be more explicit. Something like "Lazily allocated so the X/Y/Z dictionaries that share this class never allocate a cache they won't use."

core/src/main/java/dev/hardwood/internal/reader/BinaryBatchValues.java
rionmonster: This really just functions as a fallback path for cases where we don’t have an interned String to reuse, so we need to copy the value before the underlying batch buffer gets recycled (via `new String(bytes, ...)`) to ensure the string doesn't change. In cases where we _do_ have a cached interned String, the dictionary path would just take care of this.

core/src/main/java/dev/hardwood/internal/reader/PqMapImpl.java
rionmonster: Yeah, this is a good point. I love a good helper, I'll carve this out. 👍

_designs/DICTIONARY_STRING_REUSE.md
gunnarmorling: That makes sense, thanks!

core/src/main/java/dev/hardwood/internal/reader/Dictionary.java
gunnarmorling: Gotcha.

core/src/main/java/dev/hardwood/internal/reader/BinaryBatchValues.java
gunnarmorling: Ah, sorry, what I meant is: why do we assign `dictionary` to that local `dict` var. But it doesn't really matter, moving on.

gunnarmorling:
This looks great! I've pushed one more commit which should speed up the generic `getValue()` method a bit: it precomputes the different access types (STRING, INT96_TIMESTAMP, RAW, CONVERT) at reader creation time and then dispatches over that type at runtime.

Issue #668, opened 2026-06-13
Centralize Thrift Compact Protocol type codes instead of hardcoding magic bytes


iifawzi:
### Problem

Thrift Compact Protocol field-type codes (`0x0C` struct, `0x06` i64, `0x07` double, `0x05` i32, `0x08` binary, `0x09` list, …) are hardcoded as magic-byte literals throughout the `dev.hardwood.internal.thrift` package. A canonical definition of these codes *already exists* in `ThriftCompactReader.java:20-31`:

```java
private static final byte TYPE_BOOLEAN_TRUE = 0x01;
private static final byte TYPE_I32         = 0x05;
private static final byte TYPE_I64         = 0x06;
private static final byte TYPE_DOUBLE      = 0x07;
private static final byte TYPE_STRUCT      = 0x0C;
// …
```

…but it is `private`, so every other reader re-hardcodes the raw byte instead, typically with an explanatory comment that exists only because the literal is opaque:

```java
// ColumnChunkReader.java
if (header.type() == 0x0C) { // STRUCT
if (header.type() == 0x06) { // (i64)
if (header.type() == 0x05) { // (i32)
```

### Evidence

- **~85 comparison sites** (`<x>.type() == 0x..`) across 17 reader files in `core/src/main/java/dev/hardwood/internal/thrift/` (`ColumnChunkReader`, `ColumnMetaDataReader`, `SchemaElementReader`, `LogicalTypeReader`, `PageHeaderReader`, `FileMetaDataReader`, `StatisticsReader`, `RowGroupReader`, `ColumnIndexReader`, `OffsetIndexReader`, … and the in-flight `BloomFilterHeaderReader`).
- **Duplicate definition:** `BoundingBoxReader.java:17` redeclares its own `private static final int TYPE_DOUBLE = 0x07;`.
- **Test code** adds ~23 more hardcoded type bytes in `ThriftCompactReaderTest`, `ColumnIndexReaderTest`, `BoundingBoxReaderTest`, `MalformedMetadataValidationTest`.

### Why this matters

- **DRY / single source of truth:** the codes are defined once but unreachable, so the byte is copy-pasted everywhere.
- **Readability:** `header.type() == TYPE_STRUCT` is self-documenting; `== 0x0C // STRUCT` needs a comment to stay readable.
- **Discoverability:** "where do we branch on i64?" is currently a hex grep rather than a reference search.
[...]

PR #733, opened 2026-06-30
#9 Flat INT32 writer: first increment of write support

gunnarmorling:
First increment of write support (#9), delivered as a narrow but complete vertical slice so `main` never holds non-functional writer code.

## What this adds
Writes a **flat schema of `REQUIRED INT32` columns** as a single uncompressed, PLAIN-encoded row group through a columnar API:

```java
FileSchema schema = FileSchema.builder("schema")
        .addColumn("id", PhysicalType.INT32, RepetitionType.REQUIRED)
        .build();
try (ParquetFileWriter w = ParquetFileWriter.create(OutputFile.of(path), schema)) {
    w.writeInts(0, new int[] {1, 2, 3});
}
```

This exercises the whole stack end to end: `OutputFile` (local atomic-rename + in-memory backends), `ThriftCompactWriter`, the `*Writer` metadata serializers (inverse of the existing `*Reader`s), `PlainEncoder`, the V1 data page, and the footer.

## Validation
- **DuckDB differential** (`WriterDifferentialTest`): hardwood writes, DuckDB reads via `read_parquet` and the values must agree — the inverse of `DifferentialReadTest`. Covers signed boundary values (`Integer.MIN/MAX`). An independent engine accepting the bytes proves spec-correctness, not just self-consistency.
- **Round-trip** (`WriterRoundTripTest`): write → read back with hardwood (in-memory, on-disk, empty column).
- All thrift field ids / enum codes cross-checked against upstream `parquet.thrift`.
- Full core suite green (1444 tests).

## Design
`_designs/WRITER_SUPPORT.md` is the plan of record: forward-only/footer-last write model, the row-group-buffering memory bound (the write-side inverse of the reader's mmap), the parallel-package layout, and the incremental delivery ladder. `ROADMAP.md` Phase 6 points to it.

## Scope / deferred
Flat-only, columnar-only, DataPage V1, single row group, uncompressed, PLAIN. Other primitive types, nullable columns, compression, dictionary, multiple row groups, `WriterConfig`, the row-oriented API, and nested (Dremel) shredding are later rungs on the ladder.

Draft pending review of the design and the increment-1 surface before continuing to increment 2.

### DIFF
--- a/core/src/main/java/dev/hardwood/internal/thrift/PageHeaderWriter.java
+++ b/core/src/main/java/dev/hardwood/internal/thrift/PageHeaderWriter.java
@@ -1,0 +1,78 @@
+/*
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Copyright The original authors
+ *
+ *  Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package dev.hardwood.internal.thrift;
+
+import dev.hardwood.metadata.Encoding;
+
+/// Writer for a DataPage V1 PageHeader to Thrift Compact Protocol, the inverse of
+/// the PageHeader/DataPageHeader parsing in [PageHeaderReader] and
+/// [DataPageHeaderReader].
+public class PageHeaderWriter {
+
+    /// Thrift `PageType.DATA_PAGE`.
+    private static final int PAGE_TYPE_DATA_PAGE = 0;
+
+    /// Serialize a DataPage V1 page header.
+    ///
+    /// @param writer the destination
+    /// @param numValues number of values in the page
+    /// @param uncompressedSize uncompressed size of the page body (levels + values)
+    /// @param compressedSize compressed size of the page body; equals the
+    ///        uncompressed size when the page is stored uncompressed
+    /// @param valuesEncoding the encoding of the values
+    public static void writeDataPageV1(ThriftCompactWriter writer, int numValues,
+                                       int uncompressedSize, int compressedSize, Encoding valuesEncoding) {
+        writer.pushFieldIdContext();
+
+        // 1: type
+        writer.writeFieldBegin(1, ThriftCompactConstants.FieldType.I32);
+        writer.writeI32(PAGE_TYPE_DATA_PAGE);
+
+        // 2: uncompressed_page_size
+        writer.writeFieldBegin(2, ThriftCompactConstants.FieldType.I32);
+        writer.writeI32(uncompressedSize);
+
+        // 3: compressed_page_size
+        writer.writeFieldBegin(3, ThriftCompactConstants.FieldType.I32);
+        writer.writeI32(compressedSize);
+
+        // 5: data_page_header
+        writer.writeFieldBegin(5, ThriftCompactConstants.FieldType.STRUCT);
+        writeDataPageHeader(writer, numValues, valuesEncoding);
+
+        writer.writeFieldStop();
+    }
+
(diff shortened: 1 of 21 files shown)


gunnarmorling:
Hey @iifawzi, @rionmonster, here's the plan and first step towards adding write support.

See the design doc for the details and in particular the delivery plan; the idea is to ship things strictly in increments which each provide value on their own. So instead of, for instance, implementing serializers for all types, which can't be used by themselves, the idea for the first increment here is to provide the most basic end-to-end support for writing int32 columns.

It's not a lot, but it already produces valid Parquet files. Next, we'll add more types, more encodings, parallelization, and so on. Would love your feedback, just in case you're interested and have some time.

// CC @Ugbot, as we just were talking about this too.

core/src/main/java/dev/hardwood/internal/thrift/ThriftCompactWriter.java
```diff
@@ -0,0 +1,123 @@
...
+
+    /// Field type for an i32 value (zigzag varint).
+    public static final byte TYPE_I32 = 0x05;
+    /// Field type for an i64 value (zigzag varint).
+    public static final byte TYPE_I64 = 0x06;
+    /// Field type for a length-prefixed binary/string value.
+    public static final byte TYPE_BINARY = 0x08;
+    /// Element/field type for a list.
+    public static final byte TYPE_LIST = 0x09;
+    /// Field type for a nested struct.
```
iifawzi: Probably it can be kept as is, and refactored in https://github.com/hardwood-hq/hardwood/issues/668, but i was thinking we need to centralize these fields across the project, do you think we can cross reference them from the reader for now? otherwise, it can follow up later as part of 668

core/src/main/java/dev/hardwood/internal/thrift/PageHeaderWriter.java
```diff
@@ -0,0 +1,78 @@
...
+
+        // 2: uncompressed_page_size
+        writer.writeFieldBegin(2, ThriftCompactWriter.TYPE_I32);
+        writer.writeI32(uncompressedSize);
+
+        // 3: compressed_page_size
+        writer.writeFieldBegin(3, ThriftCompactWriter.TYPE_I32);
+        writer.writeI32(compressedSize);
+
+        // 5: data_page_header
```
iifawzi, line 44: commenting just in case we can, but i understand you might prefer focusing on accuracy with required attributes, i'm thinking CRC can be included now as well: https://github.com/apache/parquet-format/blob/8a5e04bdecf100e8e981daacfa117e8b5aadacb9/src/main/thrift/parquet.thrift#L837

iifawzi (approved):
noice noice, we're starting! Looks pretty good to me, and I like the proposed dimenstion-first approach. I believe as it's evolving, several things would need to be shared between methods, but this can grow as it evolves, aside from that i left two little comments

core/src/main/java/dev/hardwood/internal/thrift/ThriftCompactWriter.java
gunnarmorling: Good point; reworked this a bit to lay the foundation for common constants here and using them in the new writer code. Leaving the adoption in the reader to #668.

core/src/main/java/dev/hardwood/internal/thrift/PageHeaderWriter.java
gunnarmorling: Yes, great idea! I'll do this as part of stage 2, as this will split arrays into pages.

PR #736, opened 2026-07-01
#9 write support stage 2: page chunking

gunnarmorling:


gunnarmorling:
Stage 2 of write support, splitting up the passed int32 arrays into 1MB chunks.

@iifawzi, this also contains the CRC32 check summing as discussed over in #733.

core/src/main/java/dev/hardwood/writer/ParquetFileWriter.java
```diff
@@ -130,6 +138,26 @@ public void writeInts(int columnIndex, int[] values) throws IOException {
...
+        // CRC-32 over the page body as stored on disk (uncompressed here), which is
+        // what the reader validates against.
+        CRC32 crc = new CRC32();
+        crc.update(valueBytes);
+        ThriftCompactWriter header = new ThriftCompactWriter();
+        PageHeaderWriter.writeDataPageV1(header, count, valueBytes.length, valueBytes.length,
+                (int) crc.getValue(), Encoding.PLAIN);
+        byte[] headerBytes = header.toByteArray();
+        out.write(ByteBuffer.wrap(headerBytes));
+        out.write(ByteBuffer.wrap(valueBytes));
```
iifawzi, line 157: Any expected overhead from such two write calls when this evolves in the future (to s3, etc..)? would there be any wins/overhead if we wrote in single call, or is negligible

iifawzi (approved):

core/src/main/java/dev/hardwood/writer/ParquetFileWriter.java
gunnarmorling: Good spot! I'd say the overhead is negligible today on local files; plus, it's already reduced to one write per column chunk in the stage-3 branch.

However, the real lever for S3/network is buffering in the OutputFile implementation, which we should make explicit; i.e. ChannelOutputFile should actually buffer as its docs claim, rather than coalescing writes in the encoder. I'll take care of that in stage-3 as well.

gunnarmorling:
Merged; thanks for reviewing, @iifawzi!

Issue #740, opened 2026-07-01
Fixed-size-list read fast-path: detect clean fixed-k LIST pages and skip record reconstruction


gunnarmorling:
## Motivation

Arrow `FixedSizeList<T, k>` / `FixedShapeTensor` (embeddings, image tensors, fixed-length vectors) map to Parquet's standard 3-level `LIST` encoding on write. For such columns every present row has exactly *k* children, so the per-element repetition-level stream is a constant, redundant restatement of a shape the schema already fixes. Decoding it element-by-element through the record-reconstruction state machine is pure overhead — the literature (`apache/arrow#34510` and others) reports a ~3× read-time gap vs. the same data laid out flat, at equal file size.

## Scope (PoC)

A **read-side** fast path — no Parquet-format or writer change. It recognises the clean fixed-*k* case from the level streams alone and bypasses list reconstruction, emitting the same output faster.

- Target shape: `optional group <name> (LIST) { repeated <primitive> element }` (maxRep=1, maxDef=2), dense/present vectors only (no null rows, no null elements).
- `DataPageV2` only. `DataPageV1` and legacy standalone `BIT_PACKED` levels fall back.
- Strict optimisation: identical public output, selected automatically, transparent fallback whenever the invariants don't hold.

## Phases

1. **Detector + def-gate** — pure predicate over a page's rep/def level regions returning `CleanFixedK(k)` or `NotApplicable`; O(1) def-gate + O(rows) rep run-detection that verifies (not trusts) *k* without reading vector interiors. Unit-tested against hand-built level byte streams.
2. **Fast-path decode** — on `CleanFixedK(k)`, emit the flat `(rows × k)` layout with implicit offsets, bypassing `NestedColumnWorker` assembly.
3. **Fallback wiring** — dispatch to the existing path on `NotApplicable`; correctness harness asserts bitwise-identical output across a corpus.

Design brief: `fixed-size-list-fastpath.md`.

Issue #732, opened 2026-06-30
Unified per-column-worker reader: direct addressing for non-repeated columns, materialization only for repeated columns


gunnarmorling:
## Motivation

Today the reader choice is made **per file**: `FileSchema.isFlatSchema()` routes a whole file to either `FlatRowReader` (direct, row-aligned columnar access) or `NestedRowReader` (record assembly / materialization). As a consequence, a single repeated column (list, map, or repeated field) drags *every* column in the file onto the materialization path — including the dozens of non-repeated primitives that gain nothing from it. Mixed schemas ("wide table of scalars + one or two list columns", a very common real-world shape) pay full assembly cost on their scalar columns for no reason.

The observation: **value-index addressing and value layout are per-column properties, not per-file properties.** Any column with `maxRepetitionLevel == 0` has exactly one value per row, so `valueIndex == rowIndex` within a row-aligned batch — no offsets, no repetition levels, no record assembly. Only repeated columns need the offset machinery.

Follow-up to #475 / #497 (which widen the flat path to non-repeated *structs* but still route the entire file to `NestedRowReader` the moment any repeated column is present).

## Proposed direction: push the flat/nested split down from the reader to the column worker

Instead of selecting one reader for the whole file, select the per-column **worker** by repetition level:

- `maxRepLevel == 0` → flat worker's dense, row-aligned (scatter) layout → `valueIndex = rowIndex`, no offsets, no rep levels.
- `maxRepLevel > 0` → nested worker's compact layout → `valueIndex = offsets[rowIndex]`, full assembly.

A single reader then sees a uniform addressing seam per column:

```java
int valueIndex = offsets == null ? rowIndex : offsets[rowIndex];
```

`ColumnWorker<B>` is already polymorphic with exactly these two subclasses (`FlatColumnWorker`, `NestedColumnWorker`); this is choosing among the two that exist per column, not introducing a new abstraction.

## The codebase is already most of the way there

- The "direct vs offset" seam is **already written**, just never exercised because offsets are always allocated:
  - `NestedBatchIndex.getValueIndex` (~:163): `return recordOffsets != null ? recordOffsets[recordIndex] : recordIndex;`
  - `NestedBatchDataView` per-row cache (~:102): `cachedValueIndex[col] = recordOffsets != null ? recordOffsets[rowIndex] : rowIndex;`
[...]

PR #741, opened 2026-07-02
Fixed-size-list read fast path (opt-in)

gunnarmorling:
Refs #740.

## Summary

A read-side fast path for Parquet columns that carry fixed-shape vector data
(embeddings, tensors, fixed-length arrays) stored under the standard 3-level
`LIST` encoding. For such a column every present row has exactly *k* children, so
the per-element repetition-level stream is a constant restatement of a shape the
schema already fixes. The fast path recognises this from the level streams alone
and skips the record-reconstruction state machine, emitting the same output with
implicit `(rows × k)` offsets.

**No format or writer change** — purely a reader optimisation over files existing
writers already produce. Opt-in for this release (off by default): enable it with
the `hardwood.fixed-list-fast-path` reader option. When enabled it is selected
automatically per page and falls back transparently whenever the invariants don't
hold, so its output is always identical to the regular path.

Design doc: [`_designs/FIXED_SIZE_LIST_FASTPATH.md`](_designs/FIXED_SIZE_LIST_FASTPATH.md).

## Approach

- **Detector** (`FixedSizeListDetector`) — an O(1) definition-level gate (a single
  RLE run of `maxDef` proves no nulls/empty lists) plus O(rows) repetition
  verification. `k` is *verified*, not trusted: variable inner lengths that
  merely sum to a multiple of the row count are rejected.
- **Two rep regimes** — large *k* reads the RLE interior length from its header
  (never scanning it); small *k* (fully bit-packed) uses a word-at-a-time
  (SWAR-style) tiled compare against the pattern's period.
- **Integration** — on a fixed-width page, `PageDecoder` decodes only the values and
  stamps the shape; `NestedColumnWorker` bulk-copies with arithmetic offsets and
  omits level arrays; `ColumnReader` builds the real-items view directly. Public
  output (columnar and row-reader) is identical to the regular path.
- **Both page versions** — `DataPageV2` (levels sliced pre-decompression) and
  `DataPageV1` (RLE-encoded levels, sliced post-decompression). Legacy standalone
  `BIT_PACKED` levels stay on the regular path.

## Results

JMH, average time, 2 forks × 5 iterations; 8M float32 leaf values per file,
`PLAIN`, uncompressed. `recovered = (naive − fast) / (naive − flat)`, where
*flat* is the same values as a plain float column (the decode floor).

| k | naive (ms) | fast (ms) | flat (ms) | speedup | recovered |
|---:|---:|---:|---:|---:|---:|
| 4 | 41.2 | 12.9 | 8.0 | 3.2× | 85% |
| 8 | 47.2 | 13.0 | 8.0 | 3.6× | 87% |
| 16 | 64.8 | 16.7 | 8.0 | 3.9× | 85% |
| 128 | 64.2 | 15.5 | 8.0 | 4.1× | 87% |
| 768 | 62.2 | 14.8 | 8.0 | 4.2× | 87% |
| 1536 | 61.8 | 14.8 | 8.0 | 4.2× | 87% |

- Recovers a stable **~85–87%** of the LIST-vs-flat overhead, at **3.2–4.2×**
  over the naive decode. `DataPageV1` performs within noise of V2.
- The residual over the flat floor is value decode + per-record assembly — the
  honest ceiling a native fixed-size-list type would still have to justify.
- The SWAR tiled compare cuts small-*k* detector cost ~17× (5.3 → 0.31 ms at
  k=4) but doesn't move the recovered fraction: detection overlaps value decode
  and was never the wall-clock bottleneck.

Reproduce:
```
python performance-testing/generate_fixed_size_list_data.py <dataDir>
java -jar performance-testing/micro-benchmarks/target/benchmarks.jar \
     FixedSizeListDecodeBenchmark -p dataDir=<dataDir>
```

## Testing

- `FixedSizeListDetectorTest` (20) — both regimes, multi-byte periods, and the
  §8 edges (padding, encoder-dependent split, adversarial variable lengths).
- `FixedSizeListFastPathReadTest` (11) — fast-vs-reconstruction equality (via the
  `hardwood.fixed-list-fast-path` switch) across V1/V2 and the *k* regimes, through
  both the columnar and row-reader APIs, cross-checked against the generator.
- `DifferentialFixedSizeListTest` — fast-path output validated against a DuckDB
  oracle: plain/dict/compressed, multi-page/multi-row-group, a row-filtered read,
  and a fixed-width + null mixed column (the batch-homogeneity conversion).
- Full core suite green (1586).

## Scope / follow-ups

- In: dense present vectors, primitive elements, `maxRep==1`/`maxDef==2`.
- Follow-ups (noted in the design doc + roadmap): nullable vectors (null rows);
  the RLE-interior high-row-count regime (k≈16) still leaves ~20% detector cost
  in an O(rows) header walk.

Shipped as a real but opt-in feature (off by default) while it matures in the
field; enable it with `hardwood.fixed-list-fast-path=true`. As a bonus it
quantifies how much of the LIST read penalty is recoverable in the reader alone —
a useful data point for the `parquet-format` fixed-size-list discussion.


gunnarmorling:
Hey @iifawzi, so this is an interesting one; I got nerd-sniped by the current discussion on parquet-dev about adding a fixed-sized list type (e.g. for storing vector embeddings). Parquet's current variable length encoding is a poor fit for that, as you need to store repetition and definition levels for each vector element, whereas all you need is a single value, the fixed list length. In the absence of a dedicated type for that, I iterated with Claude on how such lists could be read faster as-is.

The idea is to avoid the materialization of the def/rep level arrays if we recognize a fixed length list. For def levels it's trivial, and we already do so as per my change from earlier this week: we just need to look for a single RLE run of the max def level. For rep levels, it's a bit more complicated. Essentially, we're looking for `0, 1, 1, …, 1 (one 0, then k−1 1s)` repeated. The encoding differs though depending on the value of k: bit-packed for k < 8, RLE-encoded otherwise.

So we have a detector which matches the def levels and walks the rep levels and if we recognize it's fixed-length list, we never materialize either array. This gives quite a nice boost, whereas the price to pay is really low: the detector only engages for lists and the rep level walk is really cheap compared to the actual page decoding. I.e. I could measure any significant regression for the regular case. It's still gated behind an opt-in flag for the time being.

Would love your take, if you have some time. Might also write a quick blog post about it, might make for an interesting topic. Avoiding the def level materialization is common practice, but the rep level pattern matching is not, AFAICS. A proper fixed-width type in Parquet would still be advantageous, but this should go to show that the gap can be closed quite a bit even with the current encoding.

gunnarmorling:
// CC @rionmonster

iifawzi:
Boom Boom Boom, that's a very nice one!! Taking a look as we talk, I'm reviewing this one without claude, to enjoy/learn it. Impressive speedups!

core/src/test/java/dev/hardwood/internal/reader/FixedSizeListDetectorTest.java
```diff
@@ -0,0 +1,274 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import dev.hardwood.internal.reader.FixedSizeListShape.FixedWidth;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/// Tests for [FixedSizeListDetector], the pure level-stream predicate behind the
+/// fixed-size-list read fast path. Level streams are hand-constructed run by run
+/// so both encoding regimes — fully bit-packed (small `k`) and RLE interior
```
iifawzi: I'm not sure if `k` is easily explained here (i needed to read first test to recognize that this represents the list length). Could be a tiny detail anyway, and perhaps because this's the first file i started with

core/src/main/java/dev/hardwood/internal/reader/FixedSizeListDetector.java
```diff
@@ -0,0 +1,437 @@
...
+            int groups = (int) (header >> 1);
+            if (groups > end - pos) {
+                return FixedSizeListShape.NOT_APPLICABLE;
+            }
+            int payloadPos = pos;
+
+            if (k == 0) {
+                if (bitAt(data, payloadPos, 0) != 0) {
+                    return FixedSizeListShape.NOT_APPLICABLE; // must open on a boundary
+                }
```
iifawzi, line 279: I understand it's just a guard, no? as it's not possible for a legitimate bit-packed stream with k=0 (the entire stream is expected to be 0), so i'm thinking (saving one bitAt call) that we can/should fail early earlier ` if (groups > end - pos || groups <= 0) {

rionmonster:
Wow this is really, really impressive! Great find @gunnarmorling, looking forward to taking some time to better grok the internals. 🚀

core/src/main/java/dev/hardwood/internal/reader/FixedSizeListDetector.java
```diff
@@ -0,0 +1,437 @@
...
+                        break;
+                    }
+                }
+                if (k == 0) {
+                    return null; // no second boundary: large k, defer to the scalar walk
+                }
+                period = k / gcd(k, 8);
+                if (period == 1) {
+                    refByte = patternByte(k, 0);
+                    refWord = (refByte & 0xFFL) * 0x0101010101010101L;
```
iifawzi: very nit: `0x0101010101010101L` if we can move it to constant

core/src/main/java/dev/hardwood/internal/reader/FixedSizeListDetector.java
```diff
@@ -0,0 +1,437 @@
...
+                    refWords = new long[period];
+                    for (int p = 0; p < period; p++) {
+                        long word = 0;
+                        for (int i = 0; i < 8; i++) {
+                            word |= (refPat[(p + i) % period] & 0xFFL) << (8 * i);
+                        }
+                        refWords[p] = word;
+                    }
+                }
+            }
```
iifawzi: Would a lookup table for k <= 8 values for refPat and refWord be more performant here? i'm thinking this can show some perf wins, we can calculaute the values for the 8 cases and have it as a constant table

iifawzi (commented):
I've taken a look on the detector, left few comments, will check the rest tomorrow. very interesting one, and definitely worht a write up

Feel free to merge as well, I will be taking a look on the rest tomorrow's night (friday, i just need to sleep :d) , and will leave any thoughts here anyway

core/src/test/java/dev/hardwood/internal/reader/FixedSizeListDetectorTest.java
iifawzi: probably this's a nit that i will hold, aftering goint through the implementation it's totally clear

gunnarmorling:
Iterated some more on this; also your comments should be addressed now, @iifawzi. Thanks a lot for checking this out. Will keep it open some more and merge later today or tomorrow.

iifawzi (approved):

gunnarmorling:
Squashed into one commit, applying now. Thanks a lot for reviewing, @iifawzi!

Quick heads-up that before I'm gonna publish the write-up I mentioned, I'm planning to land a few improvements to the regular nested path (#750, #751, maybe #732), as it has been largely unoptimized so far. I want to avoid reporting inflated wins by the fast path over a poor baseline. I'm still confident that the fast path is advantageous for the fixed-list case.

PR #747, opened 2026-07-03
#578 Extend DifferentialValueTest to cover OPTIONAL columns (null-decode path)

### LINKED ISSUE #578: Extend DifferentialValueTest to cover OPTIONAL columns (null-decode path)
Workstream of #548 — follow-up to #574.

`DifferentialValueTest` (P2) covers every physical and logical type the corpus carries, but only on `REQUIRED` columns — the schema for `diff_types.parquet` has no `OPTIONAL` column. The deliberate scope cut (per the P2 commit message: "so the values themselves — not null handling — are what's compared") was right for the initial harness, but it leaves the **null-decode path uncovered end-to-end**:

- `DifferentialNullTest` exercises the filter side: it pins that `WHERE val <op> n` drops nulls, but never reads back a row whose value is null.
- `DifferentialValueTest` exercises the decode side per type, but never on a column that can actually carry a null, so `rows.isNull(col)` / sentinel-handling is never asserted against the oracle.

A wrong null-decode (e.g. a definition-level off-by-one, a sentinel collision in dictionary-encoded pages) would slip past both classes today.

### Acceptance criteria
- The differential corpus gains either (a) a parallel `diff_types_optional.parquet` whose schema mirrors `diff_types.parquet` but with every column declared `OPTIONAL` and ~⅓ of rows null per column, or (b) selected `OPTIONAL` columns added to the existing `diff_types.parquet`.
- `DifferentialValueTest` reads through the new column(s) and asserts that for each row, `rows.isNull(col)` matches `rs.wasNull()` (or the DuckDB-side null check); for non-null rows, the value comparison runs as it does for the `REQUIRED` columns today.
- Covers at least one column per physical type already in the harness (`int32`/`int64`/`float`/`double`/`boolean`/`string`/`binary`) and one logical type (e.g. `decimal` or `timestamp`) so the null path is exercised across the decoder family, not only one slot.

### Out of scope
- Repeated / nested fields. Three-level repetition has its own null-encoding behaviour; that belongs with the broader nested-types extension flagged in the design's P3 phase.
- Filter null-semantics — already covered by `DifferentialNullTest`.

OmarAbdelmawgoud5:
fixes #578

DifferentialValueTest exercised every physical/logical type but only on REQUIRED columns, so definition-level decoding and dictionary null-sentinel handling were never checked against the oracle. A wrong null decode (e.g. an off-by-one in definition levels, or a sentinel collision in dictionary-encoded pages) could pass both DifferentialValueTest and DifferentialNullTest, since those tests never see a nullable column and the latter only checks that filters drop nulls, not that a null value decodes correctly.

## Checklist

- [x] The commit message is prefixed with the issue key ("#578")


tools/simple-datagen.py
```diff
@@ -4171,6 +4171,54 @@ def _build_dive_screenshots_chunk(start, count):
...
+    'd': [_null_every_third(r, date.fromordinal(_date_base + r)) for r in range(DIFF_TYPES_OPTIONAL_ROWS)],
+    'ts': [_null_every_third(r, datetime.fromtimestamp(1577836800 + r, tz=timezone.utc)
+                              .replace(microsecond=(r * 137) % 1000000))
+           for r in range(DIFF_TYPES_OPTIONAL_ROWS)],
+    'dec': [_null_every_third(r, Decimal(f'{r}.25')) for r in range(DIFF_TYPES_OPTIONAL_ROWS)],
+    'bin': [_null_every_third(r, bytes([r % 256, (r + 1) % 256, (r + 2) % 256]))
+            for r in range(DIFF_TYPES_OPTIONAL_ROWS)],
+}, schema=diff_types_optional_schema)
+pq.write_table(
+    diff_types_table,
```
iifawzi, line 4215: I believe you wanted to use `diff_types_optional_table`. that being said, tests would fail after the change, so the test would need to be modified as well to take care of the nullability

iifawzi (commented):
Hello @OmarAbdelmawgoud5 thanks for working on it! I left a comment on the generator if you can check it. 

On the other hand, it would be nice if you can rever tthe rest of .parquet files changes (The encrypted ones gets updated because of tht key change in each run, but data remains the same, and probably gzio_compressed has something similar that differ based on environment, so it can be reverted as well)

Issue #714, opened 2026-06-26
Variant primitive & short-string length not bounds-checked before allocation/access


gunnarmorling:
## Bug

The Variant value decoder trusts the declared length of primitive string/binary and short-string payloads without bounding it against the buffer:

- `PqVariantImpl.value()` does `new byte[VariantValueDecoder.valueLength(...)]`, where `valueLength` for a long string/binary is `4 + readIntLE(declaredLen)` with no check. A crafted `declaredLen = 0x40000000` would attempt a ~1 GiB allocation (the README's "most critical" allocation-DoS concern); `0x7FFFFFFF` overflows to a negative size.
- Short-string and truncated size fields fail with raw `ArrayIndexOutOfBoundsException` rather than a validated, contextual error.

## How it surfaced

Three fixtures from [apache/parquet-testing#113](https://github.com/apache/parquet-testing/pull/113), each rejected only incidentally today:

| Fixture | parquet-java expects | Hardwood actually throws |
|---|---|---|
| `oversized_primitive_size` | `IAE` …"extends past buffer" | `NegativeArraySizeException: -2147483644` |
| `short_string_length_exceeds_buffer` | `IAE` …"short string" | `ArrayIndexOutOfBoundsException` (arraycopy) |
| `truncated_primitive_size` | `IAE` …"truncated" | `ArrayIndexOutOfBoundsException` (reading size field) |

## Fix

- Before allocating/copying, validate that the declared primitive/short-string length (and the 4-byte size field itself) fits within the value buffer; throw `IllegalArgumentException` with a descriptive message otherwise.
- Add unit tests for all three layouts asserting the contextual rejection.

## Related

- Hardening test harness: #711.

PR #758, opened 2026-07-05
#713 Guard Variant object/array/metadata offset arithmetic against 32-bit overflow

rionmonster:
## Description
This pull request addresses #713, which observed that the Variant decoder sizes its object / array / metadata header tables with unguarded 32-bit arithmetic. `VariantValueDecoder.parseObject` / `parseArray` and the `VariantMetadata` constructor compute the id-array and offset-table extents from a file-controlled element / dictionary count (`numElements * idSize`, `(numElements + 1) * offsetSize`, `(dictionarySize + 1) * offsetSize`) in plain `int`. A hostile count such as `0x33333333` overflows the product and wraps the values-section offset to a small or negative value that looks in-bounds, so the subsequent per-access bounds checks read from an arbitrary buffer location instead of rejecting the input. Today Hardwood only avoids this by accident (an unrelated field-id-range check happens to trip first) so the vulnerability the upstream `int_overflow_in_bounds_check` fixture targets is genuinely unguarded.

The change is internal to the variant decoder: the public `PqVariant` surface is unchanged, and well-formed files decode exactly as before.

## Key Changes
- **Widened the header arithmetic to `long` over the unsigned count**
    - `parseObject` / `parseArray` compute the id-array and offset-table extents in `long` via `Integer.toUnsignedLong(numElements)`, and `VariantMetadata` widens `(dictionary_size + 1) * offset_size` the same way, so a hostile count can no longer wrap the product into an in-bounds-looking offset.
- **Consolidated the fit check into one shared helper**
    - A single package-private `VariantBinary.checkHeaderFits(region, count, headerEnd, bufferLength)` validates the `long` extent against the buffer length, returns it narrowed via `Math.toIntExact`, and throws a descriptive `IllegalArgumentException` naming the count and the buffer size when the header cannot fit. Both the decoder and the metadata constructor route through it, replacing the ad-hoc truncation check.
    - Downstream per-element arithmetic (`i * offsetSize`, `numElements * offsetSize`, …) is deliberately left as `int`: it only ever runs on a layout that has already cleared the fit check, where the product is provably bounded by the buffer length and cannot overflow.
    - The pre-existing negative-`dictionary_size` guard in `VariantMetadata` is retained ahead of the helper, so that path keeps its existing message.

## Verification and Testing
Coverage drives the real decode entry points with crafted headers rather than mocks, and each test was confirmed to fail before the guard was added:
- `VariantValueDecoderTest` - `objectElementCountOverflowRejected` / `arrayElementCountOverflowRejected` feed a `0x33333333` count through `parseObject` / `parseArray` and assert an `IllegalArgumentException` naming the count and buffer size. Pre-fix, both methods returned a wrapped, in-bounds-looking layout with no exception at all. `largeButValidArrayStillDecodes` builds a genuine 300-element array via `VariantValueEncoder.writeArray` and asserts it still decodes, so the guard rejects only counts too large to fit — not the `is_large` 4-byte-count encoding itself.
- `VariantMetadataTest` - `dictionarySizeOffsetOverflowRejected` drives the same overflow through the metadata constructor; pre-fix it threw an incidental `ArrayIndexOutOfBoundsException` from a wrapped negative offset reaching into the buffer, exactly the bounds-check bypass the issue describes.
- Full `core` module green via `./mvnw -pl core -am verify`: `Tests run: 1551` (unit) + `14` (integration), `Failures: 0, Errors: 0`, including the Error Prone `NoVar` / `NoLegacyJavadoc` gates.

The guarded methods run once per object / array / metadata parse (not per value), and on the success path the fit check is a single comparison plus `Math.toIntExact` (the message is constructed only on the throw path) so there is no measurable decode-path cost.

## Checklist

### DIFF
--- a/core/src/main/java/dev/hardwood/internal/variant/VariantBinary.java
+++ b/core/src/main/java/dev/hardwood/internal/variant/VariantBinary.java
@@ -107,6 +107,30 @@ public final class VariantBinary {
         return result;
     }
 
+    /// Validates that a Variant object/array/metadata header fits inside a buffer
+    /// of `bufferLength` bytes. `headerEnd` is the exclusive end offset of the
+    /// header's id and offset tables — the caller computes it in `long` from the
+    /// **unsigned** element or dictionary `count`, so a hostile count cannot wrap
+    /// the 32-bit product into an in-bounds-looking offset that later slips past
+    /// the per-access bounds checks. Returns `headerEnd` narrowed to `int` once it
+    /// is known to fit.
+    ///
+    /// @param region human-readable name of the count, used in the error message
+    /// @param count declared element/dictionary count (interpreted as unsigned 32-bit)
+    /// @param headerEnd exclusive end offset of the id/offset tables, computed in `long`
+    /// @param bufferLength length of the enclosing buffer
+    /// @return `headerEnd` as an `int`
+    /// @throws IllegalArgumentException if the header does not fit the buffer
+    static int checkHeaderFits(String region, int count, long headerEnd, int bufferLength) {
+        if (headerEnd > bufferLength) {
+            throw new IllegalArgumentException(
+                    "Variant " + region + " count " + Integer.toUnsignedString(count)
+                            + " does not fit in a buffer of " + bufferLength
+                            + " bytes (header tables alone require " + headerEnd + " bytes)");
+        }
+        return Math.toIntExact(headerEnd);
+    }
+
     /// Map a Variant value-header byte to its [VariantType]. Returns `null` for
     /// unrecognized primitive type tags so callers can fail-early with context.
     public static VariantType typeOf(byte header) {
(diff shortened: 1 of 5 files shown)


rionmonster:
@gunnarmorling 

Here's another one to tackle one of our open lingering `1.0.1.Final` bugs. I'm hoping to carve out some time to begin working through these as we move towards the next release. 👍

gunnarmorling (approved):

gunnarmorling:
Merged; thank you so much, @rionmonster! This one, and a few others, fell out of a conversation on parquet-dev after the 1.0 announcement, where I got pointed to a corpus of malformed VARIANT files which we should reject gracefully. The entire list is tracked in #711, in case you'd want to tackle others like this one.

rionmonster:
@gunnarmorling 

Sounds good! I just kicked off going after #714 since it seemed in the same general vein. I’ll take a look at #711 for some other candidates as well. 👍

Issue #724, opened 2026-06-28
Reuse interned dictionary Strings in the column reader


rionmonster:
## Description
Follow-up from the #636 review.

#636 makes the row reader materialize each dictionary entry's `String` once per column chunk and reuse it (`BinaryBatchValues.stringAt`). `ColumnReader.getStrings()` still re-decodes every value with `new String(bytes, …)`, even for dictionary-encoded UTF8/JSON columns. The column reader **already records** the dictionary and per-value entry indices onto each batch's `BinaryBatchValues` (recording is on for string columns) — `getStrings()` just discards them. So this is mostly about consuming work already done, not adding machinery.

## Proposed Primary Solution
Make `getStrings()` resolve through the per-chunk interned cache on every column-reader path (flat/nested, unfiltered/filtered), so repeated dictionary values return one shared `String` instance, matching the row reader. Two of the four paths drop the dictionary during batch compaction (`ColumnReader.compactBinary`, used by the filtered-flat and nested-with-repeated reads), so this also threads the per-value entry indices through compaction — gathered through the same index already used to gather the bytes. Values are otherwise unchanged; `getBinary*` and raw-byte access keep their contract.

## Alternate Solution (Minimal Approach)
If threading through compaction proves untenable, a reduced version covers only the non-compacted paths (unfiltered-flat and nested-struct): `getStrings()` reuses the cache where the dictionary survives, and the compacted paths keep today's per-value decode (correct, just un-deduplicated). Contained to `ColumnReader`, with no compaction changes.

Issue #748, opened 2026-07-03
Batch sizing ignores list fan-out: high-k repeated columns OOME (whole file drains into one batch)


gunnarmorling:
## Summary

`BatchSizing.computeOptimalBatchSize` sizes a batch by assuming **one value per row per column**. For a repeated column whose lists hold `k` elements per row, the nested worker accumulates `batchCapacity × k` *values* per batch, so the real per-batch footprint is ~`k`× the byte budget the sizing model targets. At large `k` this OOMEs.

## Reproduction

Fixed-size-list benchmark (`_tmp/hardwood-benchmarks`), 1 GB files, `-Xmx4g`:

```
./run-fixedlist.sh --k 768 --total-values 256000000 --warmup 3 --meas 3 --forks 1
```

```
Caused by: java.lang.OutOfMemoryError: Java heap space
    at java.util.Arrays.copyOf(Arrays.java:3673)
    at dev.hardwood.internal.reader.NestedColumnWorker.trimValues(NestedColumnWorker.java:500)
    at dev.hardwood.internal.reader.NestedColumnWorker.publishCurrentBatch(NestedColumnWorker.java:302)
    at dev.hardwood.internal.reader.ColumnWorker.finishDrain(ColumnWorker.java:399)
    ...
```

## Root cause

For a single `FLOAT` column, `computeOptimalBatchSize` returns `6 MB / 4 B = 1.5M`, clamped to `MAX_BATCH = 524288` **rows**. With `k = 768`:

- Total rows = `256M / 768 ≈ 333,333` — fewer than the 524,288-row cap, so the **entire column drains into a single batch**.
- The `NestedColumnWorker` accumulators grow (by doubling) to hold all 256M values:
  - `nestedValues` `float[~268M]` ≈ 1.0 GB
  - `nestedDefLevels` `int[~268M]` ≈ 1.0 GB
  - `nestedRepLevels` `int[~268M]` ≈ 1.0 GB
- `publishCurrentBatch` → `trimValues` then does `Arrays.copyOf(nestedValues, 256M)` for another ~1.0 GB → **~4.2 GB peak > `-Xmx4g`** → OOME.

This is **not** fast-path specific — the regular nested assembly path OOMEs identically at this fan-out. The fixed-size-list benchmark just exposes it because `k` is large and the file fits in one row-batch.

## Proposed fix

[...]

PR #759, opened 2026-07-05
#724 Reuse interned dictionary Strings in the column reader

rionmonster:
## Description
This pull request addresses #724, a follow-up to #636, which interns each dictionary entry's `String` once per column chunk on the **row reader**. The **column reader** already records the same per-chunk dictionary and per-value entry indices onto each batch (recording is gated by column type, not by consumer), but `ColumnReader.getStrings()`
discarded them and re-decoded every value with `new String(bytes, UTF-8)` paying the bookkeeping cost with none of the dedup benefit. This change makes the column readerconsume that work, so a dictionary-encoded `UTF8` / `JSON` column reuses one interned `String` per entry per chunk, matching the row reader.

The change is transparent: callers still receive ordinary, equal-content `String`s, with no public API change and no opt-in.

## Key Changes
The end-state design is documented under `_designs/DICTIONARY_STRING_REUSE.md`; at a high level:

- **`getStrings()` resolves through the per-chunk cache.** `ColumnReader.getStrings()` delegates to `BinaryBatchValues.stringAt`, which reuses the chunk dictionary's interned instance for a dictionary-encoded value and falls back to the packed bytes otherwise; a null position is guarded before resolution. A shared `realLeafBinary()` helper fetches the leaf value holder for both `getStrings()` and `getBinaries()`.
- **Interning survives batch compaction.** The two compacting paths (filtered-flat and repeated list / map reads) go through `compactBinary`, which now gathers each kept value's entry index alongside its bytes and carries the chunk dictionary onto the compacted batch. So all four column-reader paths intern: unfiltered-flat and struct-only-nested keep their dictionary directly; filtered-flat and repeated carry it through compaction. The gather is guarded to dictionary-encoded string columns (`dictionary != null`), so other columns are unaffected.
- **The packed-byte representation is retained**, so `getBinary` / raw-byte access keep their existing contract.

## Verification and Testing
Coverage exercises the change through real reads and related tests were added to verify execution paths as well as update related docs to reflect interning:
- `DictionaryStringReuseTest` - new column-reader cases across all four paths: flat unfiltered interning identity; struct-only nested interning + null read-back; list (repeated, compacted) interning; filtered-flat (compacted) interning; and cross-**batch** reuse over the 600k-row single-chunk fixture (one instance per value across many batches). The existing row-reader coverage is unchanged.
- User-facing `docs/content/how-to/column-reader.md` updated so the `getStrings()` allocation note reflects dictionary interning.
- Full `core` suite green: `Tests run: 1552, Failures: 0, Errors: 0`.

### Performance
`DictionaryStringReadBenchmark` (JMH, `-prof gc`) scans a low-cardinality dictionary-encoded string column (`dict_strings.parquet` — 8 distinct values over 2M rows) via `ColumnReader.getStrings()`, branch vs `main` on the same fixture. Allocation is deterministic (±<0.01% across iterations); timing is single-host indicative.

hardwood-724-perf

The same values can be reflected in the table below:

| Workload / metric (column reader, `getStrings()`) | main | Current | Δ |
|---|---|---|---|
| Dictionary string scan — alloc/op (`gc.alloc.rate.norm`) | 230.2 MB | **140.3 MB** | **−39%** (−89.9 MB) |
| Dictionary string scan — throughput | 81.5 M rows/s | **160.6 M rows/s** | **~1.97×** |
| Dictionary string scan — avg time | 24.53 ms/op | **12.46 ms/op** | **−49%** |

The ~90 MB/op saved over 2M rows is exactly the per-value `new String` (object + backing array) eliminated for repeated dictionary entries (8 distinct values interned once per chunk instead of ~2M fresh decodes; no regression from the added index gather or the in-loop branch. The benchmark and its fixture generator (`generate_dict_string_data.py`) are included so the numbers reproduce.

## Checklist


rionmonster:
@gunnarmorling 

Here's the related follow-up to #636 for the Column Reader changes to reduce allocations which also seemed to net a pretty nice ~40% reduction in allocations and roughly 2x bump in throughput (as far as dictionary scans are concerned). Feel free to take a look when you get a chance.

gunnarmorling:
Woot,  woot, awesome, gonna check it out tomorrow. Out of curiosity, do you see it make a difference to `NestedPerformanceTest`? It is string heavy, so it could be the case, but then I'm not sure whether it uses dictionary encoding.

rionmonster:
> Woot, woot, awesome, gonna check it out tomorrow. Out of curiosity, do you see it make a difference to `NestedPerformanceTest`? It is string heavy, so it could be the case, but then I'm not sure whether it uses dictionary encoding.

Hey hey! So I dug into this a bit this evening and the short answer is that `NestedPerformanceTest` doesn't really see a difference (~3% fewer allocations, throughput generally flat). While it _is_ very string-heavy, the columns it actually materializes as strings are PLAIN-encoded, so there's no dictionary to intern against (and most of them go through the row reader and not the path this PR touches).

With that being said, I had Claude help me point some reads at a column in that same file that _is_ dictionary-encoded and A/B it against the current main. This lead to a slightly interesting bit in that **the interning win turns out to be tightly tied to batch-vs-row-group alignment**:

| batch size | main | current |  Δ allocations |
| -- | -- | -- | -- | 
| auto (393k rows, the default) | 2497 MB/op | 2420 MB/op | -3.1%  | 
| 16k rows (fits inside a row group) | 2436 MB/op | 1705 MB/op | -30.0%  |

The primary reason here is that **our auto batch size (~393k rows) is way bigger than this file's row groups (~108k rows), so every batch spans 3–4 of them**. Right now we only hang onto the first chunk's dictionary per batch, so everything past that first chunk falls back to a fresh String. Shrink the batch so it fits inside a row group and interning fires on basically everything. So it does help nested dictionary string reads ...  just not at the default batch size on a file with lots of small row groups.

That makes me think there _may_ be an optimization here to chase, and Claude seemed to have two ideas to tackle it:
- **On the batch side, stop batches from straddling row groups.** Either cap the batch at row-group boundaries for these string columns, or fix our batch sizing so repeated columns get smaller batches (e.g. `computeOptimalBatchSize` counts records, but each record fans out to multiple leaf values, so we overshoot (this is why both branches were faster at 16k). The potential downside here is that it changes our batch cadence.
- **On the interning side, make interning survive a straddling batch.** Carry per-chunk dictionaries within the batch so each value resolves against the dictionary of the chunk it came from, and leave batching alone. The downside here being that we'd introduce a bit more per-batch bookkeeping.

To be honest, I'm not sure what the trade-off here would _actually_ look like or if the juice is worth the squeeze. The existing changes have already reduced allocations quite a bit, so it'd likely require some additional exploration to see what either of these look like (or if they were worth pursuing at all).

Curious what you think?

gunnarmorling:
> To be honest, I'm not sure what the trade-off here would actually look like or if the juice is worth the squeeze

Yeah, agreed; there's probably bigger fish to fry. Related to your first point, there's #748, which actually would be very worthwhile to tackle. The batch sizing currently doesn't account for repeatable columns at all, which oversizes batches for large lists/map (up to the point of OOM-ing which is how I noticed this).

gunnarmorling:
LGTM, merged. Great stuff, thanks, @rionmonster!

Issue #757, opened 2026-07-04
Compact nested leaf values on the drain, not the serial consumer


gunnarmorling:
## Motivation

After #751 (RealView on the drain) and #756 (skip identity compaction), the top
remaining `ColumnReader`-path consumer cost on a nested scan with nulls is
`compactPrimitive` — the gather of present leaf values into the accessor's return
array. It runs on the serial consumer thread even though the drain virtual thread,
which already builds the `RealView` and holds the gather map, sits on an
otherwise-idle carrier pool.

The all-present case is already handled by #756's identity pass-through; this is
about the batches that genuinely have phantom (null/empty parent) positions and so
still need a real gather — e.g. the `dense` density in `NestedListReadBenchmark`.

## Change

On the real-items path, have the drain compact the leaf values (using the map it
just computed) into a new `NestedBatch.realValues` field and let the consumer read
it directly. The `ColumnReader` array-ownership contract holds: `realValues` is a
fresh per-batch array, exactly as the consumer's compacted array is today — only
the thread that fills it changes.

`batch.values` stays raw so the record-selection path (`compactNestedBatch`, which
slices raw values) is unaffected; selection-derived batches carry no `realValues`
and the consumer compacts them lazily as before. The shared gather is extracted to
an internal `LeafCompaction` helper so the drain and the consumer fallback use one
implementation.

## Acceptance

- `compactPrimitive` no longer runs on the consumer for normal (drain-produced)
  nested batches; `columnNested` improves on `dense` with `none` unchanged.
- Array-identity, exact-filter, and nested differential suites stay green.

Issue #756, opened 2026-07-04
Skip leaf value compaction when a nested batch has no phantom positions


gunnarmorling:
## Motivation

On the `ColumnReader` real-items path, `realLeafValues` compacts the leaf value
stream into the array returned by `getInts()`/`getLongs()`/`getDoubles()` whenever
the schema chain has a `REPEATED` layer — `NestedLevelComputer.computeRealView`
returns a non-null `realToRawLeaf` gather map and `compactPrimitive` allocates a
fresh typed array and gathers `out[i] = raw[map[i]]`.

But the gather is only needed to skip **phantom positions** — raw value slots for
null/empty parents that carry no real leaf. A batch with none of those (the whole
all-present case, and any batch whose lists are all present and non-empty) has
`realLeafCount == rawValueCount`, so the map is the identity and the gather simply
reproduces the batch's own value array.

After #751 moved the `RealView` scan off the consumer, `compactPrimitive` is the
top remaining consumer on-CPU frame on the nested read benchmark (~8% on the
all-present `LIST<float64>` scan), most of it this redundant identity gather plus
its allocation.

## Change

Have `computeRealView` return `realToRawLeaf == null` when
`realLeafCount == rawValueCount` — the same pass-through signal a `REPEATED`-free
chain already uses. The consumer then reads the batch's value array directly.
This is allocation-free and preserves the `ColumnReader` array-ownership contract,
because the batch's value array is itself freshly allocated per batch (`trimValues`
copies it at publish).

Batches that do contain null/empty parents keep the gather.

## Acceptance

- `compactPrimitive` no longer runs on all-present nested batches; `columnNested`
  improves on the `none` density with `dense` unchanged (its phantoms still need
  the gather).
- Array-identity and nested differential suites stay green.

Issue #754, opened 2026-07-04
Add a general nested-read JMH benchmark isolating the reconstruction path


gunnarmorling:
## Motivation

The general (non-fast-path) nested read path is the slowest part of the reader — a `LIST<float32>` scan with the fixed-size-list fast path disabled executes on the order of 21× the instructions of the equivalent flat read, almost all of it per-element bookkeeping. Three changes target it: #732 (per-column-worker routing), #750 (bulk-copy present runs), #751 (RealView on the drain). Each needs before/after numbers on a workload that isolates the reconstruction inner loop.

No existing benchmark serves this:

- `NestedPerformanceTest` (Overture Maps) is whole-file, wall-clock, cross-engine over a **string-heavy** schema — string decode and I/O swamp the level-scan/assembly signal these changes move.
- `FixedSizeListDecodeBenchmark` / `FixedSizeListScanBenchmark` measure only the fixed-width **fast path** (#740), not the general nested path #732/#750/#751 improve.

## Proposal

A numeric-only JMH benchmark under `performance-testing/micro-benchmarks`, per the design in `_designs/NESTED_READ_BENCHMARK.md`:

- **`NestedListReadBenchmark`** (#750, #751) — `LIST<primitive>` scanned three ways on the same file: `columnNested` (ColumnReader real-items), `rowNested` (RowReader all-items), `flatFloor` (identical leaves as a plain column). Params: `elem ∈ {int64, float64}` × `nullDensity ∈ {none, sparse, dense}`; plus `list_of_list` / `list_of_struct` depth files.
- **`MixedSchemaReadBenchmark`** (#732) — scan the scalar columns of a `mixed` scalar+list file vs the same scalars in a flat file; the delta is the per-scalar tax #732 removes. Plus `struct` and `repeated_heavy` guards.

Fixtures are generated with parquet-java (`AvroParquetWriter`), DataPageV2, uncompressed, no dictionary, constant total leaf count. A correctness gate folds every contender to a numeric checksum and asserts agreement before timing. Changes are measured branch-vs-main (also #732's acceptance criterion); no `ReaderConfig` toggles.

## Acceptance

- The two benchmark classes and the `NestedReadFileGenerator` build under `performance-testing/micro-benchmarks`.
- The gate passes (all contenders agree) for every fixture.
- Design doc `_designs/NESTED_READ_BENCHMARK.md` marked completed.

Issue #751, opened 2026-07-03
Move computeRealView to the drain; drop ColumnReader-path def/rep level materialization


gunnarmorling:
## Summary

The regular nested read path materializes the raw Dremel def/rep level arrays into every batch and recomputes the real-items view from them on the **serial consumer thread**. Both are avoidable for the `ColumnReader` (real-items) path: compute the view once on the drain and stop carrying raw levels.

## Current flow

- Drain (`NestedColumnWorker.computeIndex`) builds `elementValidity` + `multiLevelOffsets` (all-items) and `publishCurrentBatch` copies the raw def/rep level arrays into the batch (`Arrays.copyOf`, one pair per batch).
- Consumer (`ColumnReader.ensureRealView`) then recomputes a `RealView` (present-collapsed: real-items `layerOffsets`, per-layer `layerValidity`, `leafValidity`, `realToRawLeaf`) from the raw def/rep levels — lazily, on the single consumer thread, backing `getValueCount`/`getLayerOffsets`/`getLayerValidity`/`getLeafValidity` and the value compaction.

Profiling a nested scan attributes ~11–19% to consumer-side `computeRealView` and a further ~19% to allocating/zeroing/copying the transient level (and derived) arrays.

## Proposal

- Compute `RealView` on the **drain** (in `computeIndex`), store it as `batch.realView`; `ensureRealView` returns it. This moves the level scan off the serial consumer onto the parallel per-column drain threads.
- With the ColumnReader path no longer reading raw levels, **drop the per-batch def/rep level materialization** (the `Arrays.copyOf` in `publishCurrentBatch`) for that path.
- **Gate** `RealView` computation to real-items reads: the `RowReader` path uses the all-items model (`NestedBatchIndex.getDefLevel`) and must not pay for a `RealView` it never reads.

## Scope and dependencies

- **ColumnReader / real-items path only.** The `RowReader` genuinely needs raw def levels for multi-level null disambiguation — `NestedBatchIndex.getDefLevel` is consumed by `PqListImpl` and, extensively, by `VariantShredReassembler` (10 sites comparing def against per-field shredding thresholds). Migrating that off raw levels is a separate, larger effort and is **not** in scope here; the RowReader keeps materializing levels.
[...]

PR #760, opened 2026-07-05
Faster nested reads: reconstruct on the drain, not the serial consumer

### LINKED ISSUE #749: Drop the raw rep/def level escape-hatch accessors from ColumnReader
`ColumnReader` exposes two public escape-hatch accessors, `getDefinitionLevels()` and `getRepetitionLevels()`, that hand back the raw Dremel level arrays for the current batch. They should be removed from the public surface.

## Why

- **No consumer.** The only caller anywhere in the repo (or the `_tmp` integrations) is `FixedSizeListFastPathReadTest`, which exists to assert that the fast path synthesizes the levels correctly. The feature and its only justification are the same code.
- **Redundant with the layer model.** Given the schema, the layer view (`getLayerOffsets`, `getLayerValidity`, `getLeafValidity`, `getValueCount`) fully determines def/rep. Raw levels are a lower-altitude restatement of information the intended API already exposes.
- **Useless for flat consumers.** Both accessors return `null` for flat columns, so the audience that only reads flat columns never touches them.
- **They tax the optimization they ride on.** The fixed-size-list fast path exists precisely to *not* materialize levels. These accessors force it to be able to reproduce exactly what it deliberately skipped (on-demand synthesis in the accessor plus the tests pinning that the synthesis matches the regular path), partially defeating the fast path's premise.
- **Subtly misleading on the fast path.** A caller reaching for `getDefinitionLevels()` likely wants the actual decoded on-wire stream; on the fast path that stream never existed, so they get synthesized-canonical levels instead.

They are marked `@Experimental`, so removal carries no deprecation obligation.

## Scope

- Remove `ColumnReader.getDefinitionLevels()` and `getRepetitionLevels()`.
- Drop the on-demand fast-path level synthesis that exists solely to back them, and the corresponding assertions in `FixedSizeListFastPathReadTest`.
- If a genuine Dremel-level interop consumer materializes later (re-encoding, a variant/shredding bridge), reintroduce the capability via an explicitly-internal API rather than a public accessor — in-repo tooling can already depend on `internal.*` directly.

gunnarmorling:
## Summary

Reading a nested column (`LIST`/`MAP`/struct) through `ColumnReader` reconstructs
the real-items view — level scan, per-layer offsets/validity, and value compaction —
on the **serial consumer thread**, while the parallel per-column decode/drain
threads sit mostly idle. Profiling a `LIST<float64>` scan showed the decode threads
~96% parked and the consumer's on-CPU time dominated by that reconstruction.

This PR moves the reconstruction onto the drain threads and removes redundant work,
so the consumer does little beyond handing back already-built arrays. Nested
`ColumnReader` reads get **~19–24% faster single-column and ~55–59% faster
across a wide (8-column) projection**, with the row-reader and flat-read paths
unchanged.

Closes #749, #751, #756, #757, #754.

## What changed

**#751 — Compute the `RealView` on the drain.** A `NestedColumnWorker` now takes an
`IndexMode` chosen by the constructing reader:
- `REAL_VIEW` (unfiltered `ColumnReader`) builds the real-items view on the drain
  from the level accumulators and **drops the raw def/rep levels and the all-items
  index** — nothing downstream reads them. This eliminates a full level scan on the
  serial consumer *and* two `Arrays.copyOf(valueCount)` per batch of level
  materialisation, cutting per-batch allocation and memory traffic.
- `REAL_VIEW_KEEP_LEVELS` (exact-filtered `ColumnReader`) keeps the raw levels so
  `applySelection` slices them per kept record, unchanged.
- `ALL_ITEMS` (`RowReader`) is untouched — `NestedBatchIndex` still reads raw levels.

**#756 — Skip compaction when there are no phantom positions.** `computeRealView`
returns a `null` gather map when every raw slot is a real leaf (the all-present
case), so the consumer passes the batch's own value array through instead of
allocating and gathering a compacted copy. The array is already fresh per batch, so
the ownership contract holds.

**#757 — Compact leaf values on the drain.** When a gather *is* needed (null/empty
parents), it runs on the drain into `NestedBatch.realValues` rather than on the
serial consumer. Shared gather logic is extracted to `LeafCompaction`.

**#749 — Remove the raw-level escape-hatch accessors.** `ColumnReader.getDefinitionLevels()`
/ `getRepetitionLevels()` are removed. They had no callers, returned `null` for flat
columns, and were the last raw-level readers on the unfiltered path — blocking the
level-drop above. **API note:** breaking removal on the `@Experimental` batch API.

**#754 — Benchmark suite.** JMH benchmarks isolating the nested reconstruction inner
loop against a flat decode floor: `NestedListReadBenchmark` (single `LIST<primitive>`),
`NestedMultiListReadBenchmark` (wide projection of such columns), and
`MixedSchemaReadBenchmark`, plus shared fixtures and correctness gates.

## Performance

`LIST<float64>`, 8M leaves, JMH avgt (2 forks × 5 iters, mean ± 99.9% CI).
Before = base + benchmark; after = this PR. Hardware: MacBook Pro M3 Max.

| Benchmark | density | before (ms) | after (ms) | change |
|---|---|--:|--:|--:|
| `columnNested` (1 column) | none | 65.9 | 53.4 | **−19%** (1.24×) |
| `columnNested` (1 column) | dense | 151.3 | 115.1 | **−24%** (1.32×) |
| `columnMulti` (8 columns) | none | 36.3 | 16.1 | **−55%** (2.25×) |
| `columnMulti` (8 columns) | dense | 97.6 | 40.2 | **−59%** (2.43×) |

Controls unchanged (≤2.4%, within noise): `rowNested` 44.9→45.6 / 96.3→97.5,
`rowMulti` 29.1→29.5 / 47.8→47.6, `flatFloor` 7.4→7.5 / 36.9→36.8, `flatMultiFloor`
9.0→9.0 / 27.9→28.6.

In the wide case the columnar path flips from **slower** than the row reader
(`columnMulti` 36.3 vs `rowMulti` 29.1) to **~1.8× faster** (16.1 vs 29.5), reaching
1.8× the flat decode floor (down from 4.0×). Reproduced with the same direction on a
second, different-architecture machine (Intel N300, x86 single memory channel:
−19% single / −38% multi), so the win is not hardware-specific.

## Design

See `_designs/NESTED_REALVIEW_ON_DRAIN.md` for the `IndexMode` model, the drain vs
consumer split, and the filtered/unfiltered distinction.

## Notes / follow-ups

- The per-column-worker routing for mixed flat+nested files (#732) remains future
  work; this PR converges the real-items reconstruction it would build on.

### DIFF
--- a/_designs/NESTED_READ_BENCHMARK.md
+++ b/_designs/NESTED_READ_BENCHMARK.md
@@ -148,15 +148,21 @@ null density `d ∈ {none, sparse, dense}`:
 Same file, three read paths, plus the floor:
 
 - `columnNested` — open a `ColumnReader` on the leaf, and for each batch fold the
-  compacted real-items leaf array (`getFloats()`/`getLongs()`/… over `getValueCount()`).
-  This is the path #750 and #751 accelerate.
+  compacted real-items leaf array (`getFloats()`/`getLongs()`/… over `getValueCount()`)
+  with `getLeafValidity()`. A flat-leaf consumer (values, count, leaf validity — no
+  offsets); the vector/embedding read pattern. This is the path #750 and #751
+  accelerate.
+- `columnNestedStructural` — the same `ColumnReader` path as a list-reconstructing
+  consumer: read `getLayerOffsets()`/`getLayerValidity()` and walk the lists. It reads
+  the per-layer view every batch, so it measures whether that view is built on the
+  drain or lazily on the serial consumer.
 - `rowNested` — open a `RowReader`, materialize the `PqList` per row, and fold its
   numeric elements. The ergonomic all-items path; #751 must not regress it.
 - `flatFloor` — a `ColumnReader` over the flat-twin leaf. The absolute decode floor.
 
-The gap between `columnNested`/`rowNested` and `flatFloor` is the reconstruction cost;
-these three methods let a change move the nested numbers toward the floor while the
-floor itself stays put.
+The gap between the `columnNested*`/`rowNested` paths and `flatFloor` is the
+reconstruction cost; these methods let a change move the nested numbers toward the
+floor while the floor itself stays put.
 
 ### `MixedSchemaReadBenchmark`
 
--- a/_designs/NESTED_READ_BENCHMARK.md
+++ b/_designs/NESTED_READ_BENCHMARK.md
@@ -1,0 +1,212 @@
+# Nested read JMH benchmark
+
+Status: **Implemented**
+
+## Motivation
+
+The general (non-fast-path) nested read path is the slowest part of the reader. For
+a `LIST<float32>` scan with the fixed-size-list fast path disabled it executes on
+the order of 21× the instructions of the equivalent flat read, almost all of it
+per-element bookkeeping around a two-instruction value copy. Three changes target
+this path:
+
+- **#732** — push the flat/nested split down from the reader to the per-column
+  worker, so non-repeated columns in a mixed file read at flat speed instead of
+  being dragged onto the assembly path.
+- **#750** — bulk-copy contiguous present, unmasked runs in `assembleRegularPage`
+  instead of copying one element at a time.
+- **#751** — compute the `RealView` on the parallel drain and drop the per-batch
+  def/rep level materialization on the `ColumnReader` path.
+
+Each of these needs before/after numbers on a workload that isolates the nested
+reconstruction inner loop. No such benchmark exists:
+
+- `NestedPerformanceTest` (end-to-end, Overture Maps places) is a whole-file,
+  wall-clock, cross-engine comparison over a **string-heavy** schema. String decode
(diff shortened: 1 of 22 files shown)


gunnarmorling:
Hey @iifawzi, @rionmonster, spent a fair amount of time on increasing performance the nested column reader path, here's the result. Quite happy with the numbers. Any feedback welcome!

rionmonster:
@gunnarmorling 

Cool! I’ll be whittling away at the backlog of existing 1.0.1 bugs tonight and I’ll check this out and play around!

gunnarmorling:
Awesome, thank you! Btw. while there's a clear improvement visible with these new JMH-based benchmarks in this PR, `NestedPerformanceTest` doesn't really benefit on my machine. Not sure why, it could be that the string-reuse work could unlock this one.

PR #761, opened 2026-07-06
#714 Bounds-check Variant value lengths and extents

rionmonster:
## Description
This pull request addresses #714, which observed that the Variant value decoder sizes string, binary, and short-string payloads and, through `value()`, whole object / array and fixed-width values from lengths and offsets read straight out of the value buffer without validating them. A hostile or truncated length drove `PqVariantImpl.value()` / `asBinary` / `asString` into an out-of-range (`~1 GiB` or negative) `new byte[...]` or an out-of-bounds read rather than a clean rejection.

The work is two commits: the first guards the primitive / short-string length prefixes (the issue's three fixtures); the second closes the same failure on the `value()` path for object / array containers and truncated fixed-width primitives, surfaced during review. The change is internal to the decoder (the public `PqVariant` surface is unchanged) and well-formed files decode exactly as before.

## Key Changes
- **Generalized the #713 fit-check helper**
    - `VariantBinary.checkHeaderFits` → `checkFits(region, declared, end, bufferLength)`. It already validated a `long`-computed end offset against the buffer for object / array / metadata header tables; it now backs payload-length and value-extent checks too, so there is one validated-extent primitive rather than a parallel helper.
- **Guarded the string / binary and short-string length reads** 
    - A single `VariantValueDecoder.stringPayloadLength` reads the 4-byte length prefix, confirms the prefix is present, and validates the payload fits via unsigned `long` arithmetic; `asString`, `asBinary`, and `primitivePayloadLength` all route through it. `valueLength` gains the analogous short-string check.
- **Guarded the object / array and fixed-width value extents** 
    - `valueLength`'s object / array branches validate the offset table's trailing offset against the buffer (`objectOrArrayValueLength`), and the primitive branch validates the full fixed-width value extent so `value()` on any malformed value throws a descriptive `IllegalArgumentException` instead of over-allocating or reading out of bounds.

## Verification and Testing
Coverage drives the real decode / accessor entry points with crafted headers rather than mocks, and every guard test was confirmed to fail before its guard was added:
- `PqVariantInvalidInputTest` - `oversizedPrimitiveLengthRejected`, `shortStringLengthExceedsBufferRejected`, `truncatedPrimitiveSizeRejected`, `binaryAccessorOversizedLengthRejected`, `stringAccessorOversizedLengthRejected`, `objectValueExtentRejected`, `fixedWidthPrimitiveTruncationRejected`, plus a positive `validPrimitiveStringDecodes`. Pre-fix these produced the exact incidental failures the issue names - `NegativeArraySizeException: -2147483644` / `-2147483643` and `ArrayIndexOutOfBoundsException` on the truncated / short-string paths and now all reject with a descriptive `IllegalArgumentException`.
- #713's overflow tests still pin the object / array / metadata header checks against the renamed `checkFits`.
- Full `core` module green via `./mvnw -pl core -am verify`: `Tests run: 1606` (unit) + `14` (integration), `Failures: 0, Errors: 0`, including the Error Prone `NoVar` / `NoLegacyJavadoc` gates.

The guards run once per value decode (not per byte), and only on the already-allocating `value()` / `asString` / `asBinary` accessors, the typed numeric accessors (e.g. `asInt`, `asLong`, `asDouble`), don't touch these paths. On the success path each guard is a comparison plus `Math.toIntExact`, with the message built only on the throw path, so there is no measurable decode-path cost and no benchmark is warranted for this change.

## Checklist

### DIFF
--- a/core/src/main/java/dev/hardwood/internal/variant/VariantBinary.java
+++ b/core/src/main/java/dev/hardwood/internal/variant/VariantBinary.java
@@ -107,28 +107,27 @@ public final class VariantBinary {
         return result;
     }
 
-    /// Validates that a Variant object/array/metadata header fits inside a buffer
-    /// of `bufferLength` bytes. `headerEnd` is the exclusive end offset of the
-    /// header's id and offset tables — the caller computes it in `long` from the
-    /// **unsigned** element or dictionary `count`, so a hostile count cannot wrap
-    /// the 32-bit product into an in-bounds-looking offset that later slips past
-    /// the per-access bounds checks. Returns `headerEnd` narrowed to `int` once it
-    /// is known to fit.
+    /// Validates that `end`, an exclusive buffer offset derived from an untrusted
+    /// count or length, lies within `bufferLength`, returning it narrowed to `int`.
+    /// The caller computes `end` in `long` from the **unsigned** value, so a hostile
+    /// count/length cannot wrap the 32-bit arithmetic into an in-bounds-looking
+    /// offset that later slips past the per-access bounds checks. Used for
+    /// object/array/metadata header tables and for string/binary payload lengths alike.
     ///
-    /// @param region human-readable name of the count, used in the error message
-    /// @param count declared element/dictionary count (interpreted as unsigned 32-bit)
-    /// @param headerEnd exclusive end offset of the id/offset tables, computed in `long`
+    /// @param region human-readable name of the count/length, used in the error message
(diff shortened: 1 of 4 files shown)


rionmonster:
@gunnarmorling 

Here's another bug fixed for the #711 variant checklist (and one more step towards completing the 1.0.1 milestone)!  🎉

gunnarmorling (approved):
LGTM, thanks, @rionmonster! Added one more commit with an additional test case. Merging.

Issue #710, opened 2026-06-26
Variant decoder: reject excessively deep nested children


gunnarmorling:
## Problem

Hardwood's Variant decoder has no recursion-depth guard. A variant whose children nest arbitrarily deep is decoded without complaint, where a robust decoder should reject it. parquet-java and Spark both cap the nesting depth; Hardwood does not.

This surfaced while probing the malformed-variant fixtures from [apache/parquet-testing#113](https://github.com/apache/parquet-testing/pull/113). Of the 12 fixtures, 11 are rejected by Hardwood's decoder; the one that is **not** is `over_deep_nested_children.parquet`, which nests variant children ~500 levels deep.

From that PR's `bad_data/README.md`:

> `over_deep_nested_children.parquet`: The hierarchy is excessively deep. [...] verifies that nested variant children over 500 levels deep is rejected. This number is subjective; it was chosen to be consistent with the JSON parser `org.apache.parquet.variant.VariantJsonParser`.

Unbounded recursion through the value tree is a denial-of-service vector (stack exhaustion) on hostile input, so this is worth fixing independently of the test-enablement work.

## Proposed fix

- Thread a depth counter through the Variant value navigation (`VariantValueDecoder` / `PqVariantObjectImpl` / `PqVariantArrayImpl`) and fail fast with a clear `IllegalArgumentException` once a configured maximum nesting depth is exceeded.
- Adopt a limit consistent with the upstream convention (~500), defined as a single named constant.
- Add a failing test first (per the bug-fix workflow), then assert it is rejected.

## Acceptance criteria

- `over_deep_nested_children.parquet` is gracefully rejected with a descriptive message naming the depth limit.
- A unit test pins the boundary (a variant at the limit decodes; one past it is rejected).

## Related

- Test harness / enablement tracked in the companion issue (variant hardening fixtures from parquet-testing#113).

PR #765, opened 2026-07-06
#710 Thread a nesting-depth cap through Variant value navigation

rionmonster:
## Description
This pull request addresses #710, which observed that Hardwood's Variant decoder had no recursion-depth guard: a Variant value nesting objects or arrays arbitrarily deep could be navigated without bound, letting a consumer that walks the value tree recurse until the stack is exhausted — a denial-of-service on hostile input. parquet-java and Spark both cap Variant nesting depth (~500 levels); this brings Hardwood in line, and closes the last parquet-testing#113 malformed-variant fixture (`over_deep_nested_children.parquet`) that Hardwood did not previously reject.

The change is internal to the Variant read path; the public `PqVariant` / `PqVariantObject` / `PqVariantArray` surface is unchanged, and well-formed values navigate exactly as before.

## Key Changes
- **Depth carried on the flyweights, checked at every descent**
    - `PqVariantImpl`, `PqVariantObjectImpl`, and `PqVariantArrayImpl` each carry a `depth`, and a single `VariantValueDecoder.descend(parentDepth)` returns `parentDepth + 1` after checking it against `MAX_NESTING_DEPTH`, throwing a descriptive `IllegalArgumentException` otherwise. It is called at every child-producing site — object field access (`getVariant` / `getValue` / `getObject` / `getArray`) and array element access (`get`) — so object, array, and mixed nesting are counted through one funnel. `asObject` / `asArray` are same-value views and pass the depth through unchanged.
- **One source of truth for the limit**
    - `MAX_NESTING_DEPTH = 500`, a single named constant matching the parquet-java / Spark convention; the exception message names the limit.

## Verification and Testing
Coverage drives the real public navigation API with values built in-memory via the existing `VariantValueEncoder` — no fixture dependency:
- `PqVariantInvalidInputTest` — `deeplyNestedArrayRejected` / `deeplyNestedObjectRejected` cover the array (`get`) and object (`getObject`) descent paths; `arrayNestedAtLimitNavigatesFully` / `arrayNestedOnePastLimitRejected` / `objectNestedAtLimitNavigatesFully` pin the boundary on both paths (exactly 500 levels navigates and reads the leaf back, 501 rejected); and `arrayIterationNestingRejected` confirms the `Iterable` for-each route, which goes through the guarded `get(int)`. Each rejection case was confirmed to fail before the guard — deep values were previously walked to the bottom (silent accept on arrays, a late `VariantTypeException` on objects).
- Full `core` module green via `./mvnw -pl core -am verify`: `Tests run: 1618` (unit) + `14` (integration), `Failures: 0, Errors: 0`, including the Error Prone `NoVar` / `NoLegacyJavadoc` gates.

The guard adds one comparison per navigation step and no allocation on the success path, so there is no measurable navigation cost.

## Checklist

### DIFF
--- a/core/src/main/java/dev/hardwood/internal/variant/PqVariantArrayImpl.java
+++ b/core/src/main/java/dev/hardwood/internal/variant/PqVariantArrayImpl.java
@@ -18,11 +18,13 @@ final class PqVariantArrayImpl implements PqVariantArray {
     private final VariantMetadata metadata;
     private final byte[] valueBuf;
     private final ArrayLayout layout;
+    private final int depth;
 
-    PqVariantArrayImpl(VariantMetadata metadata, byte[] valueBuf, int arrayHeaderOffset) {
+    PqVariantArrayImpl(VariantMetadata metadata, byte[] valueBuf, int arrayHeaderOffset, int depth) {
         this.metadata = metadata;
         this.valueBuf = valueBuf;
         this.layout = VariantValueDecoder.parseArray(valueBuf, arrayHeaderOffset);
+        this.depth = depth;
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


rionmonster:
@gunnarmorling 

Uno mas for the Variant-related bugs list. 👍

gunnarmorling (approved):
LGTM, merging. Muchas gracias, @rionmonster!

PR #774, opened 2026-07-10
#773 Skip decode setup when pruning leaves zero row groups

### LINKED ISSUE #773: Filtered reads allocate decode buffers and spawn column workers even when pruning leaves zero row groups
When a filtered read prunes **every** row group (statistics/bloom), `ParquetFileReader.buildColumnReaders` still eagerly builds the full decode pipeline before discovering there's nothing to read:

- per-column batch buffers via `BatchExchange.allocateArray`,
- the column worker threads via `ColumnWorker.start`,
- plus `SelectionEngine` + `FilterCoordinator`.

For a zero-row result that's pure waste — on the order of ~8 MB and ~11 ms per op of setup that is immediately discarded.

### How it surfaced

This became clear while testing the **bloom-filter absent** point lookup (a key present in no row group): the bloom filter drops every row group, so the work-list is empty, yet Hardwood still paid the full per-column setup. On that case it ended up ~5× slower than parquet-java's low-level reader, with the time dominated by allocation + the worker-handoff wait (`BatchExchange.poll → __psynch_cvwait`) rather than any real decode work. Allocation profiling attributes the ~8 MB to `BatchExchange.allocateArray` (the `long[]`/`double[]` batch buffers) and `SelectionEngine.create` (the `int[]` selection buffer).

### Why it's avoidable

The surviving row-group set is known **eagerly**, right after `RowGroupIterator.initialize()` — `getWorkItems().isEmpty()` — and strictly **before** any `ColumnReader` (hence any worker/buffer) is built. When it's empty we can short-circuit and return **no-op readers** over the payload columns: immediately exhausted, with no worker threads, no batch buffers, and no selection engine / coordinator. The shared `RowGroupIterator` stays owned and closed by the `ParquetFileReader`, so its file handles / prefetch futures still release normally.

### Impact (bloom-absent point lookup, measured)

- time: **~11 ms/op → ~0.4 ms/op**
- allocation: **~8 MB/op → ~17 KB/op**

Correctness unchanged (still zero rows), and the non-empty filtered paths are unaffected.

### Out of scope (follow-up)

Row groups that survive row-group pruning but are then emptied by **column-index / page-level** pruning are decided lazily during decode, so this eager `isEmpty()` check does not cover them; they still incur the setup. Worth tracking separately.

iifawzi:
Hello, this PR resolves #773, returning no-op readers, saving allocations, and workers setup overhead if all row groups are filtered out
## Checklist

### DIFF
--- a/core/src/main/java/dev/hardwood/reader/ColumnReader.java
+++ b/core/src/main/java/dev/hardwood/reader/ColumnReader.java
@@ -140,6 +140,20 @@ public class ColumnReader implements AutoCloseable {
         return new ColumnReader(column, true, layers, nestedBuffer, columnWorker, rowGroupIterator);
     }
 
+    /// A reader that yields no batches, for when row-group pruning has dropped
+    /// every row group so nothing can match. Allocates no batch buffer and
+    /// starts no worker thread — [#nextBatch()] reports exhausted immediately.
+    static ColumnReader exhausted(FileSchema schema, ColumnSchema column) {
+        NestedLevelComputer.Layers layers = NestedLevelComputer.computeLayers(
+                schema.getRootNode(), column.columnIndex());
+        boolean nested = layers.count() > 0 || column.maxRepetitionLevel() > 0;
+        ColumnReader reader = nested
+                ? forNested(column, layers, null, null, null)
+                : forFlat(column, null, null, null);
+        reader.exhausted = true;
+        return reader;
+    }
+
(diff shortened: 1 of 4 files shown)


iifawzi:
Before: 
bloom_chart
After: 99% overhead eliminated, 153x faster
bloom_chart

gunnarmorling:
Awesome, very nice. LGTM, merging. Thanks a lot, @iifawzi!

PR #775, opened 2026-07-10
#739 Verify unknown logical types fall back to the physical type

### LINKED ISSUE #739: Verify that an unknown logical type doesn't cause the reader to fail
It should be handled gracefully, i.e. the physical type should be exposed. Not quite sure how to verify at this point, given we support all currently known logical types; might require to set up a fixture with a bespoke unknown type.

Context: https://lists.apache.org/thread/flfzz94ftdrdop9d5b0o1hkqkprzj3l5

rionmonster:
## Description
Closes #739. A Parquet file written against a newer spec revision may annotate a column with a logical type this reader predates. Per [this Parquet dev discussion](https://lists.apache.org/thread/flfzz94ftdrdop9d5b0o1hkqkprzj3l5), the agreed forward-compatibility contract is that a reader meeting a logical type it doesn't know should treat the annotation as absent, keep the physical schema, and still read the column, rather than failing footer conversion.

Hardwood already behaves this way, so this is a verification rather than a fix: it pins the contract with tests and documents it. No decoder change was required.

## Key Changes
- **Thrift-decode contract (core)**
    - `LogicalTypeReaderTest` — an unrecognized union member decodes to a null annotation; a *parameterized* unknown member is skipped without desyncing the surrounding parse (a sentinel field after the union is read back to prove exact consumption).
    - `SchemaElementReaderTest` — a `SchemaElement` carrying an unknown logical type still exposes its physical type with a null logical type.
- **End-to-end against the canonical fixture (parquet-testing-runner)**
    - `UnknownLogicalTypeReadTest` reads `unknown-logical-type.parquet` from apache/parquet-testing, asserting the unknown-type column exposes `BYTE_ARRAY` with a null logical type and reads its raw values, while the sibling `STRING` column is unaffected. This fixture was previously unverified against Hardwood: the comparison suite skips it because the parquet-java reference reader cannot open it.
- **Forward-compatibility note** added to `concepts/compatibility-philosophy.md`, alongside the existing legacy-annotation discussion.

## Verification and Testing
- `core`: `Tests run: 1615, Failures: 0, Errors: 0`, including the Error Prone `NoVar` / `NoLegacyJavadoc` gates.
- `parquet-testing-runner`: `UnknownLogicalTypeReadTest` green.
- Decode tests confirmed non-vacuous by mutation: forcing the unknown-type arm to throw fails all three; removing its `skipField` fails the sentinel test. Both mutations reverted.

## Follow-ups (out of scope; to track separately)
- Column statistics should not be trusted for pushdown on columns with an unrecognized logical type (requires distinguishing "unknown" from "absent" annotation) — #XXX.
- Reject unknown VARIANT spec versions (`variant(2)`, parquet-testing#113) — #XXX.

## Checklist

### DIFF
--- a/core/src/test/java/dev/hardwood/internal/thrift/LogicalTypeReaderTest.java
+++ b/core/src/test/java/dev/hardwood/internal/thrift/LogicalTypeReaderTest.java
@@ -1,0 +1,68 @@
+/*
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Copyright The original authors
+ *
+ *  Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package dev.hardwood.internal.thrift;
+
+import java.nio.ByteBuffer;
+
+import org.junit.jupiter.api.Test;
+
+import dev.hardwood.internal.thrift.ThriftCompactConstants.FieldType;
+import dev.hardwood.metadata.LogicalType;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/// Unit tests for [LogicalTypeReader], focused on a logical-type union member
+/// the reader does not recognize. A future or bespoke logical type must decode
+/// to a null [LogicalType] rather than failing the read, so the column's
+/// physical type is what gets exposed.
+class LogicalTypeReaderTest {
+
+    @Test
+    void unknownUnionMemberDecodesToNull() throws Exception {
+        ThriftCompactWriter writer = new ThriftCompactWriter();
+        // Unrecognized field id 19 carrying an empty marker struct, the shape
+        // most known logical types take.
+        writer.writeFieldBegin(19, FieldType.STRUCT);
+        writer.writeFieldStop(); // empty member struct
+        writer.writeFieldStop(); // union STOP
+
+        assertThat(read(writer)).isNull();
+    }
+
+    @Test
+    void unknownParameterizedUnionMemberSkippedCleanly() throws Exception {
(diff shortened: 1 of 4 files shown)


rionmonster:
@gunnarmorling 

Just revisited digging into this and it appears that Hardwood is already handling this relatively well (e.g. the unknown annotation is treated as absent, the column keeps its physical type, and it's still readable). I was able to grab one of the variants from `parquet-testing` (e.g. `unknown-logical-type.parquet`), so this overall effort was more of a _verify, not fix_. The PR itself just adds some decode-related tests along with that fixture.

A few follow-ups, one being I think more important than the other, that I'll write up shortly are:

1. **Statistics - #776 (the bigger one):** Unknown logical types shouldn't surface statistics is currently a real gap for us. An unknown annotation and *no* annotation both come through as `logicalType == null`, so we can't distinguish them, and stats pruning could drop a row group wrongly. Needs its own fix.
2. **Variant version - #777:** We need to figure out handling for things like `variant(2)` mentioned in the Parquet thread as we currently accept any `specVersion >= 1` rather than rejecting versions we don't understand.

Lastly, and totally unrelated, but I think I may have come across a flaky test while I was running through this effort. I'm having Claude explore it a bit, but it flagged the `ColumnWorkerTest.closeJoinsThreadsAndAwaitsInFlightDecodes` as flaky (seems to be test-instrumentation related) which I've documented in #778.

gunnarmorling (approved):
Very nice, merging. Thanks a lot for the continued push on the correctness and reliability front, @rionmonster, this is invaluable!

PR #779, opened 2026-07-10
#778 Fix ColumnWorker test instrumentation race

### LINKED ISSUE #778: Flaky: ColumnWorkerTest.closeJoinsThreadsAndAwaitsInFlightDecodes races on the instrumentation counter
## Problem

`ColumnWorkerTest.closeJoinsThreadsAndAwaitsInFlightDecodes` fails intermittently (~1-in-5 in isolation on `main`) with:

```
[every submitted decode task should have finished]
expected: <N> but was: <N-1>
```

## Root cause

Test instrumentation race, not a product bug. The test's `stalledExecutor` wrapper increments `decodesFinished` *after* `command.run()` returns. But `worker.close()` only guarantees that `command.run()` has returned — that is the signal it awaits — so `close()` can legitimately return in the window between `command.run()` completing and the wrapper thread reaching `decodesFinished.incrementAndGet()`. The final assertion then reads the counter one short. `worker.close()`'s draining behavior is correct.

## Proposed fix

- Await the counter before asserting (poll to a short deadline, mirroring the existing thread-liveness awaits in the same test), or increment `decodesFinished` inside `command` so it is part of what `close()` drains.

## Acceptance criteria

- The test passes deterministically under repeated runs (e.g. 50x in isolation) with no product-code change.

## How it surfaced

Spotted during a full `./mvnw clean verify` while landing #739; unrelated to that change.

w3lld1:
## Summary

Wait briefly for the executor-wrapper instrumentation counter to catch up before comparing it with the submitted decode count. `worker.close()` correctly waits for the decode command, but the test wrapper increments its counter immediately after that command returns, leaving a small assertion race.

No product code is changed.

## Validation

- `./mvnw process-sources -DskipTests -pl :hardwood-core`
- Targeted test repeated 50 times: 50/50 passed
- `./mvnw verify -pl :hardwood-core` — 1,665 unit tests and 14 integration tests passed
- `git diff --check`

The repository-wide `./mvnw clean verify` was not run because this runner cannot access its Docker daemon; the affected core module's complete `verify` lifecycle passed.

## Checklist

- [ ] Build via `./mvnw clean verify` passes (full multi-module build not run; see validation above)
- [x] If this PR adds or changes user-facing behaviour, `docs/` has been updated (not applicable; test-only change)

Fixes #778

### DIFF
--- a/core/src/test/java/dev/hardwood/internal/reader/ColumnWorkerTest.java
+++ b/core/src/test/java/dev/hardwood/internal/reader/ColumnWorkerTest.java
@@ -382,6 +382,12 @@ class ColumnWorkerTest {
             assertThat(worker.drainThread.isAlive())
                     .as("drain thread must have exited")
                     .isFalse();
+
+            long deadline = System.nanoTime() + TimeUnit.SECONDS.toNanos(5);
+            while (decodesFinished.get() != decodesEntered.get()
+                    && System.nanoTime() < deadline) {
+                Thread.sleep(10);
+            }
             assertThat(decodesFinished.get())
                     .as("every submitted decode task should have finished")
                     .isEqualTo(decodesEntered.get());


rionmonster (approved):
Thanks for the contribution to Hardwood @w3lld1! This looks good to me, approving and merging.

Issue #750, opened 2026-07-03
Regular nested assembly copies values one-at-a-time: bulk-copy present runs (generalize the fast path's run copy)


gunnarmorling:
## Summary

The regular (non-fast) nested assembly path in `NestedColumnWorker.assembleRegularPage` materializes values **one element at a time** via `copyOneValue`, wrapped in a per-element sequence that dominates its cost. Profiling a `LIST<float32>` scan with the fast path **disabled** shows the reconstruction path executes ~21× the instructions of a flat read, and most of that is per-element bookkeeping around a 2-instruction value copy.

This is the general-path analogue of the fixed-size-list fast path (#741): where the fast path bulk-copies a fixed-width present run (`copyValueRun`), the regular path should bulk-copy any contiguous **present, unmasked** run.

## Evidence (N300, k=768, 8M values, `columnBaseline` vs `flatFloor`)

`perfnorm` — the gap is compute, not memory:

| per op | columnBaseline | flatFloor |
|---|--:|--:|
| time | 157 ms | 15 ms |
| instructions | 1.26 B | 60 M (**21×**) |
| IPC | 1.58 (CPU-bound) | 0.41 (mem-stalled) |
| LLC load-misses | 824 K (0.15%) | 57 K |

`async` CPU self-time (~74% is reconstruction the flat read never does):
- array zeroing on allocation (`__memset` under `ObjArrayAllocator`) ~19%
- transient array copies (`Arrays.copyOf`, arraycopy stubs) ~19%
- level -> offsets/validity scan (`NestedLevelComputer.computeRealView`) ~19%
- per-element value copy (`copyOneValue`) ~9%
- assembly loop (`assembleRegularPage`) ~6%
- *value decode shared with flat (`readFloats`, `ByteBuffer.get`) ~11%*

`perfasm` — `copyOneValue` is inlined into `assembleRegularPage`; the value load/store is the cheap part. Per value (8M×) the loop runs ~20 instructions: `ensureLevelCapacity` **+** `ensureValueCapacity` (two capacity checks), def-level and rep-level `iastore` each guarded by an array **bounds check**, and a **checkcast** of `nestedValues` to `float[]` (residue of the `switch (page)` C2 pruned to the FloatPage case). No single hot instruction — per-element overhead spread thin.

## Proposed fix

For **present, unmasked runs** (`def == maxDef`, within the mask), stop working per element:

[...]

PR #789, opened 2026-07-15
Fan-out-aware batch sizing + bulk-copy all-present nested assembly (#748, #786)

### LINKED ISSUE #748: Batch sizing ignores list fan-out: high-k repeated columns OOME (whole file drains into one batch)

### LINKED ISSUE #786: Bulk-copy whole all-present pages in regular nested assembly
## Summary

`NestedColumnWorker.assembleRegularPage` materializes leaf values **one element at a time** (`copyOneValue`), wrapped in per-element bookkeeping — two capacity checks, two bounds-checked level `iastore`s, a destination `checkcast` — that dominates its cost. On an all-present scan the regular path executes ~20× the instructions of the equivalent flat read, almost all of it this per-element overhead around a 2-instruction value copy.

When an **entire page** is all-present, its leaf values are one contiguous block, 1:1 with page positions — presence, not shape, is what makes them contiguous. Such a page can be gathered with a single `System.arraycopy`.

Keyed on presence, this is **not list-specific**: it applies to any column the regular nested assembly path serves — lists (fixed or variable), maps, structs — and, until #732 gives non-repeated columns direct addressing, the flat columns of a mixed dataset currently routed through the nested worker.

## Scope: whole page, gated O(1)

Take the fast path only when both hold, decided per page **without scanning levels**:

- **all-present** — the definition-level stream is a single RLE run of `maxDefinitionLevel` over all `numValues` (`isSingleRleRunOf` — one varint + value byte), and
- **unmasked** — `mask.isAll()`.

Otherwise fall back wholesale to the existing per-element path. The gate is O(1) and adds nothing to the fallback, so the win is free to attempt.

Sub-page present-run copying (scanning def for runs broken by nulls/mask gaps and bulk-copying each) is **not** in scope: it was tried and the O(n) per-run detection + boundary bookkeeping offset the saving. That generalization is a potential follow-up, tracked in #750.

## Design

Full design + benchmark plan: `_designs/NESTED_ALL_PRESENT_PAGE_BULK_COPY.md`. In brief: resolve the page value type once, bulk-`arraycopy` the page's values in batch-capacity-aligned (record-aligned) chunks, and produce the batch structure as cheaply as the index mode allows —

- **`REAL_VIEW`** (unfiltered `ColumnReader`, which since #751 derives offsets+validity on the drain and drops raw levels): build that view directly — offsets from a repetition-boundary scan, validity all-true — and **do not** materialize raw level arrays (the same output the fixed-size-list fast path already produces).
- **`REAL_VIEW_KEEP_LEVELS` / `ALL_ITEMS`**: bulk-fill the level arrays (def = constant `maxDefinitionLevel`, rep = `arraycopy` from the page) instead of per-element.

Shares the bulk-copy primitive with `FlatColumnWorker` and the fixed-size-list fast path (#741 — already shipped); that path's per-record `copyValueRun` is refactored onto the shared primitive.

## Relationships

- **Builds on #751** (done) — level materialization dropped on the drain; this makes producing the `REAL_VIEW` for an all-present page cheaper (direct, not per-element).
- **Generalizes #741** (done) — the fixed-size-list fast path is the fixed-`k` specialization (arithmetic offsets + skipped rep scan); this is the any-shape all-present case sharing one primitive.
- **Potential follow-up #750** — sub-page present runs (tried, didn't pay off; revisit).
- **Interacts with #732** — once non-repeated columns get direct addressing, they leave the nested worker and no longer rely on this path.

## Baseline note

The fast path (#741) currently reports its speedup against the per-element regular path as baseline, which overstates the ratio. This change makes that baseline honest — after it, the fast path's remaining advantage is only the arithmetic offsets and the skipped repetition scan.

## Acceptance

- Regular nested path (fast-path-off) instructions/op and wall time on all-present data materially reduced toward the flat floor.
- Correctness preserved across nulls, empties, masking (incl. the exact-filter keep-levels path), dictionary / byte-array / FLBA, and level semantics; differential + nested tests green.

gunnarmorling:
Two related improvements to the nested / fixed-size-list read path, landing together (the second builds on the first).

## #786 — bulk-copy all-present nested values

The regular nested assembly path materialized leaf values one element at a time, spending most of its instructions on per-element bookkeeping (capacity checks, bounds-checked level stores, a destination checkcast) around a two-instruction value copy. When a page is proven all-present by `PageDecoder`'s O(1) def-gate — now applied to nested columns, not only flat — its leaves are one contiguous, record-aligned block, now gathered in bulk: values, the constant definition levels, and the repetition levels once per batch. Byte-array leaves and masked or null-bearing pages keep the per-element path.

The fixed-size-list fast path, which previously copied one *k*-value record at a time (an `arraycopy` per record — crippling at small *k*), is rebuilt on the same span-copy primitive: one `copyValueRun` per contiguous run of kept records. Both paths now gather values the same way.

## #748 — fan-out-aware batch sizing

`BatchSizing` budgeted a batch by row count times a per-row byte estimate that assumed **one value per row**. A repeated column expands each row into many leaf values, so a wide `LIST<float32>` (768-element vectors) produced a value array ~768× the intended 6 MB L2 budget — a ~512 MB batch whenever the whole column fits under `MAX_BATCH` rows, memory-bound to assemble and to scan, and an outright OOME at higher fan-out.

The budget now folds in each column's average list fan-out (from column-chunk metadata) plus the per-value level bytes a repeated column's nested worker allocates, and the row floor is dropped. Every auto-sizing site — the single-column `ColumnReader`, the multi-column / filtered `ColumnReaders`, and `NestedRowReader` — routes through the same computation, so both reader regimes produce identical batch sizes for a projection (`FlatRowReader` is one value per row and unchanged).

## Results

Intel N300, **all 8 cores**, warm cache, `LIST<float32>` 768-element vectors, 128M values/file. Throughput in M float32 values/s (higher is better):

| contender | before (main) | after (this PR) |
|---|---:|---:|
| columnFast | 209 | **357** |
| columnBaseline | 90 | 202 |
| rowFast | 207 | **349** |
| rowBaseline | 72 | 147 |
| flat-column floor | 592 | 576 |

Every reader roughly doubles, and the fixed-size-list fast path moves from **2.8× below** the flat-column decode floor to **~1.6×**.

### Headline — reading 768-wide vectors (k=768, 128M values)

| before (main) | after (this PR) |
|:---:|:---:|
| headline before | headline after |

### Speedup over record-reconstruction across k (8M values)

| before (main) | after (this PR) |
|:---:|:---:|
| sweep before | sweep after |

The fast-path-over-baseline **ratio** shrinks at large *k* after the PR (column 2.4× → 1.8×) — this is not a regression: #786 makes the reconstruction *baseline* itself much faster, so the fast path's relative lead narrows even as absolute times fall for both. The visible sweep win is at small *k*, where the per-run copy lifts the column fast path from **0.9× (slower than the baseline) → 1.1×**.

### Balanced machine — AWS m7i.2xlarge, all cores (k=768, 128M values)

The N300 above has a single memory channel, so an all-core scan is bandwidth-bound and the fast path — which moves more memory — can't reach the floor there (its all-core ~1.6× gap is in fact *worse* than single-core's ~1.07×, as 8 cores saturate the one channel). On a balanced multi-channel host that ceiling lifts and the fast path reaches the flat-column decode floor:

headline aws

Column fast **866** vs. floor **879** M values/s — **1.4% off**, 1.9× over baseline; row fast **860** (4.2× over baseline), also at the floor.

## Validation

- Full `core` suite green (**1698** tests), including a new OOM regression test (`WideListBatchBoundTest`) — verified **red without** the fan-out fix (the column drains into one batch) and **green with** it.
- Differential (DuckDB oracle) and nested suites green.

## Notes

- Charts live on the `pr789-charts` assets branch (kept out of this PR's diff; safe to delete after merge).
- Design docs: `NESTED_ALL_PRESENT_PAGE_BULK_COPY.md` (new), `COLUMN_READER_ADAPTIVE_BATCH_SIZING.md` (extended for fan-out), `FIXED_SIZE_LIST_FASTPATH.md` (assembly description).
- Follow-up: sub-page present-run copying (#750).

Closes #786
Closes #748

### DIFF
--- a/_designs/COLUMN_READER_ADAPTIVE_BATCH_SIZING.md
+++ b/_designs/COLUMN_READER_ADAPTIVE_BATCH_SIZING.md
@@ -17,27 +17,32 @@ Status: **Completed**
 
 The reader has two batch-sizing regimes that must agree but did not:
 
-- The `RowReader` path (`FlatRowReader` / `NestedRowReader`) sizes each batch by a **byte budget**: `BatchSizing.computeOptimalBatchSize` targets ~6 MB across all projected columns and clamps the resulting row count to `[16 384, 524 288]`, so the per-batch column arrays stay within L2 cache regardless of column width or projection breadth.
+- The `RowReader` path (`FlatRowReader` / `NestedRowReader`) sizes each batch by a **byte budget**: `BatchSizing.computeOptimalBatchSize` targets ~6 MB across all projected columns and caps the resulting row count at `524 288` (`BatchSizing.MAX_BATCH`), so the per-batch column arrays stay within L2 cache regardless of column width or projection breadth.
 - The `ColumnReader` / `ColumnReaders` builders used a fixed **record count** default of `262 144` rows.
 
 Because the `ColumnReader` default was a fixed row count, the real per-batch footprint floated with the data: ~1 MB for a single `INT32`, ~2 MB for a single `INT64`/`DOUBLE`, tens of MB for a `BYTE_ARRAY` column once payload and `String` objects are counted, and ~210 MB for a 100-column `INT64` projection (each reader allocating its own `262 144`-row arrays). The batch arrays are freshly allocated per batch (no scratch reuse) and `BatchExchange` prefetch adds a queue-depth multiplier on top, so an oversized default translates directly into allocation and GC churn.
 
+The byte budget itself assumed **one value per row per column**. A repeated (list) column expands each row into many leaf values, so its per-batch value array — and the two `int` level arrays its nested worker holds alongside — are the list's *fan-out* times larger than a row-count budget predicts. A wide `LIST<float32>` (768-element vectors) whose row count fits under the clamp drained the entire column into a single ~512 MB batch: far past the L2 target, memory-bound to both assemble and scan, and at higher fan-out an outright heap exhaustion. So even once both regimes were byte-budgeted, a fan-out-blind estimate still oversized repeated columns and left the regimes disagreeing for them.
+
 ## End state
 
-The default batch size for the column path is **byte-budgeted**, routed through the same `BatchSizing.computeOptimalBatchSize` the row path uses. The two regimes now produce identical batch sizes for the same projection.
+The default batch size for the column path is **byte-budgeted**, routed through the same `BatchSizing.computeOptimalBatchSize` the row path uses, and the budget is scaled by each column's **list fan-out**. The two regimes produce identical batch sizes for the same projection.
 
+- The per-row cost of a column is its value width times its average leaf-values-per-row (fan-out), plus the def/rep level bytes a repeated column's nested worker allocates per value (two `int`s). A flat column's fan-out is `1` and it carries no per-value level arrays, so the budget collapses to the plain value width.
+- The row-count floor is removed. Because the value count per batch is now roughly constant across fan-out (`≈ budget / value-bytes`), a high-fan-out column's small row count still carries a full batch of work, so no floor is needed to amortise per-batch overhead. `BatchSizing.MAX_BATCH` (`524 288`) remains the only clamp.
 - `ColumnReaderBuilder.batchSize(int)` and `ColumnReadersBuilder.batchSize(int)` remain the explicit override and are honored verbatim. The positivity check is unchanged.
-- When the caller does not set `batchSize`, the builder leaves it unset and the size is resolved at `build()` time from the projection's column widths.
+- When the caller does not set `batchSize`, the builder leaves it unset and the size is resolved at `build()` time from the projection's column widths and fan-out.
 - The width used is that of **every column the batch decodes**: the full projection on the plain path, and the predicate-augmented projection on the exact-filtering path (the predicate columns allocate arrays too, so they count toward the budget).
-- The public `ColumnReader.DEFAULT_BATCH_SIZE` constant is removed: there is no longer a single fixed default to name. `BatchSizing.MAX_BATCH` (`524 288`) remains the upper clamp.
+- The public `ColumnReader.DEFAULT_BATCH_SIZE` constant is removed: there is no longer a single fixed default to name.
 
 This is a no-op for callers that already pass an explicit `batchSize`. Callers relying on the default get smaller, width-aware batches and lower peak allocation, especially on wide and variable-width projections.
 
 ## Mechanics
 
-`batchSize` carries a sentinel of `0` ("unset") from the builders down to the bridge methods on `ParquetFileReader`. `resolveBatchSize(requested, projectedSchema)` returns `requested` when positive and `BatchSizing.computeOptimalBatchSize(projectedSchema)` otherwise. Resolution happens at the single point where the full `ProjectedSchema` is known:
+`batchSize` carries a sentinel of `0` ("unset") from the builders down to the bridge methods on `ParquetFileReader`. `resolveBatchSize(requested, projectedSchema, rowGroups)` returns `requested` when positive and otherwise `BatchSizing.computeOptimalBatchSize(projectedSchema, BatchSizing.valuesPerRow(projectedSchema, rowGroups))`. `valuesPerRow` derives each projected column's fan-out from the row-group metadata — its total leaf value count (`Σ chunk.numValues`) over the total row count (`Σ rowGroup.numRows`) — so the estimate needs both the projection *and* the file's row groups. Resolution happens where both are known:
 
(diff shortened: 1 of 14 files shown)


gunnarmorling:
@iifawzi, @rionmonster, so this PR brings two improvements to the regular nested column handling: bulk copying of values for all-present pages (#786) and correct sizing of batches for repeated columns (#748; before, batches with embedded lists could grow arbitrarily large).

The motivation--apart from generally improving performance and robustness--is to provide a fairer baseline for the #740 gains I'd like to report (fast path for effectively fixed-size lists). Before, the story wasn't solid: over-reported gains due to the sub-optimal baseline, and at the same time an unreasonable gap to a flat schema file with the same number of total values (which represents the optimal floor, and I didn't see any categorical reason why fixed-size lists should be significantly slower than that).

As the chart from the AWS instance shows, the fast path effectively closes that gap now 🥳. I'm gonna kick of a more extensive run over night and will update the visuals here once more tomorrow. Starting to feel good about this one overall.

iifawzi:
Impressive work on the vectors support overall, and nice catch and fix for the batch size variable. I believe tuning that variable is so critical for performance overall, could even be worth its own iterations. Speedups look quite fair now compared to the baseline

gunnarmorling:
Yes, I think this is good to merge. It definitely was worth iterating here, the general nested path had been a bit neglected before performance-wise, and now it's catching up quite a bit. Thanks for taking a look.

PR #792, opened 2026-07-16
#791 Add Parquet row-group rewrite

### LINKED ISSUE #791: Rewrite capability: transform existing Parquet file(s) into a new file, byte-copying unchanged row groups where possible
## Motivation

Once the writer milestone (#9) lands, Hardwood can read a Parquet file and produce
a Parquet file, but there is no first-class path for **transforming an existing file
(or set of files) into a new one**. That transform is the core of table-maintenance
workloads:

- **Compaction / merge** — many small files → one, concatenating row groups.
- **Filtering / pruning** — drop row groups (or rows) that don't match a predicate,
  rewrite the survivors in place (Iceberg / Delta maintenance).
- **Column projection** — write out a subset of columns.
- **Re-compression / re-encoding** — change codec or encoding of an existing file.
- **Statistics / index backfill** — add stats or page indexes a legacy writer omitted.

This issue captures the **rewrite capability** as a feature in its own right. It is
deliberately scoped as post-#9 work: it composes the reader and the new writer, and
its most valuable pieces depend on writer-side index support, which is itself deferred
past #9.

## What it entails

The design space runs from a trivial full re-encode to a zero-decode byte copy. A
complete rewrite capability spans:

1. **Baseline: decode → re-encode pipeline.** Read to `ColumnBatch`, feed the writer.
   Full fidelity, works the moment #9 is complete, but decodes and re-encodes every
   value — the slow path, and the fallback whenever a transform can't be done any
   other way (row-level filtering within a surviving group, computed columns, codec or
   encoding change).

2. **Byte-copy fast path.** When a whole row group (or a whole column chunk) survives
   unchanged — no row-level filter inside it, no codec/encoding change, projection that
   drops entire columns rather than parts of one — copy its encoded bytes verbatim and
   never decode. This is where the throughput win lives.

3. **Metadata rebasing.** A copied or dropped row group shifts every offset in the
   footer by the delta between its source and destination position: `data_page_offset`,
   `dictionary_page_offset`, `index_page_offset`, the column/offset-index offsets, and
   the legacy `file_offset`. The encoded bytes are untouched; only the pointers move.

4. **Page-index carry-forward** *(implementation note under 2/3)*. On a byte-copy
   rewrite the page index should survive so downstream page-level pruning still works:
   `ColumnIndex` (which holds no file offsets) copies verbatim; `OffsetIndex`
   page-location offsets are rebased by the same delta as the chunk data. Drop the index
   cleanly when the source has none, or when a remote read only fetched the
   column-chunk + footer ranges and not the index blocks. Gated on writer-side
   `OffsetIndex`/`ColumnIndex` support, which is deferred past #9.

5. **Multi-file merge.** N input files → 1 output, concatenating row groups and
   reconciling a shared schema — the compaction case.

6. **Operation spectrum + fallback rule.** A clear predicate for "can this op be served
   by byte-copy, or must it decode?" so the fast path is taken whenever it is legal and
   the slow path is the correct-by-construction fallback.

## API sketch (to be settled in the design)

Roughly a `ParquetRewriter` over `InputFile`(s) → `OutputFile`, taking a spec of
columns to keep, an optional row-group predicate, and a target codec — the fast path
engaging automatically when the requested transform permits it.

## Prior art

zpq (Zig) implements exactly the byte-copy variant: `src/core/writer/fastpath.zig`
copies surviving row groups with a single `memcpy` and rebases footer offsets;
`reserializeOffsetIndexShifted` / `reserializeColumnIndex` carry the page index forward
(the concrete recipe for item 4). Bloom-filter offsets are dropped there — we can
choose to carry or drop.

## Sequencing

- Depends on the core writer (#9).
- The byte-copy fast path (2/3) and page-index carry-forward (4) additionally depend on
  writer-side index-structure support (post-#9).
- Warrants its own design doc under `_designs/` before implementation.

arnabnandy7:
## Summary

Adds the first runtime implementation of Parquet row-group rewriting and compaction.

The new `ParquetRewriter` can:

- Rewrite a single Parquet file without decoding column data
- Merge schema-identical files while preserving row-group order
- Copy encoded column chunks in bounded 8 MiB ranges
- Rebase dictionary-page and data-page offsets
- Reconstruct the destination footer
- Preserve common file key/value metadata and column ordering
- Remove stale page-index and bloom-filter offsets
- Discard incomplete output when rewriting fails

Also adds the implementation design under `_designs/PARQUET_REWRITE.md`.

Fixes #791

## Testing

- `./mvnw.cmd -pl core -Dtest=ParquetRewriterTest test`
  - 4 tests passed
- `./mvnw.cmd -pl core verify`
  - Full core verification passed
- `git diff --check`
  - Passed

Test coverage includes:

- Single-file byte-copy rewriting
- Multi-file compaction and row ordering
- Encoded chunk byte equality
- Dictionary and data-page offset rebasing
- Schema mismatch validation before output creation

## Checklist

- [ ] Build via `./mvnw clean verify` passes *(full multi-module clean build was not run)*
- [x] The commit message is prefixed with the issue key (`#791 Add Parquet row-group rewrite`)

### DIFF
--- a/_designs/PARQUET_REWRITE.md
+++ b/_designs/PARQUET_REWRITE.md
@@ -1,0 +1,384 @@
+# Parquet rewrite capability (#791)
+
+**Status: Proposed.** Tracking issue: #791. Depends on writer support (#9).
+
+## Context
+
+Hardwood can read an existing Parquet file and, once #9 is complete, write values to a
+new one. Table-maintenance operations need a layer between those APIs: select data from
+one or more existing files and produce a new file while preserving encoded bytes whenever
+the requested transform does not change them.
+
+The rewrite path covers two materially different operations:
+
+- **Decode and re-encode** is the general fallback. It supports row-level filtering,
+  computed or converted values, and changes to compression or encoding.
+- **Byte copy** is the fast path. A complete surviving column chunk is copied without
+  decompression or decoding, then its file-relative metadata is rebased to its new
+  position.
+
+The two paths may be mixed in one output file. Planning is per row group and column
+chunk rather than per rewrite: one row group may be dropped, another copied, and another
+decoded because only some of its rows survive.
+
+This is a forward-only operation. Source files are random-access [InputFile] instances;
+the destination is a sequential [OutputFile]. Data and auxiliary index blocks are emitted
+before a newly serialized footer, using the same footer-last layout as `ParquetFileWriter`.
+
+## Goals
+
+- Rewrite one or more input files to one output file.
+- Drop complete row groups using a row-group predicate without reading their pages.
+- Apply an exact row predicate with decode/re-encode fallback when a row group only
+  partially matches.
+- Project complete leaf columns without decoding retained column chunks.
+- Preserve encoded column-chunk bytes when schema, codec, encoding, and values are
+  unchanged.
+- Preserve supported file, row-group, and column metadata deliberately rather than
+  silently losing fields.
+- Carry forward `ColumnIndex` and rebase `OffsetIndex` on the byte-copy path.
+- Bound memory and individual reads independently of input and column-chunk size.
+- Make the chosen strategies observable and directly testable.
+
+## Non-goals
+
+- In-place modification of an input file.
+- Combining pages from different source column chunks into one destination chunk.
+- Repartitioning or sorting rows.
+- Schema evolution beyond column projection in the first implementation.
+- Copying encrypted column chunks before Hardwood models and writes Parquet encryption
+  metadata.
+- Treating a metadata field Hardwood cannot serialize as if it had been preserved.
+
+## Public API
+
+The rewrite API is separate from `ParquetFileWriter`: the ordinary writer owns page and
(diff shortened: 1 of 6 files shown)


gunnarmorling:
I think this one will have to wait a bit until the writer has matured more, most importantly, covering all physical and logical types. Working on this atm.

gunnarmorling:
Hey @arnabnandy7, thanks a lot. I think more work needs to go into the actual writer itself and its API needs to be stabilized before we can get to this one, so I'm gonna close this one for the time being. I suggest we discuss it again after the initial writer delivery plan has fully shipped. Thanks!

arnabnandy7:
> Hey @arnabnandy7, thanks a lot. I think more work needs to go into the actual writer itself and its API needs to be stabilized before we can get to this one, so I'm gonna close this one for the time being. I suggest we discuss it again after the initial writer delivery plan has fully shipped. Thanks!

Sure, once decided we can work on this again. @gunnarmorling

PR #793, opened 2026-07-16
#767 Add zero-decode Parquet merge command

### LINKED ISSUE #767: CLI: merge/stitch multiple Parquet files into one without re-encoding
## Motivation

Hardwood can read many files and write new ones, but there is no fast path to combine several same-schema Parquet files into a single file. The naive approach (read all rows, re-encode, re-write) throws away work: the source files already contain encoded, compressed pages that are byte-for-byte reusable in the output.

A merge/stitch command would let users concatenate a set of Parquet files (or the row groups within them) into one output file by **copying encoded column-chunk bytes** and rebuilding only the footer — no decode, no re-encode, no re-compression.

## Proposed scope

A CLI command, e.g. `hardwood merge <out.parquet> <in1.parquet> <in2.parquet> ...` (exact name TBD — `merge` / `stitch` / `concat`), that:

- Validates all inputs share an identical schema (fail early on mismatch).
- Preserves each input's row groups as-is (or, as a later refinement, re-packs to a target row-group size).
- Streams the encoded pages of each row group into the output via zero-decode byte copy.
- Rebuilds the footer, patching per-column-chunk `dictionary_page_offset`, `data_page_offset`, and the row group `file_offset` by the destination offset of each copied region.
- Sums `num_rows` and concatenates the row-group list.

## Reference technique

Earnix's [parquet-sort](https://github.com/Earnix/parquet-sort) does exactly this offset-patching-on-copy in its final assembly step (`RowGrpWriters.finish` / `buildNewFooterMetadata` / `copyTmpFilesIntoOutFile`): it writes each row group as its own single-RG Parquet file and then concatenates the encoded page bytes with `FileChannel.transferFrom`, bumping each chunk's dictionary/data page offsets and the row-group `file_offset` by the target offset before writing the merged footer. The offset math there is a good blueprint for the metadata-rebuild part of this feature.

## Open questions

- Statistics/column-index/offset-index handling: copy through vs. recompute vs. drop.
- Row-group re-packing (merge small groups / split large ones) — in scope for v1, or a follow-up?
- Encoding/compression-codec compatibility checks beyond schema equality (e.g. differing codecs across inputs are fine since chunks are self-describing, but worth confirming).
- Where it lives: `hardwood-cli` command wrapping a reusable helper in `core` (`internal`), so `dive` or library users could reuse the stitch logic.

arnabnandy7:
Reuse encoded input bodies and rebuild relocated footer metadata so same-schema files can be combined without the cost or fidelity loss of decoding and rewriting every row.

Fixes #767 

## Test validation

- Verified merged files preserve input bodies byte-for-byte without re-encoding.
- Verified row counts, row-group order, and values across multiple inputs.
- Verified mismatched schemas are rejected before output creation.
- Verified using the output file as an input is rejected.
- Verified the CLI produces a readable merged Parquet file.
- Focused core and CLI tests pass.
- The full core test suite passes.
- Full `./mvnw clean verify` is blocked because Docker is unavailable for the existing S3 Testcontainers tests.

## Checklist

- [ ] Build via `./mvnw clean verify` passes — blocked by unavailable Docker; existing S3 Testcontainers tests could not start
- [x] The commit message is prefixed with the issue key (`#767 Add zero-decode Parquet merge command`)
- [x] User-facing behavior is documented under `docs/`

### DIFF
--- a/_designs/PARQUET_FILE_STITCHING.md
+++ b/_designs/PARQUET_FILE_STITCHING.md
@@ -1,0 +1,51 @@
+# Parquet File Stitching
+
+**Status: completed**
+
+## Purpose
+
+The `hardwood merge` command combines same-schema Parquet files without decoding,
+re-encoding, or recompressing their pages. Each input row group remains unchanged.
+
+## File assembly
+
+For every input, the stitcher copies the complete file body between the leading
+`PAR1` magic and the footer. Copying the whole body preserves column chunks, page
+indexes, bloom filters, padding, and their relative layout. One relocation delta,
+the destination body offset minus the source body offset, is applied to every
+absolute offset from that input's footer.
+
+The output consists of one leading magic, the copied input bodies in argument
+order, and a newly serialized footer. Its row-group list is the concatenation of
+the input row-group lists and its row count is their checked sum.
+
+## Validation
+
+All inputs are opened and validated before the output is created. Inputs must have
+identical schema elements and column orders. File-level key/value metadata must be
+identical because choosing one input's conflicting value would silently change the
+meaning of another input. Encrypted files are rejected by the existing metadata
+reader.
+
+Offsets and row counts use checked arithmetic. A failure discards the temporary
+output through `OutputFile`, so a partial Parquet file is never published.
+
+## Metadata
+
+The rebuilt footer preserves file and column key/value metadata, statistics,
+column orders, page-index locations, and bloom-filter locations. Page encodings
+and compression codecs may differ between row groups because Parquet describes
+them per column chunk.
+
+`ColumnChunk.file_offset` is written as zero. That field is deprecated and zero is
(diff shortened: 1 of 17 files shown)


gunnarmorling:
Hey @arnabnandy7, thanks a lot. I think more work needs to go into the actual writer itself and its API needs to be stabilized before we can get to this one, so I'm gonna close this one for the time being. I suggest we discuss it again after the initial writer delivery plan has fully shipped. Thanks!

arnabnandy7:
> Hey @arnabnandy7, thanks a lot. I think more work needs to go into the actual writer itself and its API needs to be stabilized before we can get to this one, so I'm gonna close this one for the time being. I suggest we discuss it again after the initial writer delivery plan has fully shipped. Thanks!

Sure thanks for letting me know.

Issue #728, opened 2026-06-29
Intra-page skip-decode: drive the ColumnIndex row mask into the decoder to skip decoding masked-out runs


gunnarmorling:
## Context

ColumnIndex/OffsetIndex page-level push-down (#118) is implemented and wired: `PageFilterEvaluator.computeMatchingRows` builds a `RowRanges` of surviving rows, `RowRanges.maskForPage` derives a per-page `PageRowMask` of surviving `[start, end)` intervals, and `IndexedFetchPlan` already drops whole non-matching pages at the byte level.

Within a **surviving** page, however, the whole page is decoded and the mask is applied only at assembly time. `ColumnWorker.decode` calls `PageDecoder.decodePage(...)` on the full page; `FlatColumnWorker.assemblePage` → `copyPageRange` then copies only the masked intervals into the output batch. Non-matching row ranges inside a surviving page are **decoded, then discarded**.

## The gap

When a selective ColumnIndex filter leaves only a few sub-page ranges alive, most of the page's value/level decode + dictionary apply is wasted work. Page-granular skipping is done (`IndexedFetchPlan`); **sub-page (intra-page) skip-*decode* is not.** Decode-then-discard is pure CPU waste here.

## Proposal

Thread the `PageRowMask` into the decode path so the decoder can skip runs that fall entirely within masked-out ranges, rather than decoding them and discarding at assembly:

- Constant RLE runs inside a masked region → advance the run cursor by `length`, scatter nothing.
- Bit-packed runs straddling a mask boundary → unpack only the kept sub-range.

This is a natural consumer of the #726 run cursor (constant/bit-packed run iteration). Without the run cursor it degrades to per-element skipping, which buys little — so #726 is effectively a prerequisite.

## Relationships

- **Builds on #726** (run cursor is the enabler).
- **Extends #118 / `IndexedFetchPlan`** from page-granular page-drop to sub-page decode-skip.
- **Sibling of #381** (page-level skip for `firstRow` seeking) — same idea, different skip source (`firstRow` residue vs. ColumnIndex mask); the residue/first-kept-row plumbing may be shareable.
- **Distinct from #500** (cross-column late materialization): there the skip source is a runtime predicate on the *predicate* columns gating *payload* decode; here the skip source is the static ColumnIndex `PageRowMask`. They compose.
- `MaskCapability` already gates masking off for nested-v1-without-OffsetIndex; reuse that gate.

## Constraints / caveats

[...]

Issue #776, opened 2026-07-10
Don't trust column statistics for pushdown on columns with an unrecognized logical type


rionmonster:
## Problem

An unrecognized logical type and an *absent* logical type are indistinguishable in our schema model: `LogicalTypeReader` returns `null` for an unknown union member (the correct "treat as absent" behavior, see #739), and a plain unannotated column is also `null`. Because they collapse to the same `logicalType() == null`, statistics-based pruning cannot tell them apart.

Pruning then picks its comparator from the *physical* type alone, with no logical gating — e.g. a `BinaryColumnPredicate` resolves to an unsigned comparison in `FilterPredicateResolver`, and `StatisticsFilterSupport` compares the file's min/max on that basis. If the writer computed min/max under the unknown type's own type-defined order (which need not be unsigned-byte for `BYTE_ARRAY`, or may be signed for a `FIXED_LEN_BYTE_ARRAY`), Hardwood can prune a row group or page that actually holds matching rows — a silent wrong result.

Per the parquet dev@ thread, unknown logical types should **not** surface statistics, since interpreting min/max is itself a semantic operation.

## How it surfaced

Follow-up from #739 and the [dev@ discussion](https://lists.apache.org/thread/flfzz94ftdrdop9d5b0o1hkqkprzj3l5) (PARQUET-3633). Micah Kornfield:

> The main other thing that unknown logical types should do is not surface statistics (as that is a form of semantic interpretation).

The trigger is narrow (a filter on an unknown-logical-type column whose stats were written under a non-physical order), but it is a genuine correctness gap, not test-only work.

## Proposed fix

- Carry "unknown annotation" distinctly from "absent annotation" through the read path (e.g. an internal sentinel from `LogicalTypeReader` → `SchemaElement` → `ColumnSchema`), without changing the #739 contract: value conversion still passes the raw physical value through, and the physical type is still exposed.
- In row-group and page (Column Index) statistics evaluation, skip min/max pruning for a column whose logical type is unrecognized. Bloom filters, being exact-match on raw bytes, can be assessed separately.
- Leave plain unannotated columns (e.g. UTF-8 strings) using statistics unchanged — no pruning regression for the common case.

## Acceptance criteria

- A file with an unknown-logical-type column that carries min/max statistics, read with a filter on that column, returns every matching row (no wrongful row-group/page drop).
- A regression test pins the above and confirms unannotated columns still prune.

## Related

- Verified-graceful reader behavior: #739.
[...]

PR #796, opened 2026-07-17
#795 Skip per-row filter evaluation when statistics prove a row group matches in full

### LINKED ISSUE #795: Always-match statistics decision: skip per-row filter evaluation when stats prove a row group or page matches in full
## Motivation

Statistics evaluation currently answers only one question — *can this row group / page be dropped?* (`StatisticsFilterSupport.canDropLeaf`, `RowGroupFilterEvaluator.canDropRowGroup`, `PageDropPredicates.canDropPage`, and the ColumnIndex path in `PageFilterEvaluator`). Everything that survives goes through exact filtering, where the compiled matcher (`RowMatcher` on the row path, `CompiledBatchFilter` on the columnar drain) runs over **every** row.

For the most common selective-scan shape — a range predicate on a sorted or clustered column (timestamps, ingestion ids) — the statistics of most *surviving* units don't merely fail to exclude them; they prove **every** row matches: the whole `[min, max]` satisfies the predicate and `null_count == 0`. All of those rows are then evaluated one by one to reconfirm a foregone conclusion.

## Proposal

Make the statistics decision three-valued:

- `CANNOT_MATCH` — today's `canDrop == true`
- `MIGHT_MATCH` — today's `canDrop == false`
- `ALWAYS_MATCHES` — new

Leaf rule: `ALWAYS_MATCHES` when min and max are present, every value in `[min, max]` satisfies the operator (`gt(x)`: `min > x`; `eq(x)`: `min == max == x`; etc.), and `null_count == 0`. For `isNotNull`, `null_count == 0` alone suffices. Composition: `AND` — any `CANNOT_MATCH` wins, else all-`ALWAYS_MATCHES` wins, else `MIGHT_MATCH`; `OR` is the dual. Bloom filters never upgrade a decision — membership structures prove absence only.

Two consumers, in order:

1. **Row-group level**: an `ALWAYS_MATCHES` row group is read with filtering disabled — the matcher never runs, all rows are kept.
2. **Page level**: pages whose ColumnIndex entry evaluates to `ALWAYS_MATCHES` contribute their rows as *pre-approved* ranges — kept, but exempt from matcher evaluation. This is the complement of #728: there, masked-out ranges skip *decode*; here, fully-matching ranges skip *evaluation*. They compose on the same `RowRanges` plumbing.

## Correctness constraints

- **Truncated bounds are safe by construction**: inexact min/max (`is_min/max_value_exact = false`, #483) only *widen* `[min, max]`, and "all of the wider interval matches" implies the actual values match. No new dependency on the exactness flags.
- **Floating point**: NaN sits outside the min/max order and `nan_count` is not yet consumed (#607) — FLOAT / DOUBLE / FLOAT16 columns must never yield `ALWAYS_MATCHES` for value predicates until it is (`isNotNull` with `null_count == 0` remains fine). The signed-zero widening from the ColumnOrder handling (#595) applies to the widened bounds unchanged.
- **Unknown logical types**: same rule as #776 is adopting for drops — no statistics decision at all, so never `ALWAYS_MATCHES`.
- **`NOT`**: negating `CANNOT_MATCH` yields `ALWAYS_MATCHES` only when the child decision is null-aware (a drop derived while `null_count > 0` does not negate). Conservative rule: `MIGHT_MATCH` unless `null_count == 0`.
- Missing or partial statistics → `MIGHT_MATCH`; behavior identical to today's.

## Relationships

- **#791** item 6 asks for "a clear predicate for whether an op can be served by byte-copy": the row-group-level `ALWAYS_MATCHES` decision is exactly that predicate for the filtering case. This issue supplies the trigger; the rewrite consumes it.
- **#728 / #726**: complementary, as above — skip-decode of non-matching ranges vs. skip-evaluation of fully-matching ranges.
- **#205 / #206 / #776**: the statistics-trust discipline the new decision inherits from `canDrop`.

Prior art: zpq implements this as [`Decision.always_match`](https://github.com/chicagobuss/zpq/blob/v0.3.0/src/core/filter/prune.zig#L17-L43) — the [leaf rule](https://github.com/chicagobuss/zpq/blob/v0.3.0/src/core/filter/prune.zig#L136-L142) is whole-`[min, max]` satisfaction (per-type interval checks in [`rangeAlwaysMatches`](https://github.com/chicagobuss/zpq/blob/v0.3.0/src/core/filter/encoded.zig#L57-L59)) plus `null_count == 0` — and uses it both to skip row filtering and as the trigger for the byte-copy passthrough already described in #791's prior-art note.

## Validation

- Unit matrix mirroring the existing `canDrop` tests: operator × type × (min, max, null_count) combinations asserting the tri-state.
- Property test over high-cardinality generated data: whenever the evaluator returns `ALWAYS_MATCHES` for a unit, brute-force evaluation over the decoded values finds zero non-matching rows. (Low-cardinality fixtures pass broken decisions silently.)
- Filter-on vs. filter-off row-set identity across the existing differential matrix.
- A range-filter-on-clustered-column benchmark (taxi pickup timestamps) showing matcher work on surviving units dropping to ~0, neutral elsewhere.

Happy to take this on. Planned as two PRs: the tri-state decision + the row-group-level consumer first; the page-level pre-approved ranges as a follow-up.

chicagobuss:
## Summary

Filtered reads over sorted or clustered data should run at unfiltered-scan speed when row-group statistics already decide the filter's outcome.  That's what this PR enables.

The common shape is a time-range scan over data laid down in timestamp order: most surviving row groups don't just *survive* the min/max check - their statistics prove **every** row matches. Hardwood still evaluated the predicate row by row across those groups, re-deriving a foregone conclusion; on a 30M-row string-predicate read that per-row work was nearly half the total read time (565 -> 294 ms once skipped, within ~3% of the unfiltered scan). This PR teaches the reader to recognize the proven-full-match case and skip per-row evaluation for it, while leaving every undecided case exactly as it is today.

### Mechanism

Implements the row-group half of #795: a three-valued `StatsDecision` (`CANNOT_MATCH` / `MIGHT_MATCH` / `ALWAYS_MATCHES`) replacing the boolean drop as the single statistics evaluation (`canDropRowGroup` is now defined as `decideRowGroup(...) == CANNOT_MATCH`), plus two consumers:

- **Wholesale**: when every surviving row group is proven to match in full, the read is built as an unfiltered read - no matcher compilation, no `FilteredRowReader` wrapper.
- **Per-batch (drain-side)**: the drain flushes batches at row-group boundaries where the always-matches flag changes - the same mechanism as the existing file-boundary flush, applied uniformly across all workers of a projection so batches stay row-aligned - and fully-matching batches emit the all-ones mask without running the `ColumnBatchMatcher`.

The correctness gates: `ALWAYS_MATCHES` requires whole-`[min, max]` satisfaction *and* `null_count == 0`; floating-point value leaves never promise a full match while `nan_count` is unread (#607); `IS NULL` never promises one (null counts tally values, not rows); bloom filters only ever force `CANNOT_MATCH`; deprecated min/max fields are already nulled by MinMaxStats.of`. Truncated bounds are safe by construction - they only widen the interval. Design notes in `_designs/ALWAYS_MATCH_STATISTICS.md`.

Page-level pre-approved ranges and the `ColumnReader`/`SelectionEngine` fast path are the planned follow-up (they compose with #728), and the row-group decision is the byte-copy trigger #791 item 6 calls for.

Prior art: this is the shape zpq ships as [`Decision.always_match`](https://github.com/chicagobuss/zpq/blob/v0.3.0/src/core/filter/prune.zig#L17-L43)
- the same [leaf rule](https://github.com/chicagobuss/zpq/blob/v0.3.0/src/core/filter/prune.zig#L136-L142) (whole-`[min, max]` satisfaction via [`rangeAlwaysMatches`](https://github.com/chicagobuss/zpq/blob/v0.3.0/src/core/filter/encoded.zig#L57-L59), gated on `null_count == 0`), introduced there for the byte-copy passthrough in [`1f29c0e`](https://github.com/chicagobuss/zpq/commit/1f29c0e65631200ea757381d2ec0a20669099d9d).

## Testing

- `./mvnw verify` (full multi-module, JDK 25): green, 1719 core tests.
- `StatsDecisionTest` — combinator truth tables, per-operator/type leaf matrix, null-count gating, FP never-always, plus a seeded 10,000-round property test:  whenever the evaluator returns `ALWAYS_MATCHES` over random data, brute-force evaluation finds zero non-matching rows (and zero matching for `CANNOT_MATCH`).
- `RowGroupDecideTest` - row-group-level decisions, `AND`/`OR` composition, null predicates, deprecated-stats and missing-stats conservatism, `canDropRowGroup` equivalence.
- `AlwaysMatchingRowGroupTest` — end-to-end row-set correctness over `filter_pushdown_int.parquet` (threshold sweep crossing every full/partial/dropped row-group configuration, asserting exact row sequences) and a multi-file read mixing fully-matching and dropped row groups.

## Performance

Sorted 30M-row file (30 row groups of 1M, uncompressed, local mmap), best of 9, `RowReader` summing a payload column; baseline is `24607ee`:

| scenario | baseline | this PR | |
|---|---|---|---|
| string `gtEq`, all row groups fully match | 565 ms | 294 ms | **1.9x** - within ~3% of the 285 ms unfiltered scan |
| string `gtEq`, 15 of 30 groups match (all fully) | 267 ms | 149 ms | **1.8x** |
| int `gtEq`, all row groups fully match | 201 ms | 194 ms | ~3% (see below) |
| unfiltered scan (control) | 304 ms | 285 ms | noise band |

The record-matcher path (string/binary predicates via `FilteredRowReader`) is where the win lives - per-row evaluation disappears. The drain-side int path barely moves because its vectorized matchers were already a small fraction of the read; for that path the value of this PR is the mask-skip plumbing plus the `Batch.filterAlwaysMatches` proof that the follow-up `SelectionEngine` fast path consumes. Row counts and sums are byte-identical to baseline in every scenario.

Fixes #795

## Checklist

- [x] The commit message is prefixed with the issue key (`#795 Add always-match statistics decision to skip per-row filtering`)
- [x] If this PR adds or changes user-facing behaviour, `docs/` has been updated *(no public-API change — everything stays in `internal.*`; the new JFR field is additive)*

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -345,6 +345,7 @@ For field-level `parquet.thrift` metadata coverage (which spec fields are read/p
 - [x] Drain-side per-batch record filtering (`BatchFilterCompiler` + `ColumnBatchMatcher`, on by default for column-local AND queries; see `_designs/DRAIN_SIDE_RECORD_FILTERING.md`)
 - [x] Exact column-reader filtering — `buildColumnReader(...).filter(...)` / `buildColumnReaders(...).filter(...)` return only matching rows with no client-side residual (`SelectionEngine` + `FilterCoordinator`; see `_designs/EXACT_COLUMN_READER_FILTERING.md`)
 - [x] Bloom filter-based row group filtering: `eq` on INT32, INT64, FLOAT, DOUBLE, and binary columns, and `in` on the integer and binary types (`RowGroupBloomFilterSource`; see `_designs/BLOOM_FILTER_PUSHDOWN.md`)
+- [x] Always-match statistics decision: tri-state `StatsDecision` proves from statistics when every row in a row group matches, skipping per-row filter evaluation for that group and dropping the filter wholesale when all surviving groups fully match (see `_designs/ALWAYS_MATCH_STATISTICS.md`)
 
 ### 9.5 Fixed-size-list read fast path
 - [x] Detect fixed-width fixed-*k* `LIST` pages from level streams alone (`FixedSizeListDetector`): O(1) definition-level gate + O(rows) repetition verification
--- a/_designs/ALWAYS_MATCH_STATISTICS.md
+++ b/_designs/ALWAYS_MATCH_STATISTICS.md
@@ -1,0 +1,100 @@
+# Design: always-match statistics decisions (#795)
+
+**Status: Implemented** (row-group granularity; page-level pre-approved ranges are a
+follow-up, see "Boundaries" below).
+
+## Problem
+
+Statistics evaluation is drop-only: `StatisticsFilterSupport.canDropLeaf`,
+`RowGroupFilterEvaluator.canDropRowGroup`, and `PageDropPredicates.canDropPage` answer
+only "can no rows match?". Every surviving unit is then evaluated row by row — the
+compiled `RowMatcher` on the record path, per-column `ColumnBatchMatcher`s on the
+drain-side path.
+
+For a range predicate over a sorted or clustered column, the statistics of most
+*surviving* row groups prove the opposite extreme: every row matches (the whole
+`[min, max]` interval satisfies the predicate and the null count is zero). Per-row
+evaluation over those groups reproduces a foregone conclusion at full decode-side cost.
+
+## The decision model
+
+`StatsDecision` (internal.predicate) is the three-valued extension of the boolean drop:
+
+| decision         | meaning                             | today's equivalent |
+|------------------|-------------------------------------|--------------------|
+| `CANNOT_MATCH`   | no row matches; skip the unit       | `canDrop == true`  |
+| `MIGHT_MATCH`    | undecided; evaluate rows            | `canDrop == false` |
+| `ALWAYS_MATCHES` | every row matches; skip evaluation  | — (new)            |
+
+`canDropRowGroup` is now defined as `decideRowGroup(...) == CANNOT_MATCH`; there is one
+evaluation implementation, not two.
+
+### Leaf rule
+
+A value leaf is `ALWAYS_MATCHES` when all of:
+
+- min and max are present and trusted (deprecated unsigned min/max fields are already
+  nulled out by `MinMaxStats.of`), and
+- every value in `[min, max]` satisfies the operator
+  (`StatisticsFilterSupport.alwaysMatches` / `alwaysMatchesCompared`; for `IN`,
+  the single-point case `min == max ∈ values`), and
+- `null_count == 0` — a null row satisfies no value predicate.
+
+`IS NOT NULL` is `ALWAYS_MATCHES` on `null_count == 0` alone. Everything else that
+cannot be proven is `MIGHT_MATCH`, which preserves today's behavior exactly.
+
+Truncated (inexact) bounds need no special handling: truncation only widens
+`[min, max]`, and a predicate satisfied by the widened interval is satisfied by the
+actual values.
+
+### Deliberate non-promises
+
+- **Floating point** (`FLOAT`, `DOUBLE`, `FLOAT16`): NaN sits outside the min/max
+  ordering and `nan_count` is not consumed (#607), so a fully-satisfying interval may
+  still hide non-matching NaN rows. FP value leaves never yield `ALWAYS_MATCHES`.
+- **`IS NULL`**: the null count tallies leaf values, not rows, so `null_count ==
+  numRows` does not prove every row's leaf is null for nested columns.
+- **Bloom filters** prove absence only; they can force `CANNOT_MATCH` but never upgrade
+  a decision.
+- **Composition**: `AND` is `ALWAYS_MATCHES` only when every child is; `OR` when any
+  child is; empty composites stay `MIGHT_MATCH`.
+
+## Consumption (row-group granularity)
+
+`RowGroupIterator.filterRowGroups` evaluates the tri-state per row group: `CANNOT_MATCH`
+groups are dropped (unchanged), and each surviving `WorkItem` carries
+`filterAlwaysMatches`.
+
+Two consumers:
+
+1. **Wholesale**: when every work-list row group is `ALWAYS_MATCHES`
+   (`RowGroupIterator.isFilterSatisfiedByStatistics`), `FlatRowReader.create` /
+   `NestedRowReader.create` treat the read as unfiltered — no matcher compilation, no
+   `FilteredRowReader` wrapper, and `maxRows` caps scanned rows directly since
+   scanned == matching.
+2. **Per-batch, drain-side**: batches must be homogeneous for a batch-level skip, so
+   the drain flushes the current batch when a page crosses a row-group boundary that
+   changes the flag — the same mechanism as the existing file-boundary flush, and
+   applied **uniformly across all workers of a projection** so batches stay
+   row-aligned (a per-column flush would desynchronize the merge). For an
+   always-matching batch the worker skips `ColumnBatchMatcher.test` and writes the
+   all-ones mask a matcher would have produced; `Batch.filterAlwaysMatches` records
+   the proof for downstream consumers.
+
+The flag travels retriever → drain through a per-slot buffer written alongside
+`fileNameBuffer[slot]`, under the same happens-before chain.
+
+`RowGroupFilterEvent` gains `rowGroupsFullyMatching` so the effect is observable in JFR.
+
+## Boundaries
+
+- **Page-level pre-approved ranges** (skip evaluation inside surviving pages whose
+  ColumnIndex entry is `ALWAYS_MATCHES`) extend `RowRanges` and compose with the
+  intra-page skip-decode work (#728); follow-up to #795.
+- **`ColumnReader` / `SelectionEngine`**: `computeSelection` already has an every-record
+  fast path (`-1`); feeding it from `Batch.filterAlwaysMatches` is part of the same
+  follow-up.
+- **Record-matcher path** (`FilteredRowReader`): benefits from the wholesale case only;
+  the wrapper has no row-group visibility for per-group skips.
+- The row-group decision is exactly the trigger condition the rewrite capability's
+  byte-copy fast path needs (#791, item 6).
--- a/core/src/main/java/dev/hardwood/internal/predicate/MinMaxStats.java
+++ b/core/src/main/java/dev/hardwood/internal/predicate/MinMaxStats.java
@@ -7,6 +7,8 @@
  */
 package dev.hardwood.internal.predicate;
 
+import java.util.List;
+
 import dev.hardwood.metadata.ColumnIndex;
 import dev.hardwood.metadata.Statistics;
 
@@ -19,6 +21,9 @@ interface MinMaxStats {
 
     /// @return the maximum value as raw bytes, or `null` if absent
     byte[] maxValue();
+
+    /// @return the number of null values in the unit, or `null` if unknown
+    Long nullCount();
 
     /// Wraps a [Statistics] record.
     ///
@@ -39,6 +44,11 @@ interface MinMaxStats {
             public byte[] maxValue() {
                 return deprecated ? null : stats.maxValue();
             }
+
+            @Override
+            public Long nullCount() {
+                return stats.nullCount();
+            }
         };
     }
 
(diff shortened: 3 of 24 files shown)


chicagobuss:
One design point worth making explicit for review:  this PR asks the reader to trust statistics for something specific, and why I chose the default-on choice.

Pruning already trusts statistics for correctness - a lying `min`/`max` silently drops matching rows today. `ALWAYS_MATCHES` adds the mirror image: a lying bound can *include* non-matching rows. Same trust class, opposite failure direction; no new category of trust is introduced. And the new decision inherits every per-case gate the drop side has accumulated, since both go through `MinMaxStats.of`: deprecated unsigned min/max fields are nulled out (#205), unknown logical types are excluded once #776 lands, column-order ambiguity is handled per #595, and FP columns are withheld entirely until `nan_count` is consumed (#607).

For context on how seriously to take lying stats: zpq treats statistics trust as tiered rather than boolean, because its test corpus contains a real liar - a `parquet-mr 1.8.2` file whose recorded `min` is `2.00` for a column that contains `1.00`. Its defaults trust footer-level row-group stats (including for the byte-copy passthrough), gate the *page-level* always-match behind an explicit `--trust-stats` opt-in - page-index stats have their own writer-bug surface - and offer `--scan-all`
to distrust everything. Notably, that 1.8.2 file falls exactly in the deprecated-fields class this codebase already refuses, which is a nice validation of the per-case-gate strategy over a global knob.

The tradeoff carried forward: when the page-level follow-up lands, it's worth deciding whether ColumnIndex-derived `ALWAYS_MATCHES` deserves stricter treatment than the footer-derived one, and whether a builder-level opt-out (evaluate rows even when stats promise a full match) is wanted as a diagnostic escape hatch. Happy to add either there if you want it.

gunnarmorling:
Hey @chicagobuss, thank you so much, this is awesome! We're gonna look into this asap. Tagging @iifawzi for awareness, he was working a lot on filter evaluation.

iifawzi:
Hey @chicagobuss thank you so much; that looks very cool! I've taken a very quick look, but willing to take another complete look tonight

iifawzi:
> For context on how seriously to take lying stats: zpq treats statistics trust as tiered rather than boolean, because its test corpus contains a real liar - a parquet-mr 1.8.2 file whose recorded min is 2.00 for a column that contains 1.00. Its defaults trust footer-level row-group stats (including for the byte-copy passthrough), gate the page-level always-match behind an explicit --trust-stats opt-in - page-index stats have their own writer-bug surface - and offer --scan-all
to distrust everything. Notably, that 1.8.2 file falls exactly in the deprecated-fields class this codebase already refuses, which is a nice validation of the per-case-gate strategy over a global knob.

On the other hand, I found this quite interesting too: are there known bugs in writers that make this necessary, or is this more of a defensive measure to give zpq users the option not to trust statistics? Also, I'm curious whether the fixture with the incorrect statistics was created specifically for testing, or if it's based on a real-world case you encountered

chicagobuss:
I've encountered bad statistics fairly often when working with vendor-provided parquet data from a particularly well-known hyperscaler's cloud billing data... 😅 

`zpq` has a middle-ground default that matches what I've personally seen the most in the wild - probably accurate row counts but sometimes wrong 'higher-order stats'.  But it also allows you to go either direction fully.

That fixture file was deliberately created but only because I didn't want to use 'real' parquet files that had similar problems which might've contained traces of sensitive data

chicagobuss:
> Hey @chicagobuss, thank you so much, this is awesome! We're gonna look into this asap. Tagging @iifawzi for awareness, he was working a lot on filter evaluation.

💯  I promised I'd return the favor!  Hardwood has helped immensely in my zpq testing.  Please take a look at using/helping with zpq if/when you get the chance!

gunnarmorling:
So that question of incorrect statistics is an interesting one; I think I'm leaning towards trusting them by default, but having a reader option to disable stats. I.e. assume correctness by default but provide means of falling back to full scan filtering if needed.

chicagobuss:
For whatever it's worth, I feel like that's appropriate for Hardwood as well. (trust all stats by default)

iifawzi:
> I've encountered bad statistics fairly often when working with vendor-provided parquet data from a particularly well-known hyperscaler's cloud billing data... 😅

Oh, that's nice to know, and it is quite surprising in the same time! In that case, an option to opt out is nice to have for sure. 

On the other hand, I took a look at the PR, and it looks so good to me; thank you so much for working on it! few nits from Claude, if you can check what makes sense: 

---

## Documentation
- [ ] `ROADMAP.md` §9.4 Predicate Pushdown is not updated with the always-match
  statistics decision (#795); CLAUDE.md requires the roadmap status be updated after
  implementing a feature.
- [ ] `docs/content/reference/configuration.md:97` lists the `RowGroupFilter` JFR event
  fields but omits the newly added `rowGroupsFullyMatching`; the public `dev.hardwood.jfr`
  event changed, so its documented field list must too.

## Tests
- [ ] Cross-worker row-alignment on the mixed drain-side flush path is never asserted
  against a second projected column — every test in `AlwaysMatchingRowGroupTest` reads
  only `id`, so a desync between the filter column and a non-filter column at a
  boundary-crossing threshold would pass undetected; add a case that also reads/asserts
  `value` under a threshold mixing fully-matching and partial groups.
  → Added `nonFilterColumnStaysRowAlignedAcrossMixedGroups`: `gtEq("id", 150)` drops RG1,
  partially matches RG2, fully matches RG3, and asserts `value`/`label` stay aligned with
  `id` across the RG2→RG3 always-matches flush boundary.
- [ ] No test pins that the `ALWAYS_MATCHES` fast path actually fires (e.g. via JFR
  `rowGroupsFullyMatching`); the correctness sweeps would still pass if the optimization
  never triggered, so a regression that silently disabled it wouldn't be caught.
  → Added `RowGroupFilterEventTest#predicatePushDownReportsFullyMatchingRowGroups`:
  `gt("id", 150)` asserts the emitted event reports `rowGroupsFullyMatching == 1` (RG3),
  so a regression that stopped deriving `ALWAYS_MATCHES` would fail the test.

## Nits
- [ ] Empty-`OR` now decides `MIGHT_MATCH` where the old `canDropRowGroup` returned
  `true` (`CANNOT_MATCH`); it's intentional per the design's "empty composites stay
  `MIGHT_MATCH`" and is safe (over-keeps, never wrong), but it is a behavior shift for the
  empty-disjunction case worth being aware of.
---

Finally, question on the performance analysis: I couldn't find a matching setup on the repository; is it committed here somewhere that I might have missed? If not, can we get it included? I was curious to measure the wins on my sandbox

chicagobuss:
Thanks for the close read! 

Addressed all four: roadmap + JFR field docs, the multi-column alignment test, and the JFR assertion that ALWAYS_MATCHES actually fires.

The alignment one was worth having: it pins an invariant I ran into while building this (per-column flushes desync the drain merge), and the JFR test you suggested is exactly the right way to keep the fast path correct.

On the benchmark: happy to share it, but what shape do you want it in?

Right now it's two small standalone harnesses (one for the int/drain-side path, one for the string/record-matcher path) plus a pyarrow script that generates the 30M-row sorted fixtures.  Either of these sound good?

1. A quick gist with the harnesses + generation script
2. A proper contribution to hardwood-benchmarks 

I think I'd prefer just a gist for now if that's ok with you

gunnarmorling:
Yeah, we don't need this in hardwood-benchmarks; those are really polished (with chart generator and everything) and primarily meant for supporting publishing numbers in blog posts, announcements, etc.

We do have JMH benches in the core, see _performance-testing/micro-benchmarks_; that module is in need of some clean-up, but perhaps you can add something there?

chicagobuss:
Sounds good - added it to micro-benchmarks instead of the gist

AlwaysMatchReadBenchmark reads a sorted 10-row-group fixture at three cutoff positions (no filter / all groups fully matching / mixed) on both the record-matcher and drain-side paths, with a small pyarrow generator following the existing script conventions. The headline check is full vs noFilter: with this PR they're within noise, and a regression that stopped deriving ALWAYS_MATCHES would reopen the gap.

I also added a README cataloguing the benchmarks and their fixture recipes, and doc headers for the two classes that didn't have one.

gunnarmorling (approved):
This LGTM now. Thank you so much, really cool stuff, @chicagobuss! Leaving it to @iifawzi for a final pass and merge if there's nothing left from your end either.

gunnarmorling:
Ah, and we should log a follow-up issue for opting out of stats usage in the reader, if this isn't tracked yet.

iifawzi (approved):
Thank you @chicagobuss, this looks so good! I pushed a small commit to the benchmark so the half case (now is called `mixed` is actually exercising dropped + partial + fully matching in the same read. Also logged #797 for the stats opt-out discussion

Merging once green

Issue #799, opened 2026-07-18
Reader option to disable statistics-based filtering


chicagobuss:
Follow-up from the discussion on #796.

Statistics are trusted by default, in both directions: pruning skips row groups and pages whose min/max prove no row can match (a lying bound silently drops rows), and since #796 the always-match decision skips per-row evaluation for groups whose statistics prove every row matches (a lying bound silently includes rows). For files with unreliable statistics — encountered in the wild with vendor-produced Parquet, and represented by the deprecated-fields class already refused per #205 — the reader should offer a full-scan fallback.

Proposal: a `ReaderConfig` option (e.g. `hardwood.statistics-filtering`, default `true`) that, when set to `false`, disables all metadata-derived filter shortcuts for the read:

- row-group statistics/bloom pruning (`decideRowGroup`)
- page-index row ranges (`PageFilterEvaluator`)
- inline page-statistics drops (`SequentialFetchPlan`)
- the always-match fast path

Predicates still apply — every surviving row is evaluated against the decoded data — so results depend only on the data pages, not on footer or page-index metadata. Trust stays the default per the discussion on #796; this is the diagnostic/defensive escape hatch.

chicagobuss:
Duplicate of #797 — I filed this without spotting that @iifawzi had already tracked it (with better open questions, too). Discussion there; PR #800 implements it.

PR #800, opened 2026-07-18
#797 Add reader option to disable statistics-based filtering

### LINKED ISSUE #797: Reader option to opt out of statistics-based filtering
## Motivation

Predicate pushdown trusts row-group and page statistics for correctness. That trust is not new — a lying `min`/`max` already silently drops matching rows — but #795 added the mirror image: `ALWAYS_MATCHES` lets a lying bound *include* non-matching rows by skipping per-row evaluation entirely. Same trust class, opposite failure direction.

Bad statistics do occur in the wild. As reported by @chicagobuss on #796: vendor-produced Parquet (cloud billing exports) with unreliable higher-order stats, and a `parquet-mr` 1.8.2 file recording `min = 2.00` for a column containing `1.00`. That particular file falls in the deprecated-fields class Hardwood already refuses via `MinMaxStats.of`, which validates the per-case-gate strategy — but per-case gates only cover the failure modes we know about.

Today a user who suspects bad statistics has no recourse short of rewriting the file.

## Proposal

A `ReaderConfig` option that makes the reader ignore statistics and evaluate the predicate over every row. Default stays as-is (statistics trusted), so this is purely additive.

```java
ReaderConfig config = ReaderConfig.builder()
        .option("hardwood.trust-statistics", "false")
        .build();
```

Results must be identical with the option on or off for a file with honest statistics — that is the test oracle.

## Open questions

**1. Granularity — one knob or tiered?**

`zpq` (prior art cited by @chicagobuss on #796) treats trust as tiered rather than boolean: footer row-group stats trusted by default, page-level always-match behind an explicit `--trust-stats` opt-in (page-index stats have their own writer-bug surface), and `--scan-all` to distrust everything. Options:

- **A. Single boolean** — one knob disables all statistics-derived decisions. Simplest surface; users who distrust one layer usually distrust all.
- **B. Tiered** — separate footer-level from page-level trust, matching `zpq`. More precise, more surface area.

Worth deciding before the page-level `ALWAYS_MATCHES` follow-up to #795 lands, since that is where the page-index bug surface enters.

**2. What exactly does the opt-out disable?** These are separable and have different risk profiles:

| Mechanism | Failure if stats lie |
|---|---|
| Row-group / page **drop** (`CANNOT_MATCH`) | Matching rows silently missing |
| **`ALWAYS_MATCHES`** (skip per-row evaluation) | Non-matching rows silently included |
| Bloom filters | Matching rows silently missing |

An argument exists for disabling only `ALWAYS_MATCHES` (the newest and least battle-tested), but a user who distrusts the file's statistics presumably distrusts the drop side equally.

**3. CLI exposure?** Per `CLAUDE.md`, only options truly needed belong in the CLI. A `--no-stats` flag would be genuinely useful for diagnosing a suspect file — worth including, or leave it library-only initially?

## Acceptance criteria

- [ ] `ReaderConfig` option, default preserving today's behaviour
- [ ] Honoured by every statistics-derived decision in scope (per Q2)
- [ ] Test: a file with deliberately corrupted statistics returns wrong rows with the default and correct rows with the opt-out
- [ ] Test: identical results with the option on/off over honest-statistics fixtures
- [ ] `docs/content/reference/configuration.md` reader-options table updated
- [ ] CLI flag if Q3 resolves that way

## References

- #796 discussion — trust rationale and the `zpq` tiered model
- #795 — always-match statistics decision
- #776 — per-case gate for unrecognized logical types

chicagobuss:
Implements #797. (I'd filed #799 for the same thing without spotting the existing issue - closed as duplicate; apologies for the noise. The branch name still carries 799.)

## What it does

A new `ReaderConfig` option, following the existing string-keyed pattern:

```java
ReaderConfig config = ReaderConfig.builder()
        .option("hardwood.statistics-filtering", "false")
        .build();
```

Default `true` - trusting statistics stays the default. Set to `"false"`, a filtered read takes no metadata-derived shortcut:

- no row-group statistics/bloom pruning (`decideRowGroup` never runs)
- no page-index row ranges (`PageFilterEvaluator` skipped; index buffers aren't even fetched)
- no inline page-statistics drops (`dropLeavesByColumn` stays empty)
- no always-match fast path

Predicates still apply in full - every surviving row is evaluated against the decoded data - so filtered results depend only on the data pages, never on footer or page-index metadata.

## Positions taken on #797's open questions

**Q1 (granularity): A - single boolean.** A user who distrusts a file's statistics rarely knows *which* layer lies; one knob is the simplest correct surface. If the page-level always-match follow-up warrants a tier of its own (that's where the page-index bug surface enters), a second key can be added then without breaking this one - the string-keyed option design exists exactly for that.

**Q2 (scope): everything, bloom included.** Bloom filters are writer-produced metadata with the same failure mode as a lying `min` (matching rows silently missing), and "results depend only on the data pages" is a cleaner contract than "statistics off, bloom still on". Disabling only `ALWAYS_MATCHES` was considered and rejected: a user who distrusts inclusion-by-stats presumably distrusts drop-by-stats equally, as the issue notes.

**Q3 (CLI): not in this PR** - library-only initially, per the minimize-CLI-surface rule. Happy to add a `--no-stats` flag in a follow-up if you want it.

**Naming**: went with `hardwood.statistics-filtering` (names the mechanism being switched) over `hardwood.trust-statistics` (names the reason) - the reader still *reads* statistics into metadata objects either way; what the flag disables is filtering decisions derived from them. Easy to rename if you prefer the issue's spelling.

## Why one flag covers all four mechanisms

Everything the reader derives from writer metadata funnels through `RowGroupIterator` state: `filterRowGroups` (row-group decisions), `getSharedMetadata` (page-index ranges), and `dropLeavesByColumn` (inline page drops). One boolean there, set by `ParquetFileReader` from the resolved config before `initialize`, gates all of it - the diff is small because #795 already routed row-group evaluation through a single point.

## Tests

`StatisticsFilteringOptionTest` covers the row-reader path, the exact-filtering column-reader path, and the case the always-match fast path would normally capture (fully-matching predicate) - asserting exact row sets and, via JFR, that no `RowGroupFilter` push-down evaluation ran at all with the option set. That covers the issue's honest-statistics oracle (identical results on/off).

Not yet covered: the corrupted-statistics fixture from the acceptance criteria (wrong rows by default, correct rows with the opt-out - the strongest possible evidence the flag matters). That needs a deliberately lying fixture writer; I can add one in this PR if you want it here, or it could ride with the corrupt-fixture tooling discussion separately.

`./mvnw verify` green.


core/src/main/java/dev/hardwood/internal/reader/RowGroupIterator.java
```diff
@@ -250,6 +251,15 @@ public FileSchema openFirst() throws IOException {
         return referenceSchema;
     }
 
+    /// Disables or re-enables statistics-based filtering for this iterator. With it
+    /// disabled, a filter predicate takes no metadata-derived shortcut — no
+    /// row-group statistics/bloom pruning, no page-index or inline-page-statistics
+    /// skipping, no always-match decision — so the predicate is evaluated against
+    /// every decoded row. Enabled by default; must be set before `initialize`.
```
iifawzi: Could it be that it's better to pass to `initialize`?

core/src/test/java/dev/hardwood/StatisticsFilteringOptionTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+/// The `hardwood.statistics-filtering` reader option (#799): set to `"false"`,
+/// no row group or page is skipped from statistics, bloom filters, or page
+/// indexes, and predicates fall back to per-row evaluation — the escape hatch
+/// for files whose metadata is unreliable. Results must be identical to the
+/// default path; only the shortcut is gone.
+///
+/// `filter_pushdown_int.parquet`: 3 row groups of 100 rows — `id` 1..100,
+/// 101..200, 201..300, with statistics that would normally prune.
+class StatisticsFilteringOptionTest extends AbstractJfrRecorderTest {
```
iifawzi, line 37: I'm not sure if this could be as useful as if we have broken fixtures with invalid statistics. I mean, if the flag isn't set/working properly tests here will keep passing,

core/src/main/java/dev/hardwood/reader/ParquetFileReader.java
```diff
@@ -97,10 +99,20 @@ private ParquetFileReader(List<InputFile> inputFiles, FileMetaData firstFileMeta
     /// via [ReaderConfig] so it can be retired without breaking callers.
     private static final String FIXED_LIST_FAST_PATH_OPTION = "hardwood.fixed-list-fast-path";
 
+    /// Reader option key: set to `"false"` to disable statistics-based filtering
+    /// (enabled by default). With it disabled, no metadata-derived shortcut is
+    /// taken for a filtered read — no row-group statistics/bloom pruning, no
+    /// page-index or inline-page-statistics skipping, no always-match fast path —
+    /// and the predicate is evaluated against every decoded row, so results
+    /// depend only on the data pages. The escape hatch for files whose
+    /// statistics are wrong (#799).
```
iifawzi: Perhaps we need to mention #797, or maybe not mention the ticket at all, unless the context isn't clear

iifawzi (commented):
Thank you @chicagobuss for working on it! few comments inline, mainly I believe we need to get the broken fixtures included in this PR, one per statistics-type or maybe a single with one with invalid statistics on all of them. I'm just feeling existing testing wouldn't catch regressions of the flag behavior.

core/src/main/java/dev/hardwood/internal/reader/RowGroupIterator.java
gunnarmorling: Or rather the constructor? Immutability FTW.

gunnarmorling: Pushed to `initialize()`.

gunnarmorling:
Thanks a lot, @chicagobuss! Pushed one more commit to fix up a few lose ends.

chicagobuss:
Thanks, ya apologies, been busy with a new job so haven't had as much time to look at this lately!

gunnarmorling:
No worries at all; happy to see this land. Congrats and all the best for the new job!

PR #807, opened 2026-07-20
#740 Speed up the fixed-size-list detector's scalar rep-level walk

gunnarmorling:
Part of #740. Two optimizations to the fixed-size-list detector's **scalar fallback** — the rep-level path taken when neither bulk probe applies (list widths `9..15`, non-uniform large-`k` encoder splits, and the degenerate `k=1`) — plus the benchmark coverage that motivated and verifies them.

Before this, both scalar sub-paths walked the rep stream one element at a time. After it, no scalar sub-path has an O(rows) loop left.

### 1. Word-at-a-time boundary scan (bit-packed runs)

The width-1 rep stream packs one bit per value, so a record boundary is a `0` bit. Instead of visiting every bit, the bit-packed branch now scans a 64-bit word at a time: the boundary mask is the word's zero bits (`~word`), and `numberOfTrailingZeros` walks them in order, skipping the runs of `1`s between boundaries. Only full 64-value words take the fast path (no padding bit is read); the sub-word remainder keeps the scalar bit loop.

### 2. O(1) collapse of the RLE-of-zeros run (`k=1`)

A run of boundaries (an RLE run of `0`s) was closed one boundary at a time. Within such a run every gap is exactly `1`, so after the first boundary closes the previous record the remaining `runLen-1` gaps all imply `k==1` — verified in O(1). A list of single elements is one long run of zeros, so this is the whole detector cost for that shape.

### Measured (N300, `perfnorm`; cross-checked on Apple silicon)

Scalar case, detector's wasted-scan penalty on a failing page (`FixedSizeListFallbackBenchmark`, `k=9`):

| metric | before | after |
|--|--:|--:|
| cycles penalty | +6.40M (+8%) | +0.05M (~0) |
| instructions penalty | +31.5M | +8.3M |

Independently on Apple silicon (wall-clock): the `k=9` scalar penalty went from a cleanly-resolved +2.6% to within noise of zero. Bulk regimes (`k=4`, `k=768`) unchanged.

Detector-only for `k=1` (`FixedSizeListDecodeBenchmark.repScanOnly`): **82.2M → 0.22M instructions/op (-99.7%)**; end-to-end `columnFast[1]`: **213M → 128M instructions (-40%)**, cycles -4.7%. Controls (`columnBaseline`, `flatFloor`, `repScanOnly[4]`) flat — the effect is isolated to the detector.

### Notes

- `k=1` is a shape writers do not actually produce (a scalar stored as a `LIST`), so its win is completeness, not a workload gain. The concrete value is that the scalar path is no longer the detector's expensive case.
- Benchmark scaffolding added: `k=9` to the fallback benchmark and `k=1` to the decode benchmark (+ generator corpus), so both scalar regimes are covered.
- Two coverage follow-ups (two-level legacy lists; nullable vectors) are tracked in separate issues.

### Tests

`FixedSizeListDetectorTest` gains two cases driving the word loop directly (an all-bit-packed `k=9` page spanning two words plus a padding tail, and a shortened record whose extra boundary lands past the first word). Existing `k=1` and differential/read suites cover the RLE-of-zeros collapse.

### DIFF
--- a/core/src/main/java/dev/hardwood/internal/reader/FixedSizeListDetector.java
+++ b/core/src/main/java/dev/hardwood/internal/reader/FixedSizeListDetector.java
@@ -440,15 +440,43 @@ public final class FixedSizeListDetector {
 
             if ((header & 1) == 1) {
                 // Bit-packed run: (header >> 1) groups of 8 values, one byte each.
+                // The width-1 rep stream packs one bit per value, so a `0` bit is a
+                // record boundary. Scan boundaries a 64-bit word (64 values) at a
+                // time: the boundary mask is the word's zero bits (`~word`), which
+                // `Long.numberOfTrailingZeros` walks in order — skipping the runs of
+                // `1`s between boundaries rather than visiting every bit. Only full
+                // words (`levelsLeft >= 64`) take this path, so no padding bit is ever
+                // read; `levelsLeft`, clamped to `numValues`, is the padding-safe
+                // count. The sub-word remainder (and any run shorter than a word)
+                // falls to the scalar bit loop below.
                 int groups = (int) (header >> 1);
                 if (groups > end - pos) {
                     return FixedSizeListShape.NOT_APPLICABLE;
                 }
-                for (int g = 0; g < groups && idx < numValues; g++) {
+                int levelsLeft = Math.min(groups * 8, numValues - idx);
+                while (levelsLeft >= 64) {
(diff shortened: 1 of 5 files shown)


gunnarmorling:
Coverage follow-ups tracked separately: #808 (two-level old-list-structure LIST columns) and #809 (nullable vectors / null rows).

Issue #815, opened 2026-07-24
Honor is_min_value_exact / is_max_value_exact if we add aggregate-from-statistics pushdown


gunnarmorling:
## Background

The Parquet `Statistics` struct carries `is_min_value_exact` / `is_max_value_exact` (fields 7 / 8). They distinguish a bound that is the **actual** column extreme (`true`) from a **compact bound** that may not exist in the data (`false` / absent) — e.g. a writer storing `min_value="B"` / `max_value="C"` for an actual `"Blart Versenwald III"`, or a truncated `BYTE_ARRAY` bound.

Our `StatisticsReader` does not parse fields 7 / 8, and the predicate-pushdown code (`StatisticsFilterSupport`) does not consult exactness. **This is correct today** and there is no bug: pushdown uses `min_value` / `max_value` purely as *bounds*. Every drop / always-match decision is sound with inexact bounds, because the spec guarantees `min_value` is a valid lower bound and `max_value` a valid upper bound. The only exactness-looking shortcuts — `EQ` ⇒ ALWAYS_MATCHES and `NOT_EQ` ⇒ CANNOT_MATCH, both gated on `min == max` — stay sound too: if `min_value == max_value == P` and both are valid bounds, then `P ≤ trueMin ≤ trueMax ≤ P`, which forces every value to equal `P`.

## When this becomes load-bearing

Exactness stops being ignorable the moment we use a bound as an **actual value** rather than a pruning bound. The concrete trigger is **aggregate-from-metadata pushdown** — answering `SELECT max(col)` / `SELECT min(col)` (or feeding `min`/`max` into further computation) directly from `max_value` / `min_value` without scanning. There, a bound flagged inexact would over- / under-report the aggregate, producing a silently wrong result.

## Ask

If/when aggregate-from-statistics pushdown is implemented:

- [ ] Parse `is_min_value_exact` (field 8) and `is_max_value_exact` (field 7) in `StatisticsReader` into the `Statistics` record.
- [ ] In any path that returns `min_value` / `max_value` as an answer (not a bound), require the corresponding exact flag; fall back to a scan otherwise.
- [ ] Add read fixtures with inexact / truncated bounds (e.g. produced by parquet-mr `BinaryTruncator`) to prove the aggregate path does not answer from a non-exact bound.

No change is needed to the existing bound-based pruning.

## Write side (already handled)

[...]

PR #816, opened 2026-07-24
#9 Write support stage 11: column statistics

gunnarmorling:
Stage 11 of the writer delivery plan ([`_designs/WRITER_SUPPORT.md`](_designs/WRITER_SUPPORT.md)): produced files now carry column statistics, so a reader can prune row groups against a predicate. Tracking issue: #9.

## What

- **`StatisticsCollector`** accumulates each column chunk’s `min` / `max` / `null_count` over the shredded entry stream as it is encoded.
- **`StatisticsWriter`** (inverse of `StatisticsReader`) serializes them into `ColumnMetaData` (thrift field 12).

## Correctness notes

- **Bounds** span only present values and use signed `INT32` comparison, matching the column’s type-defined `ColumnOrder`, so they are pruning-correct.
- **Null count** is every not-present slot — a null leaf, a null or empty list, or a null struct ancestor — mirroring how the reader counts nulls, so a fully null column carries a count but no bounds.
- Written through the **preferred** `min_value` / `max_value` fields, never the deprecated `min` / `max`.
- Each bound is flagged **exact** (`is_min_value_exact` / `is_max_value_exact`), since the writer stores the actual extreme, never a truncated approximation. Stage 12 will flag truncated `BYTE_ARRAY` bounds inexact.

This is **column-chunk** statistics (row-group pruning). Per-page statistics (inline `DataPageHeader` stats + ColumnIndex/OffsetIndex, for page-level skipping) are deferred to the page-index work, as recorded in the design.

## Validation

- **Round-trip** (`WriterRoundTripTest`): min/max/null_count for required, nullable, all-null, per-row-group, and nested-list columns (the list case pins the null-count semantics: null list + empty list + null element = 3).
- **DuckDB differential** (`WriterDifferentialTest`): `parquet_metadata()` decodes our bytes back to the expected `min_value` / `max_value` / `null_count` and `min_is_exact` / `max_is_exact = true`, with the deprecated fields absent — an independent implementation reading our statistics. A second test confirms a `BETWEEN` filter across row groups drops no matching rows.
- Full core suite green (1741 tests).

## Follow-up

- #815 — read side should honor `is_*_value_exact` **if/when** aggregate-from-statistics pushdown is added (not a current bug; pushdown uses stats only as bounds).

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -307,9 +307,9 @@ For field-level `parquet.thrift` metadata coverage (which spec fields are read/p
 - [x] Implement `Statistics` record (minValue, maxValue, nullCount, distinctCount)
 - [x] Statistics deserialization (`StatisticsReader` with deprecated/preferred field fallback)
 - [x] Type-specific comparators (`StatisticsDecoder` for int, long, float, double, boolean, binary)
-- [ ] Statistics collection during writing
+- [x] Statistics collection during writing (`StatisticsCollector`, INT32 min/max/null_count)
 - [ ] Binary min/max truncation for efficiency
-- [ ] Statistics serialization
+- [x] Statistics serialization (`StatisticsWriter`, preferred `min_value`/`max_value` fields)
 
 ### 9.2 Page Index (Column Index & Offset Index)
 - [x] Implement `ColumnIndex` structure
--- a/_designs/WRITER_SUPPORT.md
+++ b/_designs/WRITER_SUPPORT.md
@@ -38,10 +38,11 @@ column's physical bytes are written by the primitive-type increment, and the ann
 is serialized onto the schema and converted at the API boundary.
 
 Sequenced as later work, each its own design: DataPage V2, the Avro write API, the
-optional index structures (OffsetIndex, ColumnIndex, Bloom filters), the optional delta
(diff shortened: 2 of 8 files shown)


gunnarmorling:
Added a follow-up commit (`dabce7d0`) restructuring the delivery plan in `_designs/WRITER_SUPPORT.md`, bundled here because it edits the same regions this PR already touches (a separate branch off `main` would conflict):

- **Fixes a contradiction**: the delta / byte-stream-split encoders were listed as both stage 14 *and* deferred "later work, each its own design." Dropped the stray deferred entry.
- **Promotes the optional index structures to numbered increments 20 (page indexes: OffsetIndex/ColumnIndex + per-page statistics) and 21 (Bloom filters)**, on the settled surface alongside the S3 backend (19) — rather than leaving them in a vague prose "later work" bucket. Page indexes are where the per-page statistics deferred by this stage-11 work land.
- Only genuinely-separate downstream layers remain sequenced as later milestones: DataPage V2, the Avro write adapter, and the CLI write/convert command.

Happy to split this into its own PR if you'd rather keep #816 to the statistics code — it's a clean revert of the last commit here.

Ari4ka (approved):

Issue #806, opened 2026-07-20
CLI: aesh help/error text polish (stale command-not-found hint, spurious 'unknown option' on group-command --help)


gunnarmorling:
Two user-facing rough edges in the CLI's help/error text surfaced after the picocli → aesh migration (#688). Both are cosmetic — commands still work and exit codes are correct — but the output is misleading.

## 1. Command-not-found suggests a command that no longer exists

```
$ hardwood bogus
'hardwood bogus' is not part of the hardwood commands. See 'help hardwood'.
...usage...

$ hardwood help
'hardwood help' is not part of the hardwood commands. See 'help hardwood'.
...usage...
```

The migration removed the `help` subcommand (help is now `--help`/`-h`), but aesh's built-in command-not-found message points the user at `help hardwood`, which doesn't exist. It should suggest `hardwood --help` instead.

This message comes from aesh's `AeshRuntimeRunner` execution path, whose builder exposes no hook to override it. aesh does ship a `CommandNotFoundHandler` interface, but wiring one in means bootstrapping a hand-built `AeshCommandRuntime` instead of `AeshRuntimeRunner` — a `Main` refactor, not a one-liner.

## 2. `inspect --help` prints a spurious "unknown option" before the help

```
$ hardwood inspect --help
The option --help is unknown.
Low-level introspection commands.
Usage: hardwood inspect [-h] [COMMAND]
...
```

`--help` is valid (the help renders), but aesh emits `The option --help is unknown.` first. This happens only on the nested **group** command (`inspect`); `hardwood --help` and leaf commands like `hardwood print --help` are clean. Looks like aesh doesn't apply `generateHelp`'s `--help` option to a nested group command the same way, so the parser rejects it before falling back to rendering help.

## Notes

- Neither is cleanly fixable from our side without either an upstream aesh change or restructuring `Main` onto a custom `AeshCommandRuntime`; worth weighing that cost against the cosmetic benefit.
- Discovered while assessing the CLI surface for the #688 migration; the reference docs and release notes are already updated.

gunnarmorling:
Update on item 1 (the `See 'help hardwood'` hint on an unknown command): this is an upstream aesh bug, not hardwood-specific — the message is hardcoded in aesh's `AeshCommandLineParser` and points at a `help` command aesh doesn't provide.

Filed and fixed upstream:
- aeshell/aesh#555 (issue)
[...]

PR #817, opened 2026-07-24
build: upgrade aesh to 3.16.6 (#806)

stalep:
Fixes two CLI help/error text issues:
- Command-not-found message now says 'See hardwood --help' instead of the stale 'See help hardwood' (aesh#557)
- 'hardwood inspect --help' no longer prints spurious 'The option --help is unknown' before the help output (aesh#558)

Also separates aesh and aesh-readline version properties since they are independent projects with independent release cycles.


stalep:
Hi @gunnarmorling, I think this should fix #806

gunnarmorling:
Ah yes, very nice. Thanks for the upgrade, and also for the clarification on the separate versions, I had assumed they're one and the same. Will await CI and merge.

Issue #810, opened 2026-07-21
Column reader under-utilizes the decode pool for many-small-page columns under compression


gunnarmorling:
## Summary

The parallel column-read pipeline under-utilizes CPU cores when a column chunk is split into **many small pages**, specifically under page compression where each per-page decode is CPU-heavy (decompression). `ColumnWorker` submits **one decode task per page** (`CompletableFuture.runAsync` on the decode executor), fed by a **single retriever thread** and throttled to `MAX_INFLIGHT_PAGES`. With many tiny pages, the per-task overhead plus the single retriever's read/parse/submit rate can't keep the decode pool saturated, so cores sit idle.

## Evidence

Reading the **same 8M float32 values** two ways, ZSTD-compressed:

| column | pagination | ms/op (N300, ZSTD) |
|---|---|---|
| flat `float` (1 value/row) | **400 pages × ~74 KB** | 25.7 (slower) |
| `LIST<float>` leaf, k=768 (768 values/row) | **29 pages × ~1 MB** | 23.2 |

The flat read is strictly *less* work yet ~10% slower on the N300 (and ~55% slower on an AWS m7i.2xlarge). The tiny pages come from parquet-mr's default 20,000-row page cap: a 1-value-per-row column hits the row cap (→ tiny pages) while a 768-value-per-row column is size-limited (→ ~1 MB pages).

- **perfnorm:** near-equal instructions, but the tiny-page read burns **~31% fewer core-cycles over more wall-time** (high IPC, cores only ~70% busy) — the pool is starved, not compute- or cache-bound (the tiny-page read actually has *fewer* LLC misses).
- **Wall-clock profile:** the hot frame in both is ZSTD Huffman decompression; the tiny-page read keeps fewer cores fed.
- **Uncompressed, the gap nearly vanishes (~7%)** — confirming it's the per-page *decompression* work being under-parallelized.
- **Repaginating the flat file into big pages** (`withPageRowCountLimit` raised → ~34 × ~870 KB) collapses the difference: flat ≈ list. This isolates pagination/page-count as the cause.

## Impact

Real flat/narrow columns commonly hit the 20k-row page cap and end up with many tiny pages, so this affects real-world reads of such columns under compression, not just a synthetic case.

## Suggested directions

- Coalesce adjacent small pages into larger decode units (each task does more work), or
- Batch multiple small pages per decode task, or
- Parallelize/pipeline the retriever (page fetch + header parse) so it can feed the pool faster.

[...]

PR #824, opened 2026-07-25
#810 Batch small compressed page decodes

arnabnandy7:
## Summary

Addresses #810 by batching adjacent small compressed pages into bounded decode work units.

Compressed work units contain up to four pages or 1 MiB of compressed input. The reorder window allows multiple work units to execute concurrently, reducing executor submission overhead without limiting decode-pool utilization.

Large compressed pages, uncompressed pages, and null placeholders continue using single-page tasks and the existing retention limit.

## Testing

- Added generated-file coverage verifying small compressed pages share decode tasks.
- Verified decoded values retain their original order.
- Verified uncompressed pages continue using one task per page.
- `./mvnw -pl core verify` passes.
- Full verification requires Docker for the existing S3 Testcontainers tests.

## Checklist

### DIFF
--- a/_designs/COLUMN_READER_SMALL_PAGE_WINDOW.md
+++ b/_designs/COLUMN_READER_SMALL_PAGE_WINDOW.md
@@ -1,0 +1,33 @@
+# Column Reader Small-Page Window
+
+Status: Implemented
+
+## Overview
+
+The column reader uses an expanded in-flight window for stored-small compressed pages. Pages below 128 KiB may occupy up to sixteen reorder slots with the default configuration. Uncompressed pages, null placeholders, and compressed pages at or above the threshold retain the original eight-page limit.
+
+Every Parquet page remains an independent executor task. The policy changes how far the retriever may run ahead, not decode work-unit granularity.
+
+## Pipeline
+
+The retriever remains the sole consumer of `PageSource`. For each page it selects an in-flight limit from the page's codec and stored size, waits for capacity, reserves a sequence number and reorder-buffer slot, and submits the page to the shared decode executor.
+
+Decode tasks publish results independently. The drain consumes them in sequence order and retains responsibility for file boundaries, filter boundaries, batch assembly, and row limits.
+
+## Bounds and transitions
+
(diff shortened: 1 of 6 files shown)


gunnarmorling:
Have you done any benchmark here? When I tried some time ago, it didn't bring any improvement or even made things worse.

arnabnandy7:
> Have you done any benchmark here? When I tried some time ago, it didn't bring any improvement or even made things worse.

Even back then I provided updated solution with betterment in benchmark which I'd suggest for you to revisit once if it can be considered.

I'll surely share the benchmark for this one as well shortly.

arnabnandy7:
@gunnarmorling thanks for raising this. The original multi-page task approach did regress, so I replaced it rather than presenting those results.

The revised implementation keeps one decode task per page and only expands the in-flight window from 8 to 16 for compressed pages smaller than 128 KiB. Large compressed and uncompressed pages retain the existing eight-page limit.

I added a reproducible JMH benchmark over the same eight million float32 values and measured baseline and adaptive behavior from the same binary using an internal diagnostic switch.

| Configuration | Small-page ZSTD |
| --- | ---: |
| Existing 8-page window | 44.120 ± 4.002 ms/op |
| Adaptive 16-page window | 34.182 ± 3.849 ms/op |

That is a **22.5% improvement**. A forward-order run measured `54.287 ± 18.110 ms/op` versus `33.065 ± 4.791 ms/op`, a **39.1% improvement**.

As a negative control, large-page ZSTD was effectively unchanged:

| Configuration | Large-page ZSTD |
| --- | ---: |
| Existing 8-page window | 39.938 ± 3.954 ms/op |
| Adaptive 16-page window | 39.874 ± 12.738 ms/op |

The benchmark generator and JMH benchmark are included in the updated change so the results can be reproduced and validated on other machines.

gunnarmorling:
@arnabnandy7, could you rebase this to current main, now that the level scratch work has been integrated? We also should have an end-to-end test exercising the larger window. Thanks!

arnabnandy7:
@gunnarmorling it's done

PR #825, opened 2026-07-25
#822 Add parquet-dev mailing list search skill

### LINKED ISSUE #822: Add a skill for searching the parquet-dev mailing list
Add a repo skill that searches the Apache `dev@parquet.apache.org` archive by downloading the monthly `mbox` files from `lists.apache.org` and grepping them locally, rather than relying on the JavaScript-rendered archive UI or a summarizing web fetch (both of which proved unreliable).

The skill caches immutable past months and always re-fetches the live current month, and ships a helper to print a full thread with quoting stripped. Motivated by needing to pull the "[DISCUSS] Inline parquet.thrift into parquet-java" thread while scoping #820.

arnabnandy7:
## Summary

Adds a repository skill for searching the Apache `dev@parquet.apache.org` mailing-list archive through raw monthly mbox files.

The skill:

- Caches immutable historical months and refreshes the current month.
- Searches downloaded archives locally instead of relying on the JavaScript-rendered archive.
- Extracts complete threads in chronological order.
- Removes duplicate messages, quoted replies, attribution lines, and signatures.
- Validates inputs and fails early for missing archives.
- Uses atomic downloads to protect valid cached archives from incomplete downloads.

## Differences from #823

This is an alternative implementation of the same issue. It follows the overall mbox-based workflow introduced in #823 while adding several safeguards:

- Validates the complete `YYYY-MM` range instead of accepting invalid months or silently skipping malformed input.
- Downloads into a temporary file and atomically replaces the cache entry only after a successful transfer.
- Rejects missing mbox paths instead of allowing Python's mailbox API to create empty files.
- Uses a content-derived identity when a message lacks a `Message-ID`, preventing unrelated messages without IDs from being collapsed.
- Ignores named multipart attachments when selecting the plain-text body.
- Removes common reply-attribution blocks alongside `>`-quoted content.
- Uses Unicode-aware case-insensitive subject matching.

Closes #822.

## Checklist

### DIFF
--- a/.claude/skills/parquet-dev-ml/SKILL.md
+++ b/.claude/skills/parquet-dev-ml/SKILL.md
@@ -53,7 +53,7 @@ SKILL="/workspace/.claude/skills/parquet-dev-ml"
    "$SKILL/thread.py" "inline parquet.thrift" "$CACHE"/parquet-dev-2026-07.mbox
    ```
 
-   `thread.py` takes a case-insensitive Subject substring and one or more mbox files, matches every message in that thread, dedupes by `Message-ID`, drops `>`-quoted lines and signatures, and prints each message's From/Date/body.
+   `thread.py` takes a case-insensitive Subject substring and one or more existing mbox files, matches every message in that thread, dedupes by `Message-ID` (or message content when the ID is absent), drops `>`-quoted lines, reply attributions, and signatures, and prints each message's From/Date/body.
 
 ## Reporting back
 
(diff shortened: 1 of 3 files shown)


gunnarmorling:
Thanks, @arnabnandy7. Could you rework this and send it as PR on top of my original commit, rather than squashing them. Thx!

arnabnandy7:
> Thanks, @arnabnandy7. Could you rework this and send it as PR on top of my original commit, rather than squashing them. Thx!

Ofcourse I'll do that

arnabnandy7:
@gunnarmorling I've changed the base from main to your branch, please review whenever possible for you.

gunnarmorling:
Nice one, thanks a lot, @arnabnandy7! Rebased and applied.

PR #826, opened 2026-07-25
#814 Pool per-page level scratch by decode slot

### LINKED ISSUE #814: Pool per-page repetition/definition level scratch across in-flight decode slots
## Problem

`PageDecoder.decodeRepetitionLevels` / `decodeDefinitionLevels` allocate a fresh `new int[numValues]` for every page decoded (`PageDecoder.java:158-187`). This scratch is pure internal state — consumed element-by-element during assembly (`NestedColumnWorker.assembleRegularPage`) to build the reused batch accumulators, and never handed to the consumer. It is not pooled.

Batch holders and the value/level *accumulators* are already recycled (`BatchExchange` recycling mode; `nestedValues`/`nestedDefLevels`/`nestedRepLevels` grown-and-reused), but the *per-page decode scratch* is a distinct array population that escapes this pooling. Because the code lives in the shared `PageDecoder`, both the flat (optional-column def levels) and nested (rep + def levels) paths pay it.

A JFR allocation profile of a full LIST<float32> column scan (fast path off) attributes the single largest `int[]` allocation site to `decodeRepetitionLevels`, and this per-batch churn is a meaningful contributor to GC pressure — which in turn shows up as run-to-run throughput variance (the general reconstruction path wobbles ~2.5–3.5% CV run-to-run while the fixed-size-list fast path, which skips level materialization, is stable at ~0.2%).

## Proposed fix

Decode runs on a pooled executor with up to `MAX_INFLIGHT_PAGES` (default 8) pages in flight, each parked in a reorder-buffer slot and drained (assembled) synchronously by the drain thread. The retriever throttle guarantees a slot is not reused until its previous occupant has been drained, so a **per-slot** scratch buffer (keyed by `slot`, not thread-local) is reused safely with no concurrent pool or release protocol:

- Add a per-slot `LevelScratch { int[] rep; int[] def; }` array of size `MAX_INFLIGHT_PAGES` in `ColumnWorker`.
- Thread the slot's holder through `decode(slot, …)` → `PageDecoder.decodePage(…, scratch)`; `decodeRepetitionLevels`/`decodeDefinitionLevels` reuse and grow the buffer instead of allocating fresh. All-present def levels still resolve to `null`.
- Single-page callers (tooling/tests) pass a null holder and fall back to fresh allocation.

Safety rests on: slot exclusivity (throttle), page level arrays consumed during assembly and never retained (confirmed across all `IndexMode`s — `REAL_VIEW_KEEP_LEVELS` keeps the accumulator levels, not the page arrays), and no consumer depending on `repetitionLevels().length` / `definitionLevels().length` (all use `page.size()`).

Over a scan the buffers grow to the largest page once and then stop allocating, dropping steady-state level-scratch allocation to ~zero for both flat and nested reads.

arnabnandy7:
## Summary

Pools repetition and definition-level decode scratch buffers per in-flight reorder slot, avoiding fresh `int[]` allocations for each decoded page.

- Reuses and grows slot-owned level buffers as needed.
- Preserves fresh allocation for standalone page-decoder callers.
- Preserves the all-present `null` definition-level fast path.
- Ensures dictionary decoding only examines the active page region when a reused buffer is larger than the current page.
- Supports both scalar and SIMD non-null counting paths.

Closes #814

## Testing

- Added coverage verifying level scratch reuse across pages.
- Added coverage ensuring unused scratch-buffer tails are ignored.
- All 1,743 core tests pass.
- Full verification reaches the S3 module but cannot run its Testcontainers tests without Docker.

## Checklist

The documentation item is marked complete because this is an internal allocation optimization with no user-facing behavior or public API changes.

### DIFF
--- a/_designs/POOLED_LEVEL_DECODE_SCRATCH.md
+++ b/_designs/POOLED_LEVEL_DECODE_SCRATCH.md
@@ -1,0 +1,77 @@
+# Design: pooled repetition/definition level decode scratch
+
+**Status: Implemented.** Tracking issue: #814.
+
+## Goal
+
+Stop allocating a fresh repetition/definition level array on every page
+decode. Level decoding runs on the hot read path for every optional or
+repeated column, and each page previously allocated `new int[numValues]`
+per level stream. For a full nested column scan this is the single
+largest `int[]` allocation site, and the per-page churn is a meaningful
+source of GC pressure — which surfaces as run-to-run throughput variance
+on the general reconstruction path (the fixed-size-list fast path, which
+skips level materialization, does not exhibit it).
+
+The batch holders and the value/level *accumulators* are already reused
+across batches; the per-page *decode scratch* is a separate array
+population that was not pooled. This design pools it.
+
+## Behaviour
+
+Level scratch is reused per in-flight reorder-buffer slot.
+
+`ColumnWorker` submits page decodes to a pooled executor, keeping up to
+`MAX_INFLIGHT_PAGES` decoded pages parked in an `AtomicReferenceArray`
+reorder buffer indexed by `slot = seq % MAX_INFLIGHT_PAGES`. The drain
+thread assembles each page synchronously and advances `consumePosition`;
+the retriever throttles submission while `nextSeq - consumePosition >=
+MAX_INFLIGHT_PAGES`, so a slot is never reused for a new decode until its
+previous occupant has been fully drained.
+
+That throttle is the exclusion guarantee the pooling relies on. Each slot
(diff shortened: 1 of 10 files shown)


gunnarmorling:
Rebased and merged. Added one more commit to pick up a few lose ends. Thanks, @arnabnandy7!

arnabnandy7:
There was a test failure in last commit @gunnarmorling was that handled?

gunnarmorling:
Yupp, it was due to a change in the upstream parquet-testing project. Fixed via #843.

Issue #827, opened 2026-07-27
Perflab: evidence for I/O planning decisions


artemkach:
#763 asks for a deterministic way to measure read wall-clock under controlled storage latency before any I/O planning code changes. This issue tracks a design proposal that extends that plan: a small set of instruments for capturing what the read planner decides, pricing those decisions under stated storage assumptions, and cross-checking the predictions against controlled execution.

The full proposal is a design document (PR to follow shortly), backed by a working prototype and a demo. Highlights:

- **iotrace** — capture the planner's fetch plan and the actual reads through default-disabled JFR events, with completeness checks; includes a `hardwood iotrace` CLI command that renders both for any file.
- **analytic scorer** — price a captured plan under a declared `(latency, bandwidth, concurrency)` profile in microseconds, enabling cheap sweeps across the parameter space of storage conditions.
- **763's sleep benchmark** as specified, as one of the lab's instruments.
- A first **cross-method comparison**: do the scorer, the sleep benchmark, and an impaired-transport run agree on which of two real plans is faster?

No production behavior changes and no new public API in this proposal; the current planner defaults stay throughout. The design doc ends with open questions where early feedback would be most valuable (first-increment scope, the CLI command, Toxiproxy now-or-later, storage profiles).

Relationship to #763: complementary, not a replacement — the benchmark remains the first deliverable and stands on its own; this proposal adds plan capture, scoring, and cross-checks around it.

Issue #763, opened 2026-07-06
Adaptive, latency-aware I/O coalescing for remote backends (bench-first)


gunnarmorling:
## Motivation

The read path's I/O coalescing and prefetch decisions are governed by byte constants that are implicitly tuned for local / memory-mapped storage:

- `MAX_CROSS_COL_GAP_BYTES = 64 KB` — max dead-byte gap bridged when coalescing adjacent column chunks into one ranged read (`RowGroupIterator`).
- `PAGE_COALESCE_GAP_BYTES = 1 MB`, `MAX_COALESCED_BYTES = 128 MB` — intra-column page coalescing and region cap.
- `SequentialFetchPlan` chunk sizing, one-ahead prefetch in `ChunkHandle`.

On an object store these constants are the wrong order of magnitude. The right gap threshold is a **break-even**: bridge a gap iff `gap_bytes / bandwidth < request_latency`, i.e. `g* = latency × bandwidth`. Backing out the constant, 64 KB assumes local-storage economics; on S3 (`L ≈ 30 ms`, `B ≈ 80 MB/s per connection`) the break-even is in the **megabytes** — so today we issue extra round trips to avoid fetching a few hundred KB we could have swept up in a single GET.

Crucially, the optimum is **not knowable at plan time** — it depends on the storage backend's latency/bandwidth, which only measurement reveals. This is a case for a small online estimator feeding a closed-form planner decision, **not** a feedback controller (the read horizon is a single bounded pipeline, and the objective has a closed form).

## The parallelism caveat

I/O is already parallelized across columns: each `ColumnWorker` runs a retriever virtual thread that issues blocking `readRange` calls, so N projected columns produce up to N concurrent GETs (the JDK `HttpClient` does not cap this — its connection pool bounds idle-connection *reuse*, not concurrent in-flight requests). Because `coalesceAcrossColumns` fuses several columns' first reads onto one thread, **coalescing and cross-column parallelism are the same knob pulled in opposite directions**. This splits gap tuning into two cases:

- **Non-zero gaps (dead bytes between chunks)** — clean `g* = L·B` break-even; raising the threshold on S3 is an unambiguous win.
[...]

PR #828, opened 2026-07-27
#827 Perflab design document, draft for early feedback

artemkach:
Refs #827

Design proposal for a set of instruments that capture what the read planner decides, price those decisions under stated storage assumptions, and cross-check the predictions against controlled execution — extending the plan laid out in #763, whose sleep benchmark remains the first deliverable and one of the instruments.

Doc-only branch: the working prototype and demo the document references live on a fork (linked from the document) and are not part of this change.

## Checklist


gunnarmorling:
Hey, thanks a lot for this! I'm traveling atm., will get to this and provide more feedback once I'm back. Thanks for your patience!

_designs/PERFLAB.md
```diff
@@ -0,0 +1,1319 @@
...
+  | Node | Stage | Range          | Length | Serves                 |
+  +------+-------+----------------+--------+------------------------+
+  |    1 |  DATA | [4 .. 7811814) | 7.4 MB | 7 requirements (fused) |
+  +------+-------+----------------+--------+------------------------+
+
+Trace (4 reads at the InputFile seam, invocation order)
+  +---+-----------+----------+---------+--------+----------+
+  | # | Begin     | Duration | Offset  | Length | Matched  |
+  +---+-----------+----------+---------+--------+----------+
+  | 1 |   t+0.0µs |  235.1ms |       0 |    4 B | metadata |
```
gunnarmorling, line 366: There's one request before which is missing from the traces: the tail fetch (last 64KB) in `S3InputFile::open()` (which also is the reason why requests 2 and 3 are that fast then, as they are served from that cache).

_designs/PERFLAB.md
```diff
@@ -0,0 +1,1319 @@
...
+contender runs always execute with capture disabled, in separate
+processes.
+
+---
+
+## 9. Open questions
+
+Early feedback is most valuable on these:
+
+1. **Scope of the first increment.** #763 step 1 (the sleep benchmark)
```
gunnarmorling, line 688: I'd say let's focus on the capture part at first; I'm all for shipping stuff in small, manageable increments as much as possible. I.e. 763 benchmark first, then capture, then scorer and cost-based request shaping.

_designs/PERFLAB.md
```diff
@@ -0,0 +1,1319 @@
...
+   stands on its own. Should plan capture + scorer land in the same design,
+   or as a follow-up once the benchmark exists? The cross-method comparison
+   needs both, but the benchmark alone already answers real questions.
+2. **The CLI command.** `hardwood iotrace` was built as an exploration
```
gunnarmorling, line 692: I love it and I think we should definitely have it. My main question is whether it should be a separate command as proposed right now, or rather a `--iotrace` flag on each other command. The latter might interfere with the actual command output though, so perhaps a separate command is better indeed. Tracing specific predicate push-down runs would be a very nice follow-up.

_designs/PERFLAB.md
```diff
@@ -0,0 +1,1319 @@
...
+   tool and turned out broadly useful for understanding I/O behavior on
+   any file. Should it ship as a supported command, or stay a development
+   tool? (Related: is the `iotrace` name agreeable — it also names the JFR
+   event namespace, which becomes a de-facto contract once recordings
+   exist.)
+3. **The gap override.** Comparing fused and split plans on identical bytes
```
gunnarmorling, line 698: System prop seems fine.

_designs/PERFLAB.md
```diff
@@ -0,0 +1,1319 @@
...
+   needs a test-scoped override for the cross-column gap. A system property
+   matching the two existing sibling knobs is the two-line version; an
+   internal policy object is the cleaner one. Preference?
+4. **Toxiproxy now or later?** The existing S3Proxy container already
```
gunnarmorling, line 702: Preferably later, if we can. Same reason as above: smaller, digestible chunks whenever we can.

_designs/PERFLAB.md
```diff
@@ -0,0 +1,1319 @@
...
+   provides configurable latency, but its bandwidth throttle is broken
+   (Appendix C), so a bandwidth-controlled transport lane means adding one
+   pinned Toxiproxy container plus the calibration protocol. Options:
+   include it in the first increment (full shared-profile cross-check), or
+   start latency-only on plain S3Proxy and add Toxiproxy the first time
+   simulation and controlled delay disagree on a bandwidth-sensitive
+   question. Preference?
+5. **Repository boundary.** The proposal assumes the main repo under
```
gunnarmorling, line 710: I think we'd want an S3 benchmark in the _hardwood-benchmarks_ repo eventually, but this seems unrelated to the work done here. For context, that repo primarily houses the benchmarks used in publications (blog posts). So for example, we could have a blog post about S3 read performance a some point, and the supporting benchmark for that post would be in that repo then.

_designs/PERFLAB.md
```diff
@@ -0,0 +1,1319 @@
...
+   `-Pperformance-test`, consistent with #763's framing. Does any part of
+   this (e.g. the transport lane and its containers) belong in
+   `hardwood-benchmarks` instead?
+6. **Storage profiles.** Are the initial profiles (§8.6) the right
```
gunnarmorling, line 714: I think we can decide on that when we get to the actual scorer/planner increment?

_designs/PERFLAB.md
```diff
@@ -0,0 +1,1319 @@
...
+   exploration points — and roughly what would you want to see before a
+   planner default changes? For calibration: is "cross-method agreement at
+   the two S3-like profiles, on the pilot fixture plus one real-world
+   file" the right neighborhood, or too low/high?
+7. **JFR event surface.** Six new default-disabled events is a real
```
gunnarmorling, line 719: Let's only ship what we actually need increment by increment.

gunnarmorling (commented):
Hey @artemkach, thanks a ton, this is really great work! Gotta admit, it's a lot to take in, though :) So I did a first pass, trying to answer your questions. One comment inline on a critical trace omission. Overall, it would be great if we could integrate this step by step (see Q1).

gunnarmorling:
_(via Claude Code, with some subsequent editing from my side)_

A few follow-ups to the pass above.

**Your traces already found a bug.** The 4-byte read at offset 0 that costs a full 235 ms in the ap-southeast-2 run is the leading `PAR1` magic check — it sits at the opposite end of the file from the tail cache, so it always misses. The first data node already starts at offset 4, so extending it to 0 removes a round trip from every remote open for 4 extra bytes. Filed as #852, crediting the traces in this PR. Nice demonstration of the point you're making with the instrument.

**On Q1, to be unambiguous about the order:** #763's sleep benchmark first, then capture, then scorer. If you'd rather start on capture and let #763 land separately, say so — I don't want the benchmark to become a blocker on the part that's already prototyped. Either way the two are independent enough that they needn't be one PR.

**One decision I'd like settled before capture code lands:** the JFR event namespace. All ten existing events are flat — `@Name("dev.hardwood.RowGroupByteRangeFilter")` in `dev.hardwood.jfr`. The proposal introduces `dev.hardwood.iotrace.*` under `dev.hardwood.jfr.iotrace`. I'm happy with the sub-namespace, since a six-event family is a fair reason to group, but let's state it in the design as a deliberate convention for event families rather than have a second naming shape appear by accident. These names become a contract as soon as anyone has a recording, so it's a thing that's genuinely hard to walk back.

**On the document itself:** given the increment-by-increment answer to Q1, I'd like it cut down to an end-state design for the first increment before merge. Our convention is that `_designs/` docs describe the intended end state — no open questions left standing, no spike narration. Concretely: drop the "draft, seeking feedback" framing and §9, fold §7 into the design as what it specifies rather than what was built, and demote scorer / transport / pilot to a short "later increments" section stated as intent. Keep §5.3, Appendix B and Appendix E verbatim — the survey and the measured runs are the durable part and I don't want to lose them. Mostly deletion rather than rewriting.

Small things to fix in the same pass:

- Pin the proof-of-concept links to a commit SHA — they point at a fork branch that can be force-pushed or deleted, and they carry the evidence for §7.1.
- §1 says the demo gave "opposite verdicts in different environments"; Appendix E is fused-wins / tie / moot and names the split-wins regime as the missing cell. Your own §7.2 lesson 1 ("the verdict is conditional") is the accurate phrasing.
- The same result is "~19% sooner" in §7.2 and "~24%" in Appendix E — different denominators on 1929 vs 2384 ms.
- §8.6's `SAME_REGION_S3` uses 50 MiB/s where #763 uses 80 MB/s per connection; since `g* = L × B` that's 1.5 MiB vs ~2.4 MiB break-even from nominally the same profile.
- Status line: `**Status: Proposed.** Tracking issue: #827.` to match the other design docs.

Happy to take the trimmed design and the first capture increment as one PR if that's easier than splitting them.

artemkach:
Glad the added observability is paying off already. I will work on the trimmed design and incremental delivery plan. (My availability might be spotty in the coming months, apologies in advance for bursty PR timing).

artemkach:
@gunnarmorling I finally had the time to get back to this. Thanks for the comments - all accepted.

Taking your offer on sequencing: **#763's sleep benchmark first** as its own PR, then **the trimmed design together with the JFR events and `iotrace`** as one. Might grab #852 as well, unless someone gets to it first.

You probably saw that DuckDB [announced](https://duckdb.org/2026/07/31/asynchronous-io) asynchronous I/O for the upcoming v2.0, which gives them the most developed I/O planning in my §5.3 survey. Their coalescing gap is now `latency × bandwidth`, clamped to `[4 KiB, 32 MiB]`, EWMA-smoothed, sitting at a 16 KiB static default until the first measurement arrives, and pinnable as a user setting. Two things there influenced how I'm thinking about capture.

The first is **where the timing boundary sits**. DuckDB uses two different estimators depending on what the transport can report: on the remote path each request contributes a measured time-to-first-byte and a body-transfer rate; on the local path, with no TTFB available, they instead fit `total_time = intercept + slope × bytes` by least squares - separating the fixed per-request cost from the per-byte cost - and take latency from the intercept. That fit refuses to produce an estimate until it has seen reads of at least two different sizes, because with one size the fixed and per-byte costs aren't separately identifiable. Our `RequestEvent` carries one duration today. I checked that TTFB is already observable inside `S3InputFile.readRange`, but not at the generic `InputFile` seam, which only sees call-in and call-out - so splitting the two terms is a per-backend contribution rather than something the seam can do, which is the same shape DuckDB arrived at. Even if the TTFB field itself waits, I'd like to pin down what the existing duration field means now, since that's the part that's hard to revise once recordings exist.

The second is **structure**. Their plan capture is a queryable log channel with one record per row group, keyed on `(file_path, row_group_id)`, and every assertion in their test suite is expressible as a query because of that key. Ours has weaker keys: no event carries file identity, and `PlanRequirement.columnRole` packs the row group and column index into a display string (`"rg0/col0"`) instead of carrying them as fields. I'd like to fix that in the capture increment rather than later, for two reasons: keys can't be added to recordings that already exist, and the reason to be in JFR at all is that our events sit in the same recording as `PageDecoded`, `RowGroupFilter` and `BatchWait`, so "this fetch was slow, what was decode doing" is a question we can answer and DuckDB's isolated log channel can't. Parseable `iotrace` output then falls out of the same work.

PR #830, opened 2026-08-02
fix: allow bloom pruning for NaN predicates

### LINKED ISSUE #829: Bloom filter push-down forfeits ±0 pruning on FLOAT/DOUBLE columns
## Summary

`BloomFilterSupport` refuses to consult the bloom filter for any `±0` probe on a `FLOAT` or `DOUBLE`
column. The guard is safe but unnecessary: it forfeits valid pruning on every float column, and the
rationale in its JavaDoc describes semantics this codebase does not use.

## Current behaviour

`core/src/main/java/dev/hardwood/internal/predicate/BloomFilterSupport.java:50`

```java
/// Bloom filters hash the raw IEEE-754 bits, which distinguish `-0.0f` from `+0.0f` and every
/// NaN bit pattern — but `FLOAT` equality treats `-0.0f == +0.0f` and `NaN != NaN`. Probing
/// those by raw bits could prove a value absent that an equal stored value would match, so they
/// are never pruned here (statistics still apply).
static boolean valueAbsent(BloomFilterSource bloomFilters, int columnIndex, float value) {
    if (value == 0.0f || Float.isNaN(value)) {
        return false;
    }
```

Note the guard is written with `==`, i.e. IEEE arithmetic equality, so it catches **both** zeros.
`-0.0f == 0.0f` is `true`, so `eq(col, -0.0f)` and `eq(col, +0.0f)` both skip the filter entirely.
Same shape in the `double` overload at line 60.

## Why the stated rationale does not hold

The carve-out rests entirely on a claim about what `col = x` *means* for a `FLOAT` column: that
equality treats `-0.0f` and `+0.0f` as the same value, and `NaN` as unequal to itself. That is Java
`==` semantics, and it is not what this reader implements.

What `col = x` means here is fixed by the code that decides which rows survive:

- `internal/predicate/matcher/floats/*`, `internal/predicate/matcher/doubles/*`
- `internal/predicate/RecordFilterCompiler` (both the struct-accessor and `RowReader` paths)

Every one of those compares with `Float.compare` / `Double.compare`. Under that total order `-0.0f`
and `+0.0f` are **distinct** and `NaN` **equals** `NaN` — both halves of the premise are inverted.

The bloom filter itself does not compare anything; it hashes. That is precisely why the matchers are
the reference point rather than the mechanism: a probe prunes a whole row group, so it is safe only
if it cannot exclude a row those matchers would have accepted. The next section states that
condition exactly.

## Why removing the `±0` half is safe

A bloom probe may prune only when the hash never separates values the matcher treats as equal —
the hash's equivalence classes must be coarser than or equal to matcher equality. If the hash is
*finer*, a probe for X can miss and prune a stored Y that would have matched.

| values | `Float.compare` classes | raw-bit hash classes | hash finer? |
|---|---|---|---|
| non-NaN, incl. `±0` | one per bit pattern | one per bit pattern | no — safe to prune |
| `NaN` | **one** class, all payloads | ~2²³ classes, one per payload | **yes — unsafe** |

For `±0` the two partitions coincide, so probing by bits agrees with matching by `Float.compare`:
a filter built from a stored `-0.0f` correctly misses a `+0.0f` probe, and a `-0.0f` row would not
have matched `= +0.0f` anyway. The row group is dropped, and a read of it would have returned no
rows.

The `±0` ambiguity that `parquet.thrift` describes is a property of statistics min/max, whose zero
bounds writers are told to normalize (*"if the computed min value is zero (whether negative or
positive), `-0.0` should be written into the min statistics field"*). A bloom filter holds hashes of
exact stored values, so nothing was normalized and there is nothing to widen. `ColumnOrder` does not
enter into it either: it describes how min/max are ordered, and a bloom filter has no ordering.

## Why the NaN half must stay

Matching collapses every NaN payload into one value; hashing splits them apart. A column storing a
NaN with payload `0x7FF00001` puts that hash in the filter; a probe for `Float.NaN` hashes
`0x7FC00000`, misses, and prunes a row group whose rows the matcher would have matched. Nothing
canonicalizes along the way — parquet-java hashes via `cacheBuffer.putFloat(value)` and
`ByteBuffer.putFloat` preserves the raw pattern, and `XxHash64.hash(float)` uses
`Float.floatToRawIntBits` explicitly.

So `Float.isNaN(value)` stays; only the `value == 0.0f` disjunct goes.

## Proposed change

```diff
-        if (value == 0.0f || Float.isNaN(value)) {
+        if (Float.isNaN(value)) {
             return false;
         }
```

Same for the `double` overload, plus a JavaDoc rewrite stating the actual rule: the hash is exact
for every non-NaN value including `±0`, and NaN alone is excluded because hashing distinguishes
payloads that `Float.compare` treats as equal.

## Test plan

- New fixture in `tools/simple-datagen.py`: `FLOAT` and `DOUBLE` columns holding `-0.0` and never
  `+0.0`, with bloom filters enabled via `bloom_filter_options` (see the existing
  `bloom_filter_test.parquet` block for the shape). A `+0.0` probe is then provably absent while a
  `-0.0` probe is present.
- Assert through `RowGroupFilterEvaluator` that `eq(col, +0.0)` drops the row group and
  `eq(col, -0.0)` keeps it, on both types.
- Assert `eq(col, NaN)` still keeps the row group, so the remaining carve-out is pinned rather
  than incidental.
- Existing `BloomFilterPushDownTest` cases on `price` / `ratio` must be unaffected.

## Notes

Behaviour today is conservative, not wrong: the carve-out only costs pruning, so this is an
optimisation plus a documentation correction rather than a correctness fix. No public API changes.

tomatotomata:
## What changed

- remove the unnecessary zero guard from FLOAT and DOUBLE bloom-filter probes
- retain the NaN carve-out because Float.compare/Double.compare collapse NaN payloads while raw-bit hashing distinguishes them
- update the predicate JavaDocs and add focused signed-zero and NaN regression coverage

Closes #829

## Validation

- `git diff --check` passes
- Focused Maven test was attempted with `./mvnw -pl core -am -Dtest=BloomFilterPushDownTest -DfailIfNoTests=false test`
- The local environment has Java 17, while the project enforces Java 25, and the temporary Docker Java 25 fallback could not connect because the Docker daemon is unavailable

This is an AI-assisted contribution prepared under maintainer direction. The diff is limited to `BloomFilterSupport` and its focused regression tests.

### DIFF
--- a/core/src/main/java/dev/hardwood/internal/predicate/BloomFilterSupport.java
+++ b/core/src/main/java/dev/hardwood/internal/predicate/BloomFilterSupport.java
@@ -43,22 +43,21 @@ final class BloomFilterSupport {
 
     /// Single-value bloom check for `FLOAT` values; see the `INT32` overload.
     ///
-    /// Bloom filters hash the raw IEEE-754 bits, which distinguish `-0.0f` from `+0.0f` and every
-    /// NaN bit pattern — but `FLOAT` equality treats `-0.0f == +0.0f` and `NaN != NaN`. Probing
-    /// those by raw bits could prove a value absent that an equal stored value would match, so they
-    /// are never pruned here (statistics still apply).
+    /// Bloom filters hash the raw IEEE-754 bits, while `FLOAT` equality uses `Float.compare`.
+    /// The matcher treats every NaN payload as equal, but the hash distinguishes those payloads,
+    /// so NaN probes stay conservative here (statistics still apply).
     static boolean valueAbsent(BloomFilterSource bloomFilters, int columnIndex, float value) {
-        if (value == 0.0f || Float.isNaN(value)) {
+        if (Float.isNaN(value)) {
             return false;
         }
         BloomFilter bloomFilter = filterFor(bloomFilters, columnIndex);
         return bloomFilter != null && !bloomFilter.mightContain(XxHash64.hash(value));
     }
 
-    /// Single-value bloom check for `DOUBLE` values. See the `FLOAT` overload for the `±0` / `NaN`
+    /// Single-value bloom check for `DOUBLE` values. See the `FLOAT` overload for the NaN
     /// carve-out.
     static boolean valueAbsent(BloomFilterSource bloomFilters, int columnIndex, double value) {
-        if (value == 0.0 || Double.isNaN(value)) {
+        if (Double.isNaN(value)) {
             return false;
         }
         BloomFilter bloomFilter = filterFor(bloomFilters, columnIndex);
--- a/core/src/test/java/dev/hardwood/internal/predicate/BloomFilterPushDownTest.java
+++ b/core/src/test/java/dev/hardwood/internal/predicate/BloomFilterPushDownTest.java
@@ -128,13 +128,31 @@ class BloomFilterPushDownTest {
     }
 
     @Test
-    void negativeZeroFloatIsNeverPrunedByBloom() {
-        // `price` stores +0.0; a query for -0.0f must still match it (-0.0f == +0.0f) even though the
-        // two have different IEEE-754 bits and therefore different bloom hashes. The ±0 carve-out
-        // keeps the row group rather than pruning on the raw-bit hash.
+    void negativeZeroFloatCanBePrunedByBloom() {
+        // `price` stores +0.0; Float.compare treats -0.0f as distinct, so its raw-bit bloom probe
+        // can prove the row group absent.
         FilterPredicate negZero = FilterPredicate.eq("price", -0.0f);
         assertThat(statisticsDrop(negZero)).isFalse();
-        assertThat(bloomDrop(negZero)).isFalse();
+        assertThat(bloomDrop(negZero)).isTrue();
+    }
+
+    @Test
+    void negativeZeroDoubleCanBePrunedByBloom() {
+        FilterPredicate negZero = FilterPredicate.eq("ratio", -0.0);
+        assertThat(statisticsDrop(negZero)).isFalse();
+        assertThat(bloomDrop(negZero)).isTrue();
+    }
+
+    @Test
+    void nanFloatIsNeverPrunedByBloom() {
+        FilterPredicate nan = FilterPredicate.eq("price", Float.NaN);
+        assertThat(bloomDrop(nan)).isFalse();
+    }
+
+    @Test
+    void nanDoubleIsNeverPrunedByBloom() {
+        FilterPredicate nan = FilterPredicate.eq("ratio", Double.NaN);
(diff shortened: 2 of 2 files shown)


iifawzi:
Hello @tomatotomata, it looks like you have misunderstood or haven't read #829. Can you elaborate on the changes applied here?

tomatotomata:
I reread #829 and you’re right that my first patch inverted the requested scope. I’ve amended the branch at da552f5: the zero guard is removed for FLOAT/DOUBLE probes, while the NaN carve-out remains because Float.compare/Double.compare treat NaN payloads as one matcher class while the bloom hash distinguishes raw payloads. I also corrected the focused zero/NaN assertions and updated the PR description. The branch is force-pushed and git diff --check passes. The local focused Maven test is still blocked by Java 25/Docker availability. Let me know what you think.

iifawzi:
> I reread #829 and you’re right that my first patch inverted the requested scope. I’ve amended the branch at [da552f5](https://github.com/hardwood-hq/hardwood/commit/da552f58536a269e4172b319c862cd4775bbf0b2): the zero guard is removed for FLOAT/DOUBLE probes, while the NaN carve-out remains because Float.compare/Double.compare treat NaN payloads as one matcher class while the bloom hash distinguishes raw payloads. I also corrected the focused zero/NaN assertions and updated the PR description. The branch is force-pushed and git diff --check passes. The local focused Maven test is still blocked by Java 25/Docker availability. Let me know what you think.

Thanks @tomatotomata. On the testing point, what's blocking it? Have you been able to verify all tests pass?

tomatotomata:
The blocker is environmental rather than a test failure: this checkout requires Java 25 (the repository `.java-version` is 25), while the host only has Temurin 17. The Maven wrapper is present, but I cannot run it successfully until Java 25 is available. Docker Desktop’s CLI is installed, but `docker info` cannot connect to the Linux engine, so Testcontainers cannot start either. I have not claimed the full suite passes; the only fresh local verification is `git diff --check` plus the focused code/test inspection. Once Java 25 and the Docker engine are available, `mvnw.cmd clean verify` is the right full check. Let me know what you think.

iifawzi:
am I talking with LLM bot @tomatotomata? looks like it's fully LLM-controlled, isn't it? Anything preventing proper repository setup for testing?

iifawzi:
Hi @tomatotomata thanks for your efforts. I'm going to close this PR for now, because while we accept LLM assisted contributions, we don't accept vibecoded contributions, without understanding, caring and owning the decisions which I assume wasn't the case in this PR. 

You're more than welcome to tackle any other issue in case you can ensure that. Thanks.

PR #835, opened 2026-08-05
#607 Parse SizeStatistics and NaN counts

### LINKED ISSUE #607: Parse SizeStatistics and NaN counts (unencoded byte size, level histograms, nan_count)

nitrogen404:
closes #607 
## Problem

Parquet 2.10 added `SizeStatistics` (`ColumnMetaData` field 16) and NaN counts (`Statistics` field 9, `ColumnIndex` field 8), plus the per-page level histograms and unencoded `BYTE_ARRAY` sizes in the page index (`ColumnIndex` 6/7, `OffsetIndex` 2). arrow-cpp writes them by default, so a large share of files carry them. All six fell through to `skipField` and were dropped.

They answer questions the reader currently pays to compute or cannot answer at all: a level histogram gives a chunk's null and empty-list counts from the footer, where the alternative is decoding the level stream; and NaN sits outside the min/max ordering, so `nan_count` is the only statistic that says anything about a floating-point chunk's NaN content which is why `StatisticsFilterSupport` refuses `ALWAYS_MATCHES` for FLOAT, FLOAT16 and DOUBLE today.

#613 removed the mis-numbered geospatial branch at `ColumnIndex` field 7 and left the parse to this change.

## Change

Read-only and additive — no read path consults the new fields, so no file reads differently than before.

- New public `SizeStatistics` record and `SizeStatisticsReader` wired into `ColumnMetaDataReader` case 16; `StatisticsReader` case 9; `ColumnIndexReader` cases 6/7/8; `OffsetIndexReader` case 2

- `ThriftCompactReader.readI64List()` centralizes the five list reads and gates on the **collection header's element type**, not just the field's type code. Fields 6/7/8 were previously skipped, so reading them is what creates the exposure: a `list<struct>` there the malformed shape #608 flagged — would otherwise decode as varints, leaving the cursor mid-struct and silently corrupting every later field.
- An omitted field reads back as `null`, distinct from present-but-empty. PyArrow writes the same column's repetition-level histogram as empty in `SizeStatistics` and absent in the `ColumnIndex`; and for `nanCount` only a recorded `0` proves a chunk holds no NaN.
- The `ColumnIndex` histograms keep their on-disk page-major layout (`pageCount × (maxLevel + 1)`); slicing needs the column's max level, which the metadata records don't carry.
- Each new component is appended to its record, so the canonical constructors change binary-incompatible additions to `dev.hardwood.metadata`, visible in the japicmp report.

Consuming the data stays out: the FLOAT/DOUBLE `ALWAYS_MATCHES` gap (#795) and the `dive` display are follow-ups.

## Testing

`./mvnw clean verify` — BUILD SUCCESS across all 13 modules.
`./mvnw -pl core test` — 1765 run, 0 failures, 0 errors, 0 skipped.

25 new tests across 5 classes. The reader tests hand-roll Thrift structs, covering each field present, absent, present-but-empty and mistyped. `size_statistics_test.parquet` (new, PyArrow 24.0.0 with the page index) asserts five of the six fields against writer-produced bytes; no available writer emits a NaN count, so `Statistics` field 9 and `ColumnIndex` field 8 have positive coverage in the unit tests only and the fixture pins them absent.

## Checklist


gunnarmorling:
Thanks a lot, @nitrogen404! Could you rebase this to current main to resolve the merge conflicts?

nitrogen404:
@gunnarmorling Done!

gunnarmorling (approved):
LGTM overall. Did a few amendments (most notably, replacing the `List<Long>` with `long[]` to avoid boxing overhead). Merged. Thanks a lot, @nitrogen404!

PR #836, opened 2026-08-05
#265 Fix wide-character alignment in print output

### LINKED ISSUE #265: Wide-character alignment in hardwood print (StreamedTable + transpose mode)
## Problem

Fixed for `inspect pages` in #263 via a display-width-aware renderer in `RowTable`. The `print` command has the same class of bug on two paths that weren't migrated:

1. **Normal table output** — `printTable` uses `StreamedTable` (sampling-based streaming renderer with `--max-width` truncation and cell wrapping). Its column-width math uses `String.length()` and padding uses `printf(\"%-Ns\")`, neither of which accounts for East Asian wide characters (CJK / Hangul / Kana / Fullwidth). A row whose cell contains e.g. `말도나도주` (5 code units, 10 terminal cells) will visually exceed its allocated column width, shifting the trailing pipe rightward.
2. **Transposed mode (`-tp`)** — `printTransposed` uses `AsciiTable.builder()` directly to render one mini 2-column table per row. Same `String.length()`-based sizing; same bug.

## Reproduction

Run `hardwood print -f <file-with-cjk>` on any Parquet file where at least one string column contains East Asian content. Rows with CJK values will have misaligned borders relative to rows with Latin-only content.

## Proposed fix

1. **StreamedTable**: change its width computation, column-width clamping, truncation, and padding to use `RowTable.displayWidth(String)` instead of `String.length()`. Can't delegate wholesale to `RowTable.renderTable` because StreamedTable streams rows and supports wrapping / truncation that `RowTable` doesn't. ~30 lines of changes.
2. **Transposed mode**: migrate from `AsciiTable.builder()` to `RowTable.renderTable`. Trivial.

## Acceptance

Output of `hardwood print` on a CJK-containing file has consistent visual column widths and aligned trailing pipes across all rows.

## Notes

- `inspect dictionary` is not affected — it prints one entry per line with no cross-entry alignment.
- `RowTable.displayWidth` covers CJK Unified + extensions, Kangxi, Hangul, Kana, Bopomofo, Fullwidth Forms.

mturac:
Fixes #265.

`hardwood print` used string length when it sized, wrapped, and truncated cells. East Asian wide characters could shift table borders, and truncation could split a Unicode code point.

This change:

- uses terminal display width for streamed output;
- renders transpose mode through `RowTable` so both paths share the same width handling;
- keeps wrapping and truncation on code-point boundaries.

## Verification

- `./mvnw clean verify` with Java 25 and Docker: passed
- focused `RowTable`, `StreamedTable`, and `PrintCommand` tests: 36 passed

## Checklist

- [x] Documentation is unchanged because this fixes existing rendering without changing a command or public API

### DIFF
--- a/cli/src/main/java/dev/hardwood/cli/command/PrintCommand.java
+++ b/cli/src/main/java/dev/hardwood/cli/command/PrintCommand.java
@@ -26,8 +26,6 @@ import org.aesh.command.invocation.CommandInvocation;
 import org.aesh.command.option.Mixin;
 import org.aesh.command.option.Option;
 
-import com.github.freva.asciitable.AsciiTable;
-
 import dev.hardwood.InputFile;
 import dev.hardwood.cli.internal.table.RowTable;
 import dev.hardwood.cli.internal.table.StreamedTable;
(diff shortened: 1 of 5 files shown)


gunnarmorling (approved):
LGTM, merging. Thanks a lot, @mturac!

gunnarmorling:
Also logged #839 as a follow-up, in case you're interested in picking up this one too.

PR #840, opened 2026-08-06
fix: floor streamed table columns at min-content width

### LINKED ISSUE #839: Floor `print --max-width` at the widest indivisible glyph (min-content width)
## Background

Follow-up to #265 / #836. Once `hardwood print` sizes columns by terminal display width, wide (2-cell) East Asian glyphs expose a degenerate case: a column capped narrower than a single glyph.

`StreamedTable` clamps each column to `--max-width` as a hard cap:

```java
widths[i] = Math.min(widths[i], maxWidth);
```

In **wrap** mode (no truncation, all data preserved) a single wide glyph cannot be wrapped below its own 2-cell width, so with `--max-width 1` a `가` overflows its column and the right border wobbles on that row:

```
+---+
| x |
+---+
| 가 |   <- one cell past the border
| 나 |
+---+
```

The force-progress guard in `StreamedTable.printRow` (the `end == start` branch) keeps this from looping forever, so it is correct but visually broken. Overflow is the current least-bad choice only because a hard cap leaves no alternative that avoids silent data loss.

## Proposal

Treat `--max-width` as a cap that floors at the **min-content width** — the widest single glyph the column must render — rather than a hard limit. Standard text-layout behavior: you cannot wrap a box below its widest unbreakable token.

```java
// widestGlyph(s) = max RowTable.charWidth(cp) over the code points of s
minWidths[i] = Math.max(minWidths[i], widestGlyph(cell));
...
widths[i] = Math.max(Math.min(widths[i], maxWidth), minWidths[i]);
```

`RowTable.charWidth(int)` already exists (added in #836), so this needs no new width logic. All-Latin columns have `widestGlyph == 1`, so their behavior is unchanged; a CJK column gets an effective floor of 2 and renders flush with the border.

**Semantic change:** `--max-width` gains an effective per-column floor — a column containing a 2-cell glyph can no longer be squeezed below 2. This is honest: a 1-cell column cannot faithfully show a 2-cell character.

## Caveat: sampled widths

Column widths are computed from a sample of rows. If a wide glyph appears **only** in an unsampled row while that column's sampled rows are all Latin, the floor stays 1 and that row still overflows. The `end == start` force-progress guard must therefore remain as a backstop — it stops being the primary path but still prevents a hang.

## Acceptance

- `hardwood print --max-width 1` on a CJK column produces aligned borders (no wobble) for wide glyphs present in the sampled rows.
- All-Latin columns are unaffected at any `--max-width`.
- The force-progress guard is retained and covered by a test that places the wide glyph in an **unsampled** row (`sampleSize < rows`).
- `StreamedTableTest.emitsWideCharacterInColumnNarrowerThanOneWideCell` is reworked: with the floor, its wide glyph is now sampled and should render aligned; move the guard coverage to the unsampled-row case.

slegarraga:
## Summary

Treats `--max-width` as a cap that floors at the widest indivisible glyph per column, so a 2-cell CJK/Hangul character cannot squeeze a column below its own width.

## Changes

- Computes a min-content width per column from headers and sampled rows using the existing `RowTable.charWidth` logic.
- Applies the floor after the `max-width` cap: `Math.max(Math.min(widths[i], maxWidth), minWidths[i])`.
- Reworks the narrow-wide-character test to assert aligned borders when the wide glyph is sampled.
- Adds coverage for the retained force-progress guard when the wide glyph appears only in an unsampled row.

## Verification

- Ran `./mvnw -Dtest=StreamedTableTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false test` under JDK 25; all tests pass.

Closes #839

### DIFF
--- a/cli/src/main/java/dev/hardwood/cli/command/PrintCommand.java
+++ b/cli/src/main/java/dev/hardwood/cli/command/PrintCommand.java
@@ -44,7 +44,7 @@ public class PrintCommand implements Command<CommandInvocation> {
     @Option(shortName = 's', name = "sample-size", defaultValue = "10", description = "Max number of lines used to auto-adjust the column width.")
     int sampleSize;
 
-    @Option(shortName = 'w', name = "max-width", defaultValue = "50", description = "Max width in characters of a column.")
+    @Option(shortName = 'w', name = "max-width", defaultValue = "50", description = "Max width of a column.")
     int maxWidth;
 
     @Option(shortName = 't', name = "truncate", hasValue = false, negatable = true, defaultValue = "true", description = "Should rows be truncated instead of wrapping on next line when too long.")
@@ -73,6 +73,7 @@ public class PrintCommand implements Command<CommandInvocation> {
         }
 
         try (ParquetFileReader reader = ParquetFileReader.open(inputFile)) {
+            validateMaxWidth();
             int rowLimit = RowLimits.parse(n);
             ColumnProjection projection = parseColumnProjection();
             FileSchema fileSchema = reader.getFileSchema();
@@ -98,6 +99,16 @@ public class PrintCommand implements Command<CommandInvocation> {
         }
 
         return CommandResult.SUCCESS;
+    }
+
+    /// Rejects widths below one cell. A column has to be at least one cell wide to
+    /// render anything at all, so a smaller value has no faithful rendering rather
+    /// than merely an ugly one.
+    private void validateMaxWidth() {
+        if (maxWidth < 1) {
+            throw new IllegalArgumentException(
+                    "Invalid value for option '-w': expected a positive integer, got '" + maxWidth + "'");
+        }
     }
 
     private void printTransposed(Stream<Object[]> stream, String[] headers, List<SchemaNode> fields, AtomicLong rowIndex) {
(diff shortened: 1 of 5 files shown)


gunnarmorling:
@slegarraga, I see you're creating dozens of PRs across many completely unrelated repos. What's your motivation for doing so?

slegarraga:
Thanks for asking, and happy to be transparent about it. I genuinely enjoy reading how different projects approach real problems, and small, reviewable contributions (docs, tests, and narrowly scoped fixes) are a useful way to give back while learning from those codebases. I try to keep each PR as small and verifiable as possible, and I don't expect maintainers to prioritize them if they don't fit the project's direction. If there is anything about this change that would make it more useful for Hardwood specifically, I'm happy to adjust it.

gunnarmorling:
Got it, thank you for the transparency!

The reason for asking is that we've seen a wave of AI-generated drive-by contributions recently and it's hard to stay on top of them (reviewing and helping to bring PRs up-to-standard essentially is an investment which needs weighing against other things).

This one was directionally right; I've pushed one more commit to fix up a few loose ends and ensure consistent behavior for `-w` for different character widths. Also reworded your commit message as per project conventions. And logged and linked a few related issues which came up during review. Any help with those welcome, too. Thanks again!

PR #848, opened 2026-08-08
#433 Add head and skip to column readers

### LINKED ISSUE #433: feat(reader): head() and firstRow() parity for ColumnReaderBuilder and ColumnReadersBuilder
## Motivation

`RowReaderBuilder` exposes `head(long)`, `firstRow(long)`, and `tail(long)` to bound the row range a reader emits. The columnar builders (`ColumnReaderBuilder` and `ColumnReadersBuilder`) only support `filter(FilterPredicate)` and `filter(RowGroupPredicate)` — there is no equivalent way to cap or seek by row count.

This forces consumers to maintain their own row counters and break out of `nextBatch()` loops manually, which:

- doesn't push the limit down to Hardwood (so on remote backends, leading/trailing pages are still fetched even though they will be discarded);
- duplicates logic that already exists for `RowReader`;
- is asymmetric in a way that's surprising — \"why does the row reader have head() but the columnar one doesn't?\"

A concrete example: Flink's vectorized integration handles checkpoint-resume by tracking `recordsToSkip` itself and discarding rows in the consumer loop. With `ColumnReader.firstRow(N)`, it could push the seek into Hardwood and skip the leading row groups' pages entirely.

## Proposal

Add to `ColumnReaderBuilder` and `ColumnReadersBuilder`:

```java
public Builder head(long maxRows);     // cap at the first maxRows rows
public Builder firstRow(long firstRow); // start from absolute row firstRow
```

Semantics match `RowReaderBuilder`:

- `head(N)` and `firstRow(N)` compose: `firstRow(F).head(N)` → window `[F, F+N)`.
- `firstRow(N) + filter(RowGroupPredicate)`: `firstRow` indexes into the *row-group-filtered* sequence — the same convention adopted for `RowReaderBuilder` in #431, so a caller doing split-aware reading can seek inside *its* split.
- `firstRow(N)` opens only the row group(s) needed to reach the start point (no leading I/O on remote backends), matching the existing `RowReader` behavior.

For `ColumnReadersBuilder`, the settings configure all leaves uniformly — there's no per-leaf knob to keep in sync.

## Out of scope

`tail(long)` is deferred. The `RowReader.tail(...)` path has nontrivial machinery (`tailRowGroups` plus the `canFastSkipAllRowGroups` page-mask gate for nested-v1 columns without an OffsetIndex). Replicating it for the columnar path means either parallel implementations or refactoring the tail logic into a shared helper, and the use case for \"tail of a single column / column projection\" is narrower than head/firstRow. File separately if a concrete demand surfaces.

## Tests

- `head(N)` caps reading at exactly N rows for both single-column and multi-column readers.
- `firstRow(N)` starts at the right row (resumes mid-row-group correctly, walks past intra-row-group residue).
- `firstRow(F).head(N)` produces the right `[F, F+N)` window.
- `firstRow(N) + filter(RowGroupPredicate)` indexes into the filtered sequence (parity with `RowReaderBuilder`).
- Edge cases: `firstRow(0)` is a no-op; `firstRow == totalRows` yields an empty reader; `firstRow > totalRows` throws.

## Docs

Update `docs/content/usage.md` to mention `head` and `firstRow` on the columnar builders alongside the existing `RowReader` coverage.

## Related

- #431 (RowGroupPredicate) — establishes the \"firstRow indexes into the filtered sequence\" convention.
- #381 (page-level skip via OffsetIndex when seeking with firstRow) — independent optimization that would also reduce within-row-group decode cost when these new entry points use `firstRow`.

Guflly:
Adds head() and skip() to both column reader builders. Unfiltered reads seek to the physical offset, while filtered reads apply OFFSET/LIMIT to matches. Projected columns share the same window.

## Checklist
- [x] Build via ./mvnw clean verify passes
- [x] If this PR adds or changes user-facing behaviour, docs/ has been updated

Closes #433

### DIFF
--- a/_designs/ROW_SELECTION_SEMANTICS.md
+++ b/_designs/ROW_SELECTION_SEMANTICS.md
@@ -1,14 +1,15 @@
 # Design: row selection semantics
 
-**Status: In progress.** Tracking issues: #538 (`head` + filter), #540 / #541
-(`skip` + filter), #542 (`tail` + filter), #543 (`firstRow` → `skip` rename).
+**Status: In progress.** Tracking issues: #433 (column-reader parity), #538 (`head` + filter),
+#540 / #541 (`skip` + filter), #542 (`tail` + filter), #543 (`firstRow` → `skip` rename).
 The reader-facing counterpart is `docs/content/concepts/row-selection.md`.
 
 ## Scope
 
-How the row-selection controls on `RowReaderBuilder` — `head`, `tail`, `skip` —
-relate to the two filtering mechanisms — the row-level `FilterPredicate` and the
-group-level `RowGroupPredicate` (e.g. `byteRange`). The goal is a single model
+How the row-selection controls on `RowReaderBuilder`, `ColumnReaderBuilder`, and
+`ColumnReadersBuilder` — `head`, `skip`, and the row-reader-only `tail` — relate to the two
+filtering mechanisms — the row-level `FilterPredicate` and the group-level
+`RowGroupPredicate` (e.g. `byteRange`). The goal is a single model
 under which the controls compose predictably, so a caller can reason about
 `skip(a).filter(p).head(b)` without knowing the file's row-group structure or
 the filter's selectivity.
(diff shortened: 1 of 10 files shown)


gunnarmorling:
@Guflly, I see you're creating dozens of PRs across many completely unrelated repos. What's your motivation for doing so?

Guflly:
I'm trying to build a stronger open-source contribution record. I've been picking issues I can reproduce, fixing them with tests, and following up on reviews.

gunnarmorling:
Thanks for replying. I'm gonna close this one, as we're not looking for contributions from accounts mass-submitting PRs to large numbers of unrelated repos. I've also clarified our contribution guide in this regard.

Note if building up an OSS contribution track record actually is your goal, I'd recommend a very different approach. Instead of sending LLM-based PRs to random repos, rather focus on a project you care about and contribute to it in a sustained way. All the best.

PR #858, opened 2026-08-08
#856 Expose ColumnMetaData.encoding_stats on the public metadata API 

### LINKED ISSUE #196: Dictionary-based filter pushdown
## Summary

When a column is dictionary-encoded, the dictionary can be read cheaply and used to eliminate row groups or pages where no dictionary entry matches the predicate. This is especially effective for EQ and IN predicates on low-to-medium cardinality columns.

parquet-java supports this via its `DictionaryFilter`. The dictionary is typically small and already present in the column chunk metadata.

## Approach

After row-group statistics filtering but before page-level filtering, read the dictionary page for dictionary-encoded columns and check if any dictionary entry satisfies the predicate. If not, the entire column chunk can be skipped.

### LINKED ISSUE #856: Expose ColumnMetaData.encoding_stats on the public metadata API
Part of #483 (`ColumnMetaData: encoding_stats (13)` and the `PageEncodingStats` struct).

`ColumnMetaData.encoding_stats` (Thrift field 13) records how many pages a writer produced per (page type, encoding) pair in a column chunk. Hardwood currently skips the field when parsing the footer, so it is invisible to callers.

Beyond mirroring the spec, this is the only metadata that proves a chunk's data pages are *entirely* dictionary-encoded, which readers need for dictionary-based row-group filtering (#196). It is also useful on its own for tooling that reports how a file was written.

iifawzi:
Hello, this PR resolves #856, that can be shipped on its own, and make it possible to close #196

The footer parser skipped Thrift field 13, so callers had no way to see how many pages a writer produced per (page type, encoding) pair. Beyond mirroring the spec, it is the only metadata that proves a chunk's data pages are entirely dictionary-encoded, which a reader needs before it can trust a dictionary page as a complete inventory of the chunk's values.

Surfacing it means the page-type enum becomes public, so it moves out of the internal `PageHeader` into `dev.hardwood.metadata`. It gains an `UNKNOWN` constant: `encoding_stats` is optional and purely informational, so a page type from a future release (or invalid one) must not make an otherwise readable file unreadable. A page header declaring such a type is still rejected, because that page has to be decoded and cannot be.

The writer does not emit the field yet.

## Checklist

### DIFF
--- a/cli/src/main/java/dev/hardwood/cli/command/InspectPagesCommand.java
+++ b/cli/src/main/java/dev/hardwood/cli/command/InspectPagesCommand.java
@@ -35,6 +35,7 @@ import dev.hardwood.metadata.ColumnIndex;
 import dev.hardwood.metadata.ColumnMetaData;
 import dev.hardwood.metadata.FileMetaData;
 import dev.hardwood.metadata.OffsetIndex;
+import dev.hardwood.metadata.PageType;
 import dev.hardwood.metadata.RowGroup;
 import dev.hardwood.metadata.Statistics;
 import dev.hardwood.reader.ParquetFileReader;
@@ -403,7 +404,7 @@ public class InspectPagesCommand implements Command<CommandInvocation> {
             PageHeader header = PageHeaderReader.read(headerReader);
             int headerSize = headerReader.getBytesRead();
 
-            boolean isDictionary = header.type() == PageHeader.PageType.DICTIONARY_PAGE;
+            boolean isDictionary = header.type() == PageType.DICTIONARY_PAGE;
             String label = isDictionary ? "dict" : String.valueOf(pageIndex);
             Long firstRowIndex = null;
             Statistics inlineStats = null;
@@ -425,7 +426,7 @@ public class InspectPagesCommand implements Command<CommandInvocation> {
                     inlineStats
             ));
 
-            if (header.type() == PageHeader.PageType.DATA_PAGE || header.type() == PageHeader.PageType.DATA_PAGE_V2) {
+            if (header.type() == PageType.DATA_PAGE || header.type() == PageType.DATA_PAGE_V2) {
                 valuesRead += numValues(header);
                 pageIndex++;
                 if (valuesRead >= cmd.numValues()) {
@@ -449,16 +450,17 @@ public class InspectPagesCommand implements Command<CommandInvocation> {
                 DataPageHeaderV2 dp = header.dataPageHeaderV2();
                 yield dp != null ? dp.statistics() : null;
             }
-            case DICTIONARY_PAGE, INDEX_PAGE -> null;
+            case DICTIONARY_PAGE, INDEX_PAGE, UNKNOWN -> null;
         };
     }
 
-    private static String shortType(PageHeader.PageType type) {
+    private static String shortType(PageType type) {
         return switch (type) {
             case DATA_PAGE -> "DATA";
             case DATA_PAGE_V2 -> "DATA_V2";
             case DICTIONARY_PAGE -> "DICT";
             case INDEX_PAGE -> "INDEX";
+            case UNKNOWN -> "UNKNOWN";
         };
     }
 
@@ -467,7 +469,7 @@ public class InspectPagesCommand implements Command<CommandInvocation> {
             case DATA_PAGE -> shortEncoding(header.dataPageHeader().encoding().name());
             case DATA_PAGE_V2 -> shortEncoding(header.dataPageHeaderV2().encoding().name());
             case DICTIONARY_PAGE -> shortEncoding(header.dictionaryPageHeader().encoding().name());
-            case INDEX_PAGE -> "N/A";
+            case INDEX_PAGE, UNKNOWN -> "N/A";
         };
     }
 
@@ -484,7 +486,7 @@ public class InspectPagesCommand implements Command<CommandInvocation> {
             case DATA_PAGE -> header.dataPageHeader().numValues();
             case DATA_PAGE_V2 -> header.dataPageHeaderV2().numValues();
             case DICTIONARY_PAGE -> header.dictionaryPageHeader().numValues();
-            case INDEX_PAGE -> 0;
+            case INDEX_PAGE, UNKNOWN -> 0;
         };
     }
 }
--- a/core/src/main/java/dev/hardwood/internal/writer/ColumnChunkBuffer.java
+++ b/core/src/main/java/dev/hardwood/internal/writer/ColumnChunkBuffer.java
@@ -204,7 +204,9 @@ final class ColumnChunkBuffer implements RecordShredder.LevelSink {
                 values.statistics(),
                 null,
                 null,
-                null);
+                null,
+                // The writer does not emit encoding_stats yet.
+                List.of());
     }
 
     /// Seals the current page as a dictionary-indexed page and switches the rest of the chunk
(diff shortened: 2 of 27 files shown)


core/src/main/java/dev/hardwood/metadata/PageType.java
```diff
@@ -0,0 +1,32 @@
...
+package dev.hardwood.metadata;
+
+/// The kind of page stored in a column chunk.
+///
+/// @see <a href="https://parquet.apache.org/docs/file-format/data-pages/">File Format – Data Pages</a>
+/// @see <a href="https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift">parquet.thrift</a>
+public enum PageType {
+    /// A v1 data page: the repetition levels, definition levels and values are compressed
+    /// together as a single payload.
+    DATA_PAGE,
```
gunnarmorling, line 17: Should we name it `DATA_PAGE_V1`? Or does symmetry with the spec suggest just `DATA_PAGE`?

cli/src/main/java/dev/hardwood/cli/command/InspectPagesCommand.java
```diff
@@ -403,7 +404,7 @@ private List<PageInfo> collectPageHeaders(ColumnMetaData cmd, InputFile inputFil
             PageHeader header = PageHeaderReader.read(headerReader);
             int headerSize = headerReader.getBytesRead();
 
-            boolean isDictionary = header.type() == PageHeader.PageType.DICTIONARY_PAGE;
+            boolean isDictionary = header.type() == PageType.DICTIONARY_PAGE;
```
gunnarmorling, line 407: I realize it's a bit of scope creep, but given you're touching the CLI already, could we expose that metadata also in the CLI (both non-interactive--probably via `hardwood footer`--and dive TUI)?

core/src/main/java/dev/hardwood/metadata/ColumnMetaData.java
```diff
@@ -42,5 +43,6 @@ public record ColumnMetaData(
         Statistics statistics,
         GeospatialStatistics geospatialStatistics,
         Long bloomFilterOffset,
-        Integer bloomFilterLength) {
+        Integer bloomFilterLength,
```
gunnarmorling, line 46: Strictly speaking, this is a breaking API change, but I think we should accept constructor changes like this, as the metadata types are not meant to be instantiated by user code. It will show up in japicmp but we can ignore it.

gunnarmorling (approved):
Nice one, thanks a lot, @iifawzi! LGTM overall. A few minor questions/suggestions inline. Also pasting review output from Claude verbatim below, apply/ignore as you see fit.

--

## Decisions

- **Q:** How strict should `encoding_stats` parsing be? Today a *field*-level type mismatch is skipped (`nonListFieldIsSkipped`) but a `list<i32>` element type or an entry missing `count` aborts the entire footer read — so one malformed informational entry makes an otherwise readable file unreadable, which contradicts the PR's own stated principle.
  - [x] **A.** Be lenient throughout: `reader.skipField(0x09)` skips a wrongly-typed list safely (it re-reads the list header and skips `size` elements by type), and an entry missing a required field is dropped rather than thrown. `encoding_stats` then never blocks a read, matching how the field-level mismatch is already handled.
  - [ ] **B.** Be strict throughout: also fail on a field-level type mismatch, so any footer that deviates from `parquet.thrift` at field 13 is reported rather than silently ignored.
  - [ ] **C.** Keep as-is (lenient on field type, strict on element type and required fields).
  - **Rec:** A — the element-type check exists to avoid *misreading* bytes, and `skipField` already solves that without aborting; leniency also keeps the behaviour consistent with `Encoding.UNKNOWN` / `PageType.UNKNOWN` tolerance one level down. Whichever is picked, the tests already pin the current split, so they need updating with it.

## Implementation semantics

- [ ] `PageEncodingStatsReader.read` presizes `new ArrayList<>(listHeader.size())` from an unvalidated Thrift list size; a footer whose size varint casts to a negative `int` throws `IllegalArgumentException: Illegal Capacity`, which escapes `ParquetMetadataReader`'s `catch (IOException)` so the caller gets an unchecked exception with no file name (verified with the byte sequence `D9 FC 80 80 80 80 08 00`), and a large positive size allocates gigabytes before a single element is read. Drop the presize (the read loop then fails with a clean `EOFException`) or bound the size against the reader's remaining bytes, and add a test for it — `readNonNegativeI32` exists precisely for this class of adversarial-metadata failure. Note `KeyValueMetadataReader` has the same shape today, while `RowGroupReader` / `ColumnMetaDataReader` correctly do not presize.

## Nits

- [x] `PageFormatProbe.firstDataPageType` — the parameter continuation line is left over-indented now that the return type shortened from `PageHeader.PageType` to `PageType`; same in `PageFormatProbeTest`, where `PageType type =` and its initializer now fit on one line.
- [x] `PageEncodingStatsReader.read` returns `List.copyOf(result)` on a list it just built and never leaks, copying twice; the sibling `KeyValueMetadataReader` uses `Collections.unmodifiableMap` for the same job.

core/src/main/java/dev/hardwood/metadata/PageType.java
iifawzi: Yess, I'm kinda leaning toward matching spec naming https://github.com/apache/parquet-format/blob/2076361bb64e2de9ca6a8d06eda025a6fa4e9df6/src/main/thrift/parquet.thrift#L682-L687

gunnarmorling: Makes sense!

core/src/main/java/dev/hardwood/metadata/ColumnMetaData.java
iifawzi: We have no other option I believe, it's breaking indeed, but your rationale makes sense to me

iifawzi:
> PageEncodingStatsReader.read presizes new ArrayList<>(listHeader.size()) from an unvalidated Thrift list size; a footer whose size varint casts to a negative int throws IllegalArgumentException: Illegal Capacity, which escapes ParquetMetadataReader's catch (IOException) so the caller gets an unchecked exception with no file name (verified with the byte sequence D9 FC 80 80 80 80 08 00), and a large positive size allocates gigabytes before a single element is read. Drop the presize (the read loop then fails with a clean EOFException) or bound the size against the reader's remaining bytes, and add a test for it — readNonNegativeI32 exists precisely for this class of adversarial-metadata failure. Note KeyValueMetadataReader has the same shape today, while RowGroupReader / ColumnMetaDataReader correctly do not presize.


Thanks for taking a look @gunnarmorling, now that I have seen https://github.com/hardwood-hq/hardwood/issues/860 logged, perhaps we can check it there already. The rest of the comments have been resolved, I just created to tackle the exposure to the CLI https://github.com/hardwood-hq/hardwood/issues/861

core/src/main/java/dev/hardwood/internal/writer/ColumnChunkBuffer.java
```diff
@@ -204,7 +204,9 @@ ColumnMetaData flushTo(OutputFile out, ColumnSchema column, long chunkStartOffse
                 values.statistics(),
                 null,
                 null,
-                null);
+                null,
+                // The writer does not emit encoding_stats yet.
+                List.of());
```
iifawzi, line 209: well, I forgot to ask you. does it sound good @gunnarmorling, I assume it's not handled for now, should we create a separate issue to track?

gunnarmorling: Could you add this to the [writer delivery plan](https://github.com/hardwood-hq/hardwood/blob/main/_designs/WRITER_SUPPORT.md#delivery-plan), probably somewhere around stage 16?

iifawzi: Will add it to existing open PR, thanks for pointing out

PR #862, opened 2026-08-08
#388 Skip non-actionable data preview modal fields

### LINKED ISSUE #388: cli: go up/down through non actionable fields in data preview modal
Generally, the stance has been that we'd skip non-actionable (here: expansible/collapseable) lines, but this is not the case here.

hurshh:
Hello, this PR resolves #388 by making the cursor in the data preview row modal represent only actionable fields.

Previously, scalar fields displayed a selection cursor even though pressing
Enter on them had no effect. The modal now places the cursor on the first
expandable field and Up/Down navigation skips fields that cannot be expanded.

A regression test using `list_basic_test.parquet` verifies the initial cursor
position, navigation between expandable fields, and behavior at the navigation
boundaries.

Closes #388.

## Checklist

- [x] The commit message is prefixed with the issue key ("#388")

### DIFF
--- a/_designs/INTERACTIVE_DIVE_TUI.md
+++ b/_designs/INTERACTIVE_DIVE_TUI.md
@@ -222,12 +222,13 @@ Projected rows read via `RowReader`. Per-frame the screen knows the viewport hei
 and shows that many rows. `←` / `→` scrolls the visible column window with the row
 number column pinned; `PgDn` / `PgUp` flips pages forward (and backward via cursor
 re-creation). The title shows `rows X-Y of T · cols A-B of C · physical|logical`.
-`Enter` opens a row-detail modal with one line per field. When the modal body fits,
-`↑` / `↓` move between actionable fields; when it overflows, they move one rendered
-line at a time so all content remains reachable. `Enter` toggles inline expansion
-of the focused field's full value, and `▶` marks actionable field headers rather
-than scalar lines selected only for scrolling (the cursor is hidden when no field
-is expandable and content fits).
+`Enter` opens a row-detail modal with one line per field. `↑` / `↓` step between the
+fields `Enter` can expand — a field is expandable when its full value does not
+already fit the collapsed line, measured in display cells — and `PgDn` / `PgUp`
+scroll the body, so fields the cursor skips over are still readable. The modal opens
+with the cursor on the first expandable field, scrolled into view, and `▶` marks it;
+a record with nothing to expand shows no cursor at all. `Enter` toggles inline
+expansion of the focused field's full value.
 `e` / `c` expand / collapse all fields. `t` toggles logical types both in-table
 and in the modal. Long cell values get `…` indicators sized to the actual rendered
 column width.
--- a/_designs/INTERACTIVE_DIVE_TUI.md
+++ b/_designs/INTERACTIVE_DIVE_TUI.md
@@ -222,9 +222,12 @@ Projected rows read via `RowReader`. Per-frame the screen knows the viewport hei
 and shows that many rows. `←` / `→` scrolls the visible column window with the row
 number column pinned; `PgDn` / `PgUp` flips pages forward (and backward via cursor
 re-creation). The title shows `rows X-Y of T · cols A-B of C · physical|logical`.
-`Enter` opens a row-detail modal with one line per field; in the modal `↑` / `↓`
-moves the cursor across fields, `Enter` toggles inline expansion of the focused
-field's full value (cursor is hidden when no field is expandable and content fits).
+`Enter` opens a row-detail modal with one line per field. When the modal body fits,
+`↑` / `↓` move between actionable fields; when it overflows, they move one rendered
+line at a time so all content remains reachable. `Enter` toggles inline expansion
+of the focused field's full value, and `▶` marks actionable field headers rather
+than scalar lines selected only for scrolling (the cursor is hidden when no field
+is expandable and content fits).
 `e` / `c` expand / collapse all fields. `t` toggles logical types both in-table
 and in the modal. Long cell values get `…` indicators sized to the actual rendered
 column width.
--- a/cli/src/main/java/dev/hardwood/cli/dive/ScreenState.java
+++ b/cli/src/main/java/dev/hardwood/cli/dive/ScreenState.java
@@ -216,8 +216,10 @@ public sealed interface ScreenState {
     /// modal is currently open (-1 when closed). `logicalTypes` controls
     /// whether values render via their logical type (default) or the raw
     /// physical-type form — toggled with `t`. Inside the record modal,
-    /// `modalCursorLine` is the per-line cursor (collapsed fields = 1 line,
-    /// expanded fields = N lines); `expandedColumns` is the set of fields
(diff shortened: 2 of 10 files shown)


gunnarmorling:
Hey @hurshh, thanks a lot!

I've asked Claude to review this, and it came back with the findings below. Could you check those out? Also, could you rework the commits using an email address linked to your GH account (there should be no octocat icon shown as the author next to the commit messages).

_(via Claude Code)_

Thanks for picking this up! The problem in #388 is real — on `main` the modal cursor does step onto scalar fields where `Enter` is a visual no-op — and skipping is the right direction. But I reproduced two regressions against `main` that make modal content permanently unreachable, so this isn't ready to merge as-is.

## Blockers

**1. Fields below the last expandable field become unreachable once the record overflows the viewport.**

In `renderRecordModal` the scroll offset is derived solely from the cursor:

```java
int scroll = Math.max(0, Math.min(totalLines - viewport,
        Math.max(0, cursorLine - viewport / 2)));
```

So pinning the cursor to expandable fields pins the scroll too. Verified with a 30-field record (field 0 a list, fields 1–29 scalars) rendered at 100×24: on `main`, 29 `↓` presses walk the cursor to line 29 and `f29` renders; on this branch `↓` is never handled (0 moves, cursor stays at 0) and `f18`–`f29` cannot be reached — while the hint still reads `↓ 12 more lines · ↑↓ navigate`.

This is the case you asked about on #388. The PR handles the *no-expandable-field-at-all* variant (the `±1` fallback in `previousNavigableLine` / `nextNavigableLine`), but not the mixed one.

**2. `isExpandableField` uses the wrong definition of "actionable".**

It classifies a field as actionable only when its expanded form contains `\n`. But a long *single-line* scalar is actionable too: collapsed it renders via `truncate(value, valueBudget)`, expanded via `Strings.hardWrap(fullValue, valueBudget)` — which reveals the `…`-truncated tail.

Verified with a record of `tags` (list) + `note` (135-char string) at 60 columns. On `main`, `↓` then `Enter` wraps `note` across four lines and shows its tail. On this branch `↓` returns `false`, the cursor never reaches `note`, and its truncated text is permanently unreachable.

We already define actionability this way elsewhere — `_designs/INTERACTIVE_DIVE_TUI.md` §10 (Column index): *"`Enter` opens a modal with the full untruncated min / max only when at least one of them is actually truncated in the table cell"*.

## The underlying design question

Both issues share a root cause: actionability (*is this field truncated?*) and reachability (*does the record overflow?*) depend on render-time geometry — `valueBudget` and the modal viewport height, both derived from the terminal size — which `handleModal` has no access to. Options:

- **A.** Plumb the modal geometry into the state layer (the `Keys.hasObservedViewport()` / `Keys.viewportStride()` channel already used for page sizing is the precedent), then implement both rules correctly: skip only genuinely non-actionable fields, *and* keep line-by-line stepping whenever the record overflows.
- **B.** Keep the state layer geometry-free: narrow this PR to initial cursor placement only (open on the first expandable field) and leave `↑`/`↓` line-by-line. No unreachable content, but only half of #388.
- **C.** Decouple scrolling from the cursor — give the modal its own scroll offset (`PgUp`/`PgDn`, or `↑`/`↓` scrolling once past the last actionable field) — so the cursor is free to skip without stranding content.

I'd lean **A**: it's what #388 actually asks for, and the geometry channel it needs already exists. **B** is a reasonable smaller scope if you'd rather not touch the plumbing here.

## Implementation

- [ ] `withModalRow` assigns `firstExpandableField(s, modalRow)` — a **field** index — into the `modalCursorLine` **line** slot. It's correct today only because `expandedColumns` is always empty when a modal opens (both the close path and `loadPage` reset it to `Set.of()`), so the two indices coincide. Opening a modal with any field pre-expanded would silently misplace the cursor. `firstLineForField(s, s.expandedColumns(), field)` already exists and is the index-correct call.
- [ ] The hint's `canNavigate = showCursor && totalLines > 1` still prints "↑↓ navigate" for a record with exactly one expandable field and no overflow, where both keys are now no-ops. Pre-PR the hint was accurate because navigation always moved.

## Documentation

- [ ] `_designs/INTERACTIVE_DIVE_TUI.md` §12 (Data preview) still says *"in the modal `↑` / `↓` moves the cursor across fields"*. The design doc describes the intended end state, so it needs updating alongside whichever option above we land on.

## Tests

- [ ] The `!hasExpandableField(state)` fallback — scalar-only records retaining line-by-line navigation, the behaviour you explicitly designed and JavaDoc'd on both helpers — has no test.
- [ ] No test covers a record that has an expandable field *and* overflows the viewport (Blocker 1). `list_basic_test.parquet` happens to place both list fields last, so the added test can't catch it.
- [ ] No test covers a long single-line scalar alongside an expandable field (Blocker 2).

For reference, the existing `DiveStateTest` / `DiveRenderTest` (139 tests) pass on this branch — nothing in the suite currently guards these paths, which is why both regressions slipped through.

hurshh:
Thanks for the feedback! I’ll review and address the findings

hurshh:
Thanks for the detailed feedback! I’ve addressed the findings and force-pushed the rebased commits.

I went with approach A and shared the rendered modal geometry with the key handlers. The updated behavior is:

- When the modal fits, `↑`/`↓` skip fields that have no additional content to reveal.
- When the modal overflows, navigation moves one rendered line at a time so all content remains reachable.
- Long single-line values are actionable when their collapsed representation is truncated.
- The initial expandable field is converted to its rendered line index using `firstLineForField(...)`.
- The `↑↓ navigate` hint is shown only when navigation can actually move.
- Scalar fields selected for overflow scrolling remain highlighted without showing the `▶` action marker.

I also added regression coverage for:

- scalar-only overflow;
- mixed expandable/scalar overflow;
- long truncated single-line values;
- navigation-hint accuracy;
- actionable versus non-actionable field markers.

The design and CLI documentation have been updated as well. The branch is rebased onto the latest `main`, the commits now use my GitHub-linked email address, and `./mvnw clean verify` passes locally.

gunnarmorling:
Hey @hurshh, thanks a lot for the update! This looks good overall now.

I've pushed one more commit to adjust a few things after some more testing: ↑/↓ now navigate between actionable (expandable) rows, as is the case elsewhere in dive, whereas PageUp / PageDown can be used for scrolling, for instance allowing you to page through all data when having a node expanded which spans multiple screens.

If you're looking for more dive (or other) issues to work on, let me know, there's plenty to do :)

hurshh:
Thanks for the explanation; it makes sense! I really like the project and would love to keep contributing and help shape it as it grows. So I’d be happy to pick up more issues in dive or elsewhere in the project as I get more familiar with the codebase :)

gunnarmorling:
Awesome! If anything piques your interest and there's any questions, just tag me on that issue and we'll get it sorted out. Thank you!

gunnarmorling:
On this particular one, after some more testing, I ended up reworking this some more, see PR #1010. So the cursor is stopping at every row now (apart from titles and empty rows), which felt better after all, and there's consistent highlighting of the current row and actionable rows.

hurshh:
Yeah, that makes sense. I like the consistent row highlighting too; it makes the navigation behaviour look much smoother.

PR #871, opened 2026-08-09
#863 Preserve empty rows in wrap mode

### LINKED ISSUE #863: print --no-truncate silently drops rows whose cells are all empty
## Problem

In wrap mode (`--no-truncate`), a row is omitted from the output entirely when every one of its cells renders as an empty string. No warning, no blank line — the row simply is not there. The default (truncate) mode renders the same row correctly.

This is silent data loss: the row count in the output does not match the row count in the file.

## Reproduction

A Parquet file with a single string column holding `alpha`, `""`, `gamma`:

```
$ hardwood print -f empty_string.parquet          $ hardwood print -f empty_string.parquet --no-truncate
+-------+                                         +-------+
| s     |                                         | s     |
+-------+                                         +-------+
| alpha |                                         | alpha |
|       |   <- row 2                              | gamma |   <- row 2 is gone
| gamma |                                         +-------+
+-------+
```

Three rows in, two rows out.

## Root cause

`StreamedTable.printRow`, wrap branch. Each cell is wrapped by

```java
for (int start = 0; start < cell.length();) { ... lines.add(...); }
```

which produces zero lines for an empty cell. `maxLines` is the max over the row s cells, so a row of all-empty cells has `maxLines == 0` and the emitting loop

```java
for (int line = 0; line < maxLines; line++) { ... }
```

never runs. The truncate branch is unaffected because it prints exactly one line per row unconditionally.

Reachable from the CLI whenever every projected column of a row is an empty string — a single empty-string column, or `-c` narrowed to one.

## Expected

A row whose cells are all empty renders as one line of padding, matching truncate mode.

## Acceptance

- `hardwood print --no-truncate` emits exactly as many row blocks as the file has rows, for any combination of empty and non-empty cells.
- A regression test in `StreamedTableTest` covering an all-empty row in wrap mode, and one covering a row that is empty in some columns but not others.

## Notes

Pre-existing; unrelated to the `--max-width` work in #839 / #840, though it lives in the same renderer. Verified present on `main` and on the #840 branch alike.

alloutflo:
Fixes #863

In wrap mode, empty cells produced no rendered lines. When every cell in a row was empty, the row-wide line count therefore stayed at zero and the emitting loop skipped the row entirely.

This change represents an empty cell as one blank wrapped line. Non-empty wrapping is unchanged, while an all-empty row now produces the same single padded row block as truncate mode.

Regression coverage includes both an all-empty multi-column row and a mixed row whose non-empty cell wraps across multiple lines.

## Verification

- `StreamedTableTest`: 17 tests passed
- All non-S3, non-native CLI tests: 339 tests passed
- Full reactor compilation/package/license checks passed with tests skipped
- `./mvnw clean verify` reached the existing S3 Testcontainers suite; local completion is blocked only because this machine has no Docker daemon
- GitHub created the PR Build check suite, but it is currently awaiting the repository's first-time-contributor workflow approval (`action_required`)

## Checklist

- [ ] Build via `./mvnw clean verify` passes (awaiting maintainer approval of the fork workflow)
- [x] The commit message is prefixed with the issue key (`#863`)
- [x] No documentation update is needed; this restores row preservation without changing CLI options or documented APIs

### DIFF
--- a/cli/src/main/java/dev/hardwood/cli/internal/table/StreamedTable.java
+++ b/cli/src/main/java/dev/hardwood/cli/internal/table/StreamedTable.java
@@ -148,6 +148,9 @@ public class StreamedTable {
                 cell = "";
             }
             List<String> lines = new ArrayList<>();
+            if (cell.isEmpty()) {
+                lines.add("");
+            }
             for (int start = 0; start < cell.length();) {
                 int end = displayPrefixEnd(cell, start, widths[i]);
                 if (end == start) {
--- a/cli/src/test/java/dev/hardwood/cli/internal/table/StreamedTableTest.java
+++ b/cli/src/test/java/dev/hardwood/cli/internal/table/StreamedTableTest.java
@@ -46,6 +46,41 @@ class StreamedTableTest {
     }
 
     @Test
+    void rendersAllEmptyRowsWhenWrapping() {
+        String output = render(
+                List.<String[]>of(new String[]{"", ""}),
+                new String[]{"left", "right"},
+                10,
+                false);
+
+        assertThat(output).isEqualTo("""
+                +------+-------+
+                | left | right |
+                +------+-------+
+                |      |       |
+                +------+-------+""");
+        assertEqualDisplayWidths(output);
+    }
+
+    @Test
+    void wrapsRowsWithEmptyAndNonEmptyCells() {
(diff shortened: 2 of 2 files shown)


gunnarmorling:
LGTM, merging. Thx, @alloutflo!

PR #880, opened 2026-08-09
#875 Escape names in Avro schema output

### LINKED ISSUE #875: cli: `schema -F AVRO` and `-F PROTO` emit invalid output for names outside the target grammar
## Problem

`hardwood schema -F AVRO` and `-F PROTO` build their output by string concatenation and interpolate Parquet names straight into positions that have their own grammar. Parquet permits arbitrary column names; neither format does.

### `-F AVRO`

Names are not escaped (`cli/src/main/java/dev/hardwood/cli/command/SchemaCommand.java:100`):

```java
sb.append(p).append("{ \"name\": \"").append(node.name()).append("\", \"type\": ");
```

A column named `say "hi"` produces:

```json
{ "name": "say "hi"", "type": ["null", "string"], "default": null }
```

which is not parseable JSON. The record name goes through the same path (`SchemaCommand.java:81`, via `capitalize`).

Escaping alone is not sufficient: Avro restricts record and field names to `[A-Za-z_][A-Za-z0-9_]*`, so an escaped name still yields a document that parses as JSON and is then rejected by every Avro parser.

### `-F PROTO`

The same names reach identifier positions, where there is no escaping mechanism at all — only rewriting. A column named `total (usd)` produces:

```proto
optional double total (usd) = 1;
```

which no `protoc` accepts. Message names go through the same path via `capitalize`.

## Reproduction

Write a Parquet file with a column whose name contains a double quote or a space, then:

- `hardwood schema -f FILE -F AVRO` piped to any JSON parser, or parsed with `new Schema.Parser().parse(...)`.
- `hardwood schema -f FILE -F PROTO` piped to `protoc`.

## Fix

Map names onto the target grammar — Avro names and proto identifiers share `[A-Za-z_][A-Za-z0-9_]*` — rather than only escaping them. The mapping is not injective, so names that collide within one record/message need disambiguating, and the original Parquet name should stay visible in the output (Avro `doc`, proto comment) rather than being silently dropped.

JSON escaping is still needed for the strings that do carry the original name. The CLI already has a `jsonEscape` helper in `ConvertCommand.java:241`; per the DRY guidance in `CLAUDE.md` this should move somewhere shared rather than be duplicated a second time.

Per `CLAUDE.md`, start with a failing test that asserts the emitted schema parses.

fzlzjerry:
Fixes #875

## Summary

- Move the existing JSON string escaping from `ConvertCommand` into a shared CLI utility.
- Escape both record and field names in `schema -F AVRO` output.
- Add an end-to-end regression test using record and field names that contain quotes and backslashes.

## Root cause

The AVRO formatter interpolated Parquet names directly into JSON string literals. Quotes and backslashes in otherwise valid Parquet names therefore produced JSON that could not be parsed.

The regression test writes a Parquet file with affected names, parses the CLI output with Avro's parser with name validation disabled for arbitrary Parquet names, and verifies that both names round-trip unchanged.

## Validation

- `./mvnw -ntp -B -pl cli -Dtest="*Test,!*S3CommandTest" test` (340 tests)
- `./mvnw process-sources`
- `./mvnw clean verify -DskipTests`

## Checklist

- [x] Documentation is not required because this fixes invalid output without changing the documented CLI interface

### DIFF
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -244,9 +244,10 @@
       <artifactId>hardwood-test-support</artifactId>
       <scope>test</scope>
     </dependency>
+    <!-- Parses the emitted Avro schema in SchemaCommandTest -->
     <dependency>
-      <groupId>dev.hardwood</groupId>
-      <artifactId>hardwood-avro</artifactId>
+      <groupId>org.apache.avro</groupId>
+      <artifactId>avro</artifactId>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -244,6 +244,11 @@
       <artifactId>hardwood-test-support</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>dev.hardwood</groupId>
+      <artifactId>hardwood-avro</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <profiles>
--- a/cli/src/main/java/dev/hardwood/cli/command/ConvertCommand.java
+++ b/cli/src/main/java/dev/hardwood/cli/command/ConvertCommand.java
@@ -21,6 +21,7 @@ import org.aesh.command.option.Mixin;
 import org.aesh.command.option.Option;
 
 import dev.hardwood.InputFile;
+import dev.hardwood.cli.internal.JsonStrings;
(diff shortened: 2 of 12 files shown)


gunnarmorling:
Thx, @fzlzjerry. Pushed two more commits to ensure valid Avro field names and apply the same transformation for protobuf output.

PR #882, opened 2026-08-09
cli: align footer output columns

### LINKED ISSUE #874: cli: `hardwood footer` magic-byte lines misaligned by one column
## Problem

`hardwood footer` pads its label column inconsistently (`cli/src/main/java/dev/hardwood/cli/command/FooterCommand.java:72-76`):

```java
System.out.println("File Size:     " + fileSize + " bytes");
System.out.println("Footer Offset: " + footerOffset + " bytes");
System.out.println("Footer Length: " + footerLength + " bytes");
System.out.println("Leading Magic:  " + new String(leadingMagic, StandardCharsets.US_ASCII));
System.out.println("Trailing Magic: " + new String(trailingMagic, StandardCharsets.US_ASCII));
```

The first three values start at column 16; the two magic values start at column 17, because `"Leading Magic:"` is followed by two spaces rather than one. Output:

```
File Size:     1234 bytes
Footer Offset: 1100 bytes
Footer Length: 126 bytes
Leading Magic:  PAR1
Trailing Magic: PAR1
```

## Fix

Pad all five labels to a common width so the values line up. `Trailing Magic:` is the longest label, so that sets the column.

nightcityblade:
## Summary

- pad the first three `hardwood footer` labels to the same value column as the magic-byte labels
- update the command contract test and CLI skill output example

Fixes #874

## Test results

- `./mvnw process-sources` — passed
- `./mvnw -pl cli -am -Dtest=FooterCommandTest -Dsurefire.failIfNoSpecifiedTests=false test` — 4 passed
- `./mvnw clean verify` — reached the S3 module; 124 S3 tests passed and 5 Testcontainers tests could not start because the local Docker daemon is unavailable
- `./mvnw -pl cli test` — 339 tests passed; 64 S3-backed tests could not start for the same Docker limitation

## Checklist

- [ ] Build via `./mvnw clean verify` passes (Docker daemon unavailable locally)
- [ ] If this PR adds or changes user-facing behaviour, `docs/` has been updated (no footer output example exists under `docs/`; `skills/hardwood-cli/SKILL.md` was updated)

### DIFF
--- a/cli/src/main/java/dev/hardwood/cli/command/FooterCommand.java
+++ b/cli/src/main/java/dev/hardwood/cli/command/FooterCommand.java
@@ -69,9 +69,9 @@ public class FooterCommand implements Command<CommandInvocation> {
 
             long footerOffset = fileSize - TRAILER_SIZE - footerLength;
 
-            System.out.println("File Size:     " + fileSize + " bytes");
-            System.out.println("Footer Offset: " + footerOffset + " bytes");
-            System.out.println("Footer Length: " + footerLength + " bytes");
+            System.out.println("File Size:      " + fileSize + " bytes");
+            System.out.println("Footer Offset:  " + footerOffset + " bytes");
+            System.out.println("Footer Length:  " + footerLength + " bytes");
             System.out.println("Leading Magic:  " + new String(leadingMagic, StandardCharsets.US_ASCII));
             System.out.println("Trailing Magic: " + new String(trailingMagic, StandardCharsets.US_ASCII));
         }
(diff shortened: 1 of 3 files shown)


gunnarmorling:
Going to close this one.

This account opened dozens of pull requests across unrelated projects in a short period of time. Hardwood doesn't take unsolicited contributions from accounts mass-submitting PRs at that scale — reviewing them costs maintainer time the project would rather spend elsewhere, and that holds whether or not an individual change is good.

I've since clarified this in the Hardwood [contribution guidelines](https://github.com/hardwood-hq/hardwood/blob/main/CONTRIBUTING.md#a-note-on-ai-assisted-contributions) for future reference.

If I've misread this, say so and I'll take another look.

PR #888, opened 2026-08-10
#887 Clarifying rules for LLM-based contributions

gunnarmorling:
Hey @iifawzi and @rionmonster, this is to deal with the recent influx of apparently bot-generated PRs and to generally disambiguate the rules around LLM usage. Any feedback welcome!

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,3 +1,12 @@
+## Authorship
+
+Hardwood is built with AI, not by AI. LLM assistance is welcome; submitting raw LLM output without understanding is not.
+Neither are unattended agents opening pull requests.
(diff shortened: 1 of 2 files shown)


iifawzi (approved):
Looks good to me, It started to get crazy recently, and I can't really understand why people do so, it has zero benefits on them as well, so strange, I hope with that things are clearer

gunnarmorling:
Yeah, it's becoming a problem fore sure. I feel like the "help wanted" label contributes to it, seems bots are targeting it. Might help to drop that label from all issues carrying it.

iifawzi:
> Yeah, it's becoming a problem fore sure. I feel like the "help wanted" label contributes to it, seems bots are targeting it. Might help to drop that label from all issues carrying it.

It's really tricky, we might lose interested humans in that case. I personally attempt to start with good first contribution, and help wanted whenever I approach a new project, so not sure, but yeah most likely it's automatically targeted, what has been happening last 5 days is crazy.. perhaps we can come with our own variant, that can't be targeted, like first-friendly, or idk something similar

PR #913, opened 2026-08-11
#912 Widening the Thrift type-code table so every module can reach it

### LINKED ISSUE #912: Propagate #668's centralized Thrift type codes across the repo
#668 centralized the Thrift Compact Protocol type codes into `ThriftCompactConstants` and converted the production readers and writers in `dev.hardwood.internal.thrift`. It stopped at the package boundary.

The cause is visibility. `ThriftCompactConstants` is package-private:

```java
// core/src/main/java/dev/hardwood/internal/thrift/ThriftCompactConstants.java:13
final class ThriftCompactConstants {
```

Its nested `FieldType` / `ElementType` enums are `public`, but a public member of a package-private class is unreachable from outside. The single source of truth is single only within one package; everything else must re-declare the literals.

So three copies of the same wire table are still checked in — `ThriftStructBuilder.java:21-33` (`TYPE_*` and `ELEM_*`, the table every other thrift test builds on), `s3/.../TestParquetGenerator.java:222-226` (`T_*`, in `int` rather than `byte`), and `BoundingBoxReaderTest.java:23` (`TYPE_DOUBLE`, the exact duplicate #668 deleted from `BoundingBoxReader`) — alongside raw literals at the call sites in `ThriftCompactReaderTest`, `PageEncodingStatsReaderTest`, `MalformedMetadataValidationTest`, and `ColumnOrderReaderTest`, still carrying the `// TYPE_MAP`-style comments #668 set out to delete.

#668's own criterion "Tests updated to reference the shared definition" is therefore ~13 files short: 2 of 15 files in `core/src/test/java/dev/hardwood/internal/thrift/` reference the shared constants.

iifawzi:
Hello, this PR is a follow up on #668 that I wanted to do since a while, de-hardcoding all the bytes across the repo (I asked Claude to find them and utilize the constants), this makes the code easier to understand and more maintainable 

## Authorship

Check this box to confirm:

## Checklist

### DIFF
--- a/core/src/main/java/dev/hardwood/internal/thrift/ThriftCompactConstants.java
+++ b/core/src/main/java/dev/hardwood/internal/thrift/ThriftCompactConstants.java
@@ -7,10 +7,11 @@
  */
 package dev.hardwood.internal.thrift;
 
-/// Thrift Compact Protocol type codes, defined once and shared by the reader and
-/// writer so the wire constants are not duplicated.
+/// Thrift Compact Protocol type codes, defined once and shared by every reader,
+/// writer and test fixture that encodes or decodes the wire format, so the
+/// constants are not duplicated.
 /// Reference: https://github.com/apache/thrift/blob/master/doc/specs/thrift-compact-protocol.md
-final class ThriftCompactConstants {
+public final class ThriftCompactConstants {
 
     /// Type codes for struct field headers.
     public enum FieldType {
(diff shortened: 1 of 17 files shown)


gunnarmorling:
Hey @iifawzi, thanks a lot. Pushed up one more commit after review from Claude, could you check it out? I'm not 100% sure whether it matches what you had in mind; in particular, it reverted the public visibility change of `Codes` which wasn't needed. WDYT?

iifawzi:
ops ops.. that's indeed better, making codes public wasn't necessary for sure.. feeling bad I've not caught that, and even worse that Claude decided to do that initially... thanks for the double check

iifawzi:
Going to merge it, thank you!

PR #916, opened 2026-08-12
#655 Expose metadata for each input file

### LINKED ISSUE #655: Expose per-file metadata from a multi-file ParquetFileReader
## Summary

`Hardwood.openAll(...)` returns a single `ParquetFileReader` over many files, but `getFileMetaData()` reports only the *first* file's footer (the schema the files share). There is no way to reach the other files' footers through the combined reader, so any per-file or whole-dataset metadata (row counts, on-disk sizes, statistics) has to be recomputed by opening each file separately and summing manually.

This came up in review of hardwood-hq/hardwood-examples#27, where the example opens every month individually just to print a per-file breakdown and derive the true dataset total:

```java
// A multi-file reader reports only the first file's footer, so summing
// per-file footers is the only way to get the dataset's true total.
for (Path month : months) {
    try (ParquetFileReader reader = hardwood.open(InputFile.of(month))) {
        long rows = reader.getFileMetaData().numRows();
        // ...
    }
}
```

## Proposal / discussion points

*(Not decided — opening this to track the idea.)*

Provide an accessor on the multi-file reader that exposes each backing file's metadata, e.g.:

- `getFilesMetaData()` (or similar) returning the list of per-file `FileMetaData`, in file order.
- Open question: how this relates to the existing single-footer `getFileMetaData()` — keep it as "the shared schema" and add a plural accessor, or rethink the contract for multi-file readers.
- Open question: whether to also offer convenience aggregates (total row count, etc.) or leave folding to the caller.

### Risk: lazy footer reads

The multi-file reader is designed to stream — it prefetches across file boundaries rather than opening everything up front, so it does **not** necessarily read every file's footer eagerly. A `getFilesMetaData()` accessor that materializes all footers would force eager I/O across the whole file set, undercutting that laziness (and surfacing failures for files that a partial scan would never have touched). Design points to settle:

- Should the accessor be lazy (e.g. return a lazy/streaming view or read footers on demand) rather than eagerly loading all of them?
- What are the semantics when only part of the dataset has been scanned so far — read-ahead only, or trigger reads for not-yet-touched files?
- Be explicit in the docs/contract about the I/O cost so callers don't unknowingly turn a streaming scan into an all-footers read.

## References

- hardwood-hq/hardwood-examples#27 (review comment: https://github.com/hardwood-hq/hardwood-examples/pull/27/#discussion_r3370266528)

k-barseghyan:
Fixes #655.

## Summary

* Adds indexed metadata access for every input file while preserving the existing first-file accessor.
* Loads later footers lazily and shares completed, in-progress, and failed loads between metadata and data-reader paths.
* Preserves per-reader schema validation and column-ordinal mappings.
* Coordinates cache closure with input ownership and preserves specific checked I/O exceptions.
* Documents the API, lifecycle, compatibility, and validation behavior.

## Testing

Added coverage for metadata ordering and bounds, concurrent footer reuse, cached failures and reopen retries, close coordination, encrypted files, schema compatibility, ordinal remapping, and input ownership.

`./mvnw clean verify` passes across all 13 reactor modules.

## Authorship

Check this box to confirm:

* [x] This change is coming from a human who understands what it does and why, and can discuss it in review

## Checklist

* [x] Build via `./mvnw clean verify` passes
* [x] The commit message is prefixed with the issue key ("#123 Adding support for...")
* [x] Code changes are covered by tests
* [x] If this PR adds or changes user-facing behaviour, `docs/` has been updated

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -231,6 +231,7 @@ For field-level `parquet.thrift` metadata coverage (which spec fields are read/p
 - [x] Data page reading and decoding
 - [x] Page decompression
 - [x] Parallel column batch fetching
+- [x] Lazy per-file metadata access with a parent-owned footer cache (see [`_designs/PER_FILE_METADATA.md`](_designs/PER_FILE_METADATA.md))
 
 ### 7.3 Record Assembly (Inverse Dremel)
 - [x] Column reader synchronization (via RowReader)
--- a/_designs/PER_FILE_METADATA.md
+++ b/_designs/PER_FILE_METADATA.md
@@ -1,0 +1,68 @@
+# Per-File Metadata for Multi-File Readers
+
+**Status:** Completed
+
+## Context
+
+A multi-file `ParquetFileReader` exposed only the first file's `FileMetaData`. Consumers that
+need the total physical row count before allocating storage had to reopen every input and parse
+its footer independently, even though Hardwood would parse those same footers while reading.
+
+## Public contract
+
+- `getFileCount()` returns the number of physical inputs, in supplied order, without I/O.
+- The existing `getFileMetaData()` continues to return the eagerly parsed first footer.
+- `getFileMetaData(int)` returns one physical input's footer; index `0` returns that same first
+  `FileMetaData` instance.
+- Later footers remain lazy. Metadata access is synchronous and may join an asynchronous load
+  already started by a data reader.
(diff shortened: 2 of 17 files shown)


gunnarmorling:
This LGTM, merging. Thanks a lot, @k-barseghyan, happy to see this change land and that you're getting better numbers out of Hardwood with it 🥳.

PR #922, opened 2026-08-13
#919 Wide-schema metadata decode: benchmark, comparison and allocation work

### LINKED ISSUE #919: Benchmark footer metadata decode time for wide schemas (10 … 100,000 FLOAT64 columns)
InfluxData's ["How Good is Parquet for Wide Tables (Machine Learning Workloads) Really?"](https://www.influxdata.com/blog/how-good-parquet-wide-tables/) measures Parquet footer decode time as a function of column count: files of 10 … 100,000 `FLOAT64` columns, ten row groups each, so the footer carries `10 × columns` `ColumnChunk` structures. For parquet-rs they report roughly 5 µs of decode time and ~700 bytes of footer per column.

We have no equivalent number for Hardwood. Metadata decode is pure Thrift work on the read path and is paid in full before a single value is read, so on wide schemas it can dominate a projected scan of a handful of columns.

Add a JMH micro-benchmark sweeping the same shape, splitting the cost into:

- Thrift decode of the footer bytes (no I/O),
- `FileSchema` construction from the decoded schema elements,
- the full `ParquetFileReader.open()` a user actually pays (mmap + footer read + decode + schema build).

Fixtures are generated on demand with Hardwood's own writer and cached in the benchmark data directory.

### LINKED ISSUE #920: Cut allocation in footer metadata decode
The wide-schema benchmark added in #919 shows footer decode allocating ~7.5 bytes of garbage per footer byte: 750 MB per decode of a 69 MB footer (100,000 columns × 10 row groups), which retains 456 MB of live metadata. Allocation on that scale makes the decode a GC measurement as much as a parsing one.

An allocation profile attributes it to, in order:

- `ThriftCompactReader.FieldHeader` — one small object per field read, tens of millions per wide footer, escaping into `acceptField` rather than being scalarized.
- The `byte[]`/`String` pair per string read, and the `ArrayList` + backing array + unmodifiable wrapper per `path_in_schema` and per encodings list — two such lists per column chunk.
- `FieldPath` per column chunk, when a column's path is identical in every row group.

Addressed by packing the field header into an `int`, decoding strings in place from a heap-backed buffer, building the small lists with `List.of`, and caching each column's path positionally (`RowGroup.columns` is required to follow schema order, checked against the encoded bytes so a file that breaks the ordering still decodes correctly).

Content-keyed interning of strings and paths was measured as an alternative and rejected: it cut allocation further but cost 40-60% more wall-clock, since hashing every path and probing a table of them is a cache miss per lookup where the decode itself is a sequential scan.

gunnarmorling:
Closes #919, closes #920.

Prompted by InfluxData's [How Good is Parquet for Wide Tables Really?](https://www.influxdata.com/blog/how-good-parquet-wide-tables/), which measures footer decode against column count and reports ~5 µs and ~700 bytes of footer per column for parquet-rs 51. We had no comparable number.

## What is here

**`#919` — the benchmark.** `WideSchemaMetadataBenchmark` sweeps 10 … 100,000 `FLOAT64` columns over ten row groups (so `columns × 10` column chunks), splitting the cost into Thrift decode, `FileSchema` construction, and the `open()` a caller actually pays. `WideSchemaMetadataParquetJavaBenchmark` runs the same three steps through parquet-java over the identical fixtures. Fixtures are written by our own writer on demand rather than checked in; the widest is 150 MB with a 69 MB footer.

**`#920` — the decode.** The benchmark showed decode allocating ~7.5 bytes per footer byte — 750 MB of garbage for a 69 MB footer, retaining 456 MB. Addressed by packing the Thrift field header into an `int` (it was one object per field read, tens of millions per wide footer, escaping into `acceptField` rather than being scalarized), decoding strings in place from heap-backed buffers, building one- and two-element lists with `List.of`, and caching each column's `path_in_schema` positionally.

The path cache deserves a note for review: `RowGroup.columns` is *required* to follow schema order ("This list must have the same order as the SchemaElement list in FileMetaData"), but the cache treats that as a **prediction** and verifies the cached entry against the encoded bytes before reusing it. A file that breaks the ordering decodes correctly and merely loses the shortcut. `RepeatedPathCacheTest` covers shuffled row groups, row groups of differing column counts, and multi-component paths.

Content-keyed interning of strings and paths was implemented and measured as the alternative, and rejected: it cut allocation further but cost 40-60% more wall-clock, because hashing and probing a table is a cache miss per string where the decode is otherwise a sequential scan. The positional cache is the same trick `Dictionary.internedString` already uses on the value path — an index, not a hash. parquet-java reaches for interning too, through `String.intern()` in its `InterningProtocol`, and pays for it (see the `stock Thrift` column below).

**`#920` — skipping.** Skipping a field allocated as much as reading one: the binary case read the value into a fresh array only to drop it, and every list header — several per column chunk — escaped as a record. Both are fixed, taking a full footer walk from 131 MB of allocation to 453 B. Nobody pays this on a full decode; it is the floor for anything that walks past metadata rather than through it, an unrecognised field today and selective decoding later.

**`#919` — the skip floor.** `WideSchemaSkipFloorBenchmark` measures what that floor is: a structural walk that materializes nothing still costs about two thirds of a full decode, because Thrift carries no index and the byte range of each chunk is only found by parsing forward. Worth having recorded — it bounds what any selective or deferred decoding scheme can win, and says such a scheme is a memory change first and a latency change second.

## Numbers

n300, 8 Gracemont cores pinned at 1.50 GHz (calibrated, no thermal throttling), Temurin 25. Same fixtures for every contender, and the same warmup for every JMH benchmark (2 iterations), matching the warmup phase the parquet-rs harness uses. `decode` is Thrift decode of an in-memory footer; note parquet-rs builds its `SchemaDescriptor` inside decode, which we split out as `buildSchema`.

Decode, ms:

| columns | Hardwood before | Hardwood after | Hardwood after, mapped | parquet-rs 59.2 | parquet-java 1.17.1 | stock Thrift |
|---:|---:|---:|---:|---:|---:|---:|
| 10 | 0.062 | 0.050 | 0.049 | 0.039 | 0.218 | 0.184 |
| 100 | 0.607 | 0.490 | 0.465 | 0.392 | 2.182 | 1.914 |
| 1,000 | 6.04 | 5.07 | 4.73 | 4.70 | 21.8 | 21.1 |
| 10,000 | 61.5 | 46.5 | 48.4 | 47.3 | 228.3 | 204.5 |
| 100,000 | 709.3 | 471.2 | 497.5 | 709.4 | 2360.5 | 2233.6 |

`decode` reads an in-memory footer off the heap. `decode, mapped` reads the same bytes off a
memory-mapped, and so direct, buffer, which is what every input this reader has — a mapped local
file, the range cache, S3 — actually hands it. The two differ by at most a few percent and change
sign across the sweep, so the heap column is representative rather than favourable; see below.

`buildSchema` adds 0.24 / 2.54 / 25.2 ms at 1k / 10k / 100k; parquet-rs's separate Arrow-schema step is 0.32 / 5.54 / 33.5 ms.

Full open, ms: 0.079 / 0.565 / 5.13 / 51.7 / **555** (Hardwood) against 0.055 / 0.418 / 5.02 / 48.5 / **754** (parquet-rs) and 1.8 / 14.6 / 159 / 1466 / **15870** (parquet-java).

So `#920` buys 16-34% on decode, most of it at the wide end. Against parquet-rs the honest reading is still a **crossover**: comparing like with like (our decode + schema build against its decode), parquet-rs is ahead up to 10,000 columns — though only by 3.6% there now, 49.0 ms against 47.3 — and we are ahead at 100,000, 497 ms against 709, where our full open wins too. Allocation per decode fell 752 -> 305 MB and the retained metadata 456 -> 273 MB, which is why the wide end moves most.

The `stock Thrift` column reads the same generated structures off a bare `TCompactProtocol`, without the `InterningProtocol` that parquet-java's `Util` wraps around it: interning every repeated string through the JVM string pool costs them 6-19% here.

`#920`'s skip fixes do not move these numbers — they pay only where a walk passes metadata by rather than materializing it, where they take a full footer walk from 131 MB of allocation to 453 B.

Two caveats on the comparison. It is eager against eager: parquet-rs has since shipped `ParquetMetaDataOptions` with statistics policies (`skip_except` for named columns), so a narrow-projection query there can skip decoding statistics it will never read, and would land below these numbers — we decode everything, as does the parquet-rs configuration measured here. And the widths below 100,000 columns favour parquet-rs on this hardware; the crossover is real but it is a crossover, not a lead.

## Review follow-ups

`#920` grew three changes out of @iifawzi's review.

**A varint's single-byte case returns before entering the loop.** A footer's varints are
overwhelmingly one byte — every field id, every small size, every enum value — and a non-negative
first byte is exactly that case. Worth 10% of the heap decode and 8.7% of the mapped one at
100,000 columns, at no cost in allocation, and it lands on the page-header path too since that
shares `readVarint`. The decode column above includes it.

**Two reads answered a malformed footer with a wrong number.** `readVarint` let its shift grow
unbounded, and Java applies a `long` shift modulo 64, so an eleventh byte folded its payload back
over the low bits of the result. `readFieldHeader` narrowed a long-form field id to `short`, which
maps id 65537 onto id 1 and has the struct reader decode that field's bytes as whatever it holds
field 1 to be. Both are now rejected, each covered by a test that fails without the fix, and
neither check is measurable in the decode.

**The buffer the decoder is handed does not matter.** Every input here produces a direct buffer —
a mapped local file, the range cache over a sparse temp file, and S3's `allocateDirect` fetches —
so the `hasArray()` paths that read a string in place and compare a cached column path through
`Arrays.equals` never fire in production. Buying them with one copy of the footer onto the heap
was implemented and measured, and it loses at every width: 19% / 4.8% / 0.6% slower on decode and
16% / 7% / 1.6% on open, plus the footer's own size in allocation. A direct buffer's `get` is a
bare memory load where a heap buffer's is an array access the JIT bounds-checks, and that offsets
what the array-only paths win. `decodeFooterMapped` is the arm that pins this down; the decoder
goes on reading whatever buffer the file gives it.

## Measurement provenance

The Hardwood columns are a fresh run of the branch head; the `before` column and the three
contenders are from the original run, on the same box at the same pinned clock. `buildSchema`,
which nothing in this branch touches, reproduces across the two sessions to within 1% — 25.4 ms
against 25.2 at 100,000 columns — which is the check that the two are comparable. `openFile` at
100,000 columns is the least stable arm at ±42 ms; every other figure quoted here is within 1.5%.
The parquet-java version is stamped by the benchmark from the artifact on its classpath, so the
table names the version it measured rather than the one it was assumed to be.

Footer size is ~690 bytes per column, matching the post's ~700.

### DIFF
--- a/core/src/main/java/dev/hardwood/internal/thrift/ThriftCompactReader.java
+++ b/core/src/main/java/dev/hardwood/internal/thrift/ThriftCompactReader.java
@@ -13,6 +13,7 @@ import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
@@ -45,9 +46,26 @@ public class ThriftCompactReader {
     private static final byte TYPE_MAP = Codes.MAP;
     private static final byte TYPE_STRUCT = Codes.STRUCT;
 
+    /// Returned by [#readFieldHeader] for the STOP field that ends a struct. No real header packs
+    /// to this value: it would take a wire type of `0xFF`, and the type comes from a nibble.
+    public static final int STOP_FIELD = -1;
+
+    private static final int FIELD_ID_SHIFT = 8;
+    private static final int TYPE_MASK = 0xFF;
+
+    /// Returned by [#acceptListHeader] for a list field that is absent or carries the wrong
+    /// element type. No real header packs to this value: it would need a negative element count.
+    static final long ABSENT_LIST = -1;
+
+    /// Element type occupies the low byte of a packed list header, the count the rest.
+    private static final int ELEMENT_TYPE_BITS = 8;
+
     private final ByteBuffer buffer;
     private final int startPosition;
     private short lastFieldId = 0;
+    /// Per-decode cache of repeated column paths, created on first use so the page-header path —
+    /// which has none — never allocates one.
+    private RepeatedPathCache pathCache;
 
     /// Creates a reader that reads directly from a ByteBuffer.
     ///
@@ -74,6 +92,52 @@ public class ThriftCompactReader {
     /// Returns the number of bytes still available to read in the buffer.
     public int remaining() {
         return buffer.remaining();
+    }
+
+    /// This decode's cache of repeated column paths.
+    RepeatedPathCache pathCache() {
+        if (pathCache == null) {
+            pathCache = new RepeatedPathCache();
+        }
+        return pathCache;
+    }
+
+    /// The reader's current offset into its buffer, for capturing a byte range that was read.
+    int position() {
+        return buffer.position();
+    }
+
+    /// Whether the bytes at the current position are `expected`, without moving the position.
+    boolean matchesAt(byte[] expected) {
+        if (buffer.remaining() < expected.length) {
+            return false;
+        }
+        int from = buffer.position();
+        if (buffer.hasArray()) {
+            int offset = buffer.arrayOffset() + from;
+            return Arrays.equals(expected, 0, expected.length, buffer.array(), offset, offset + expected.length);
+        }
+        for (int i = 0; i < expected.length; i++) {
+            if (expected[i] != buffer.get(from + i)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    /// Advances past `length` bytes already known to be present.
+    void skipBytes(int length) throws EOFException {
+        if (buffer.remaining() < length) {
+            throw new EOFException("Unexpected EOF while skipping " + length + " bytes");
+        }
+        buffer.position(buffer.position() + length);
+    }
+
+    /// A copy of `length` bytes from `from`, independent of this reader's buffer.
+    byte[] copyRange(int from, int length) {
+        byte[] bytes = new byte[length];
+        buffer.get(from, bytes, 0, length);
+        return bytes;
     }
 
     /// Returns a zero-copy, read-only, little-endian view of the next `length` bytes and advances
@@ -197,18 +261,39 @@ public class ThriftCompactReader {
     }
 
     /// Read a string value.
+    ///
+    /// A heap-backed buffer — which is what the footer and every page header are read from — is
+    /// decoded in place, so the string costs one object rather than a `byte[]` copy and then the
+    /// string built from it.
     public String readString() throws IOException {
-        return new String(readBinary(), StandardCharsets.UTF_8);
+        int length = checkedBinaryLength(readVarint());
+        if (buffer.hasArray()) {
+            if (buffer.remaining() < length) {
+                throw new EOFException("Unexpected EOF while reading bytes");
+            }
+            String value = new String(buffer.array(), buffer.arrayOffset() + buffer.position(), length,
+                    StandardCharsets.UTF_8);
+            buffer.position(buffer.position() + length);
+            return value;
+        }
+        byte[] data = new byte[length];
+        readBytes(data);
+        return new String(data, StandardCharsets.UTF_8);
     }
 
-    /// Read a field header and return field info.
-    /// Returns null when STOP field is encountered.
-    public FieldHeader readFieldHeader() throws IOException {
+    /// Read a field header and return it packed into an `int`: the field id in the upper bits,
+    /// the wire type in the low byte. Returns [#STOP_FIELD] when the STOP field is encountered.
+    ///
+    /// The header is packed rather than returned as an object because a struct-heavy footer
+    /// reads one per field — a wide file's footer runs to tens of millions — and the object
+    /// escapes into [#acceptField], so it is allocated for real rather than scalarized.
+    /// Unpack it with [#fieldId] and [#fieldType].
+    public int readFieldHeader() throws IOException {
         byte b = readByte();
 
         if (b == ThriftCompactConstants.STOP) {
             lastFieldId = 0;
-            return null;
+            return STOP_FIELD;
         }
 
         byte type = (byte) (b & 0x0F);
@@ -225,7 +310,18 @@ public class ThriftCompactReader {
         }
 
         lastFieldId = fieldId;
-        return new FieldHeader(fieldId, type);
+        return (fieldId << FIELD_ID_SHIFT) | type;
+    }
+
+    /// The field id of a header returned by [#readFieldHeader].
+    public static short fieldId(int fieldHeader) {
+        return (short) (fieldHeader >> FIELD_ID_SHIFT);
+    }
+
+    /// The wire type of a header returned by [#readFieldHeader], one of
+    /// [ThriftCompactConstants.FieldType.Codes].
+    public static byte fieldType(int fieldHeader) {
+        return (byte) (fieldHeader & TYPE_MASK);
     }
 
     /// Gate a struct field on its declared wire type: returns `true` with the reader positioned
@@ -239,11 +335,12 @@ public class ThriftCompactReader {
     ///
     /// @param header the field header just read
     /// @param expectedType wire code the field must declare, from [ThriftCompactConstants.FieldType.Codes]
-    boolean acceptField(FieldHeader header, byte expectedType) throws IOException {
-        if (header.type() == expectedType) {
+    boolean acceptField(int header, byte expectedType) throws IOException {
+        byte type = fieldType(header);
+        if (type == expectedType) {
             return true;
         }
-        skipField(header.type());
+        skipField(type);
         return false;
     }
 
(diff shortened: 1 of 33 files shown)


gunnarmorling:
Hey @iifawzi, @rionmonster, spent some cycles for improving the metadata read path, in particular notable for wide schemas with tens of thousands of columns. The Thrift footer doesn't allow for random access, unfortunately, so you can't just address just those parts you actually need (e.g. we'd need statistics only for columns we filter on). In theory, we could skip over stats during `open()`, keep the raw bytes around, and then read only the required stats later on during predicate push-down. I don't think it's worth the effort at this point though. Any feedback welcome.

core/src/main/java/dev/hardwood/internal/thrift/ThriftCompactReader.java
```diff
@@ -331,33 +441,36 @@ CollectionHeader acceptListHeader(byte expectedElementType, String fieldName) th
...
     List<byte[]> readBinaryList(String fieldName) throws IOException {
-        CollectionHeader header = requireListHeader(Codes.BINARY, fieldName);
-        List<byte[]> values = new ArrayList<>(header.size());
-        for (int i = 0; i < header.size(); i++) {
+        long header = requireListHeader(Codes.BINARY, fieldName);
+        List<byte[]> values = new ArrayList<>(listSize(header));
+        for (int i = 0, n = listSize(header); i < n; i++) {
             values.add(readBinary());
         }
         return Collections.unmodifiableList(values);
```
iifawzi, line 503: Not to be part of this PR,  would it be worth applying the same trick here and for the struct method? dropping ArrayList + Collections.unmodifiableList. These two are being called for column indexes per chunks

core/src/main/java/dev/hardwood/internal/thrift/ThriftCompactReader.java
```diff
@@ -331,33 +441,36 @@ CollectionHeader acceptListHeader(byte expectedElementType, String fieldName) th
...
             values.add(elementReader.read(this));
         }
         return Collections.unmodifiableList(values);
     }
 
     /// Read a required `list<string>` in full.
     ///
+    /// The result is an immutable [List#of] list rather than a wrapped [ArrayList], saving two
+    /// objects per list — worth having where a footer holds one such list per column chunk.
+    ///
```
iifawzi: nit: it loses context after the PR is merged (Rather than a wrapper [ArrayList]. Perhaps we can just comment the what?

no strong opinion here, sometimes it's actually useful to always know why a specific decision it has been taken, and it's helpful

core/src/main/java/dev/hardwood/internal/thrift/ThriftCompactReader.java
```diff
@@ -314,13 +424,13 @@ CollectionHeader requireListHeader(byte expectedElementType, String fieldName) t
     /// @param fieldName fully-qualified field name for the log message
     /// @return the header, or `null` if the list declares a different element type and has been
     ///     skipped
```
iifawzi: nit: stale after return values changes

core/src/main/java/dev/hardwood/internal/thrift/ThriftCompactReader.java
```diff
@@ -76,6 +94,52 @@ public int remaining() {
...
+        int from = buffer.position();
+        if (buffer.hasArray()) {
+            int offset = buffer.arrayOffset() + from;
+            return Arrays.equals(expected, 0, expected.length, buffer.array(), offset, offset + expected.length);
+        }
+        for (int i = 0; i < expected.length; i++) {
+            if (expected[i] != buffer.get(from + i)) {
+                return false;
+            }
+        }
```
iifawzi, line 129: This loop isn't vertorizable, maybe it's not a big deal, discussing it with claude it suspects it might be good to save especially for nested columns on a wide schema.  
> ByteBuffer.mismatch / slice(...).equals(...) would be vectorised, but measure on the mapped path before changing — the current loop is correct.

iifawzi (approved):
Hey, thanks for working on that, it looks so good to me, it's a type of performance imprvoements that reminds me why I like being part of Hardwood.. that looks so neat, little details, massive improvement. (yet, looking forward to beat rs numbers :p) 

I left some little comments, and through my discussions with claude about the changes it mentioned these couple of points, in case you can check them: 

---
1. readVarint doesn't bound the byte count (line 158). The loop keeps going while the continuation bit is set, and shift grows past 63. Java masks long shifts to 6 bits, so at shift = 70 the payload lands at bit 6 and silently corrupts the value instead of failing. A 10-byte cap (max for a 64-bit varint) with an IOException would make a malformed/adversarial varint fail early rather than produce a wrong number.

2. readFieldHeader narrows the long-form field id with (short) readZigzag() (line 306). Field ids are i16 by spec, so anything outside is malformed — but the cast makes id 65537 come out as 1, and the reader then happily decodes some other field's bytes as field 1. That's the "unsafe downcast with value loss" CLAUDE.md forbids; a range check → IOException is what it wants. (lastFieldId + delta at line 310 has the same theoretical wrap at 32767, less likely.)

3. Micro-perf, only with a benchmark: a single-byte fast path at the top of readVarint (b >= 0 → return b) — most footer varints (field ids, small sizes, enum values) are one byte, and it skips the loop setup. Cheap to try on this branch's benchmark; don't add without numbers. 
---

On the other hand, haven't thought about it so much, but not sure if a potential improvement here can span multi files for the path cache

iifawzi:
> n theory, we could skip over stats during open(), keep the raw bytes around, and then read only the required stats later on during predicate push-down.

This looks like a nice idea on paper too, but I kinda agree it not worthing the effort unless measured

gunnarmorling:
Thanks a lot for reviewing, @iifawzi. I've addressed the key remarks, that micro-optimization you suggested made for a nice additional improvement 🤗. Merging.

PR #924, opened 2026-08-13
#914 Reject non-string keys in Avro maps

### LINKED ISSUE #914: Avro reader fails on maps with non-string keys
## Symptom

Reading a Parquet `map` whose key is not a UTF8 binary through `AvroRowReader` throws a `ClassCastException` from deep inside the batch layer:

```
java.lang.ClassCastException: class [I cannot be cast to class
    dev.hardwood.internal.reader.BinaryBatchValues
```

Reproduced against `core/src/test/resources/map_types_test.parquet`, whose `int_map` column is `map<int32, int64>`:

```java
try (ParquetFileReader fileReader = ParquetFileReader.open(InputFile.of(fixture));
     AvroRowReader reader = AvroReaders.rowReader(fileReader)) {
    reader.getSchema();     // ["null",{"type":"map","values":["null","long"]}]
    reader.next();          // ClassCastException
}
```

Other fixtures with the same shape: `time_keyed` (`map<int32, int32>`) and `long_keyed` (`map<int64, string>`).

## Cause

Two halves of one assumption that keys are always strings.

`AvroSchemaConverter.convertMap` emits `Schema.createMap(valueSchema)` for every Parquet map. Avro's `MAP` has no key schema — the spec fixes keys as strings — so the converted schema silently claims string keys for a map whose Parquet key is an `INT32`.

`AvroRowReader.materializeMap` then calls `entry.getStringKey()` unconditionally, which reaches `BinaryBatchValues` against an `int[]`-backed key column.

The schema is wrong before the read even starts, so the exception is the second symptom rather than the defect.

## Fix direction

Three options, in increasing order of fidelity:

1. **Reject at conversion.** The converter knows the key's physical type, so it can throw naming the column and its key type instead of leaving a `ClassCastException` to surface per row. Smallest change, and it makes the limitation legible.
2. **Stringify the key.** Keeps the Avro `map` shape for every Parquet map, at the cost of round-trip fidelity — an `int` key comes back as `"42"` and the original type is not recoverable from the record.
3. **Emit an array of key/value records** for non-string-keyed maps, which is the shape Avro leaves for maps it cannot express. Highest fidelity and the largest change: the decode plan needs a key node, which it does not have today precisely because an Avro `MAP` gives a key nothing to pair with.

Worth checking what parquet-java's `AvroSchemaConverter` does for this case before choosing, so a user moving between the two readers is not surprised. I have not verified it.

## Acceptance criteria

- A `map<int32, V>` column read through `AvroRowReader` either produces well-defined values or fails with an error naming the column and its key type — not a `ClassCastException` from the batch layer.
- Whichever option is taken applies at every map position: top level, nested in a struct, as a list element, and as a map value.
- The behaviour is covered against the existing `int_map`, `time_keyed` and `long_keyed` fixtures.

arnabnandy7:
## Summary

Fixes #914.

Rejects Parquet maps with non-string keys during Avro schema conversion. The resulting error identifies the map and its key type instead of allowing an incorrect Avro schema to cause a later `ClassCastException`.

This matches parquet-java’s behavior because Avro maps only support string keys.

## Changes

- Validate that Parquet map keys are `BYTE_ARRAY` annotated as `STRING`
- Cover the existing `int_map`, `time_keyed`, and `long_keyed` fixtures
- Document the Avro map-key limitation and recommend `RowReader` when typed keys must be preserved

## Authorship

Check this box to confirm:

## Checklist

## Acceptance Criteria

- [x] A map<int32, V> column read through AvroRowReader either produces well-defined values or fails with an error naming the column and its key type — not a ClassCastException from the batch layer.
- [x] Whichever option is taken applies at every map position: top level, nested in a struct, as a list element, and as a map value.
- [x] The behaviour is covered against the existing int_map, time_keyed and long_keyed fixtures.

### DIFF
--- a/avro/src/main/java/dev/hardwood/avro/AvroReaders.java
+++ b/avro/src/main/java/dev/hardwood/avro/AvroReaders.java
@@ -7,10 +7,10 @@
  */
 package dev.hardwood.avro;
 
+import dev.hardwood.avro.internal.AvroPlanNode;
 import dev.hardwood.avro.internal.AvroSchemaConverter;
 import dev.hardwood.reader.FilterPredicate;
 import dev.hardwood.reader.ParquetFileReader;
-import dev.hardwood.reader.RowReader;
 import dev.hardwood.schema.ColumnProjection;
 
 /// Factory for creating [AvroRowReader] instances from a
@@ -97,6 +97,11 @@ public final class AvroReaders {
         }
 
         public AvroRowReader build() {
+            // Convert before building the underlying reader: conversion rejects
+            // schemas Avro cannot represent, and building the underlying reader
+            // already starts a column worker per projected column — workers that
+            // a rejection here would orphan, with no AvroRowReader to close them.
+            AvroPlanNode plan = AvroSchemaConverter.plan(fileReader.getFileSchema(), projection);
             ParquetFileReader.RowReaderBuilder underlying = fileReader.buildRowReader()
                     .projection(projection);
             if (filter != null) {
@@ -108,9 +113,7 @@ public final class AvroReaders {
             if (tailRows > 0) {
                 underlying.tail(tailRows);
             }
-            RowReader rowReader = underlying.build();
-            return new AvroRowReader(rowReader,
-                    AvroSchemaConverter.plan(fileReader.getFileSchema(), projection));
+            return new AvroRowReader(underlying.build(), plan);
         }
     }
 }
--- a/avro/src/main/java/dev/hardwood/avro/internal/AvroSchemaConverter.java
+++ b/avro/src/main/java/dev/hardwood/avro/internal/AvroSchemaConverter.java
@@ -195,6 +195,13 @@ public final class AvroSchemaConverter {
         }
         SchemaNode inner = mapGroup.children().getFirst();
         if (inner instanceof SchemaNode.GroupNode kvGroup && kvGroup.children().size() >= 2) {
+            SchemaNode keyNode = kvGroup.children().getFirst();
+            if (!(keyNode instanceof SchemaNode.PrimitiveNode key)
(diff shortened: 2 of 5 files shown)


gunnarmorling:
So what are other readers serializing into Avro do in this case, parquet-java in particular?

arnabnandy7:
> So what are other readers serializing into Avro do in this case, parquet-java in particular?

Sorry got a bit late was in another call. @gunnarmorling I checked parquet-java, and it rejects these maps during Parquet-to-Avro schema conversion. Its `AvroSchemaConverter` requires map keys to be binary with the UTF-8/string annotation otherwise, it throws an `IllegalArgumentException`.

I followed the same approach here so we fail early with a useful error instead of producing an incorrect Avro schema and failing later during row materialization. Stringifying the keys would lose their original type, while representing the map as an array of key/value records would behave differently from parquet-java.

Reference: https://github.com/apache/parquet-java/blob/master/parquet-avro/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java#L464-L487

gunnarmorling:
Hey @arnabnandy7, thanks a lot! LGTM overall. Pushed one more commit to refine the behavior a bit: enum/json keys are now allowed too (they're considered strings generally, so nothing speaks against treating them as such here too), whereas un-annotated binary is refused as it may yield collisions when converting to UTF-8.

arnabnandy7:
> Hey @arnabnandy7, thanks a lot! LGTM overall. Pushed one more commit to refine the behavior a bit: enum/json keys are now allowed too (they're considered strings generally, so nothing speaks against treating them as such here too), whereas un-annotated binary is refused as it may yield collisions when converting to UTF-8.

Thanks for the review and improvement.

Issue #510, opened 2026-05-29
dive: exact nested shape from decoded levels (list-length distribution, null clustering)


gunnarmorling:
## Problem

`SizeStatistics` gives only *aggregates*: `histogram[i]` counts level slots equal to `i`, so the footer yields totals and therefore a **mean**, and nothing else. A mean fan-out of 6.19 values per record is equally consistent with "every list holds about 6 elements" and "97% hold one, a handful hold 200k". Those two files behave completely differently — the second is what makes batch sizing allocate a cache-hostile value array — and the metadata-derived rows render them identically.

Three more things the footer structurally cannot answer:

- **Null clustering.** Whether nulls arrive in runs or spread evenly drives RLE efficiency and page-skip potential. Aggregates cannot see it.
- **Files without size statistics.** Anything written before parquet-format 2.10, or by a writer that does not emit the struct. There the metadata pane shows nothing at all.
- **Whether the declared statistics are true.** Once the levels are decoded, comparing against the footer's histograms validates the writer.

## Relationship to #870 and #607

#607 parses `SizeStatistics`; #870 surfaces it passively on `ColumnChunkDetailScreen` —
free, always visible, approximate. This issue is the opposite trade: an explicit, on-demand computation that decodes the level streams to produce the exact shape. The two share the schema-derived level labels described in #870.

Everything #870 covers is **out of scope here**: null-vs-empty-list counts, present-value counts, mean fan-out. Show them if they fall out of the decode anyway (they feed the declared-vs-actual check), but they are not the reason for this feature.

## Proposed screen

Reached explicitly from the `ColumnChunkDetailScreen` drill menu — this decodes a whole column chunk, so it must be an action with progress and cancellation, not a passive pane. Same example column as #870 (`order.tags.list.element`, `LIST<optional STRING>`):

```
╭ order.tags.list.element — shape (RG #0) ─────────────────╮
│ Computed from 6,488,062 decoded levels · 1.4 s           │
│                                                          │
│ Lists                 996,148  (52,428 records: no list) │
│ List length           min 0 · p50 3 · p95 14 · max 4,096 │
│ Mean                  6.36 per list · 7.10 non-empty     │
│                                                          │
│ Length distribution                                      │
│         0   104,857  10.5% ████▏                         │
│         1   249,037  25.0% ██████████                    │
│       2–3   298,844  30.0% ████████████                  │
│       4–7   199,230  20.0% ████████                      │
[...]

PR #934, opened 2026-08-13
870 dive size statistics

### LINKED ISSUE #870: dive: surface SizeStatistics and rep/def level histograms on the column chunk detail screen
## Context

Once #607 lands the `SizeStatistics` parse (chunk-level `ColumnMetaData.size_statistics`, per-page `ColumnIndex` fields 6/7, `OffsetIndex.unencoded_byte_array_data_bytes`), the data is available but nothing displays it. This issue covers surfacing it on `ColumnChunkDetailScreen` — the natural home, since it is a detail pane with vertical room for the full histograms.

Depends on #607. Related to #510, which proposes the same kind of insight for the same screen but from a different source: #510 decodes the level streams to get a true array-length *distribution*, while this issue reads level *aggregates* out of the footer with no decode. The level histograms cannot produce a length distribution — `histogram[i]` counts level slots equal to `i`, so it yields totals and averages only. The two should share the label column and derived rows described below.

## Proposed screen

Example column is `order.tags.list.element`, a `LIST<optional STRING>` inside an optional group (`max_def = 3`, `max_rep = 1`) — the shape where the histograms say something the existing rows cannot. Everything above the `Size statistics` separator is unchanged.

```
 hardwood dive │ orders.parquet │ 1.4 GiB │ 12 RGs │ 12.6M rows
 Overview › Row groups › RG #0 › Column chunks › order.tags.list.element
╭ order.tags.list.element (RG #0) ─────────────────────────╮╭ Drill into ──────────────────────────╮
│ Path                  order.tags.list.element            ││▶ Pages           96 pages            │
│ Column idx            7                                  ││  Column index    present · levels    │
│ Physical              BYTE_ARRAY                         ││  Offset index    present · unencoded │
│ Logical               String                             ││  Dictionary      present             │
│ Codec                 ZSTD                               ││                                      │
│ Encodings             RLE_DICTIONARY, PLAIN, RLE         ││                                      │
│                                                          ││                                      │
│ Data offset           4,194,304                          ││                                      │
│ Dict offset           4,096                              ││                                      │
│ Column index offset   68,157,440                         ││                                      │
│ Offset index offset   68,159,020                         ││                                      │
│                                                          ││                                      │
│ Values                6,488,062                          ││                                      │
│ Nulls                 196,606                            ││                                      │
│ Uncompressed          31.8 MiB                           ││                                      │
│ Compressed            12.4 MiB                           ││                                      │
│ Min                   "airport"                          ││                                      │
│ Max                   "wheelchair"                       ││                                      │
│                                                          ││                                      │
│ Size statistics       chunk + 96 pages                   ││                                      │
│ Unencoded             66.0 MiB  (+24.0 MiB lengths)      ││                                      │
│ Records               1,048,576                          ││                                      │
│ Present values        6,291,456                          ││                                      │
│ Avg fan-out           6.19 slots/record                  ││                                      │
│ Avg list length       7.10  (non-empty)                  ││                                      │
│                                                          ││                                      │
│ Def levels            max 3                              ││                                      │
│   0  tags null            52,428   0.8% ▏                ││                                      │
│   1  tags empty          104,857   1.6% ▏                ││                                      │
│   2  element null         39,321   0.6% ▏                ││                                      │
│   3  element present   6,291,456  97.0% ███████████▋     ││                                      │
│                                                          ││                                      │
│ Rep levels            max 1                              ││                                      │
│   0  new record        1,048,576  16.2% █▉               ││                                      │
│   1  tags.list         5,439,486  83.8% ██████████▏      ││                                      │
╰──────────────────────────────────────────────────────────╯╰──────────────────────────────────────╯
 [Tab] pane  [↑↓] move  [Enter] open  [t] logical types  [Esc] back  │  [?] help  [q] quit
```

## Where each row comes from

| Row | Source |
|---|---|
| `Size statistics` | presence: `ColumnMetaData.size_statistics` (chunk) and `ColumnIndex` fields 6/7 (per page) — `chunk + 96 pages` / `chunk only` / `—` |
| `Unencoded` | `unencoded_byte_array_data_bytes`; the parenthetical is `4 × present values`, the length prefixes the field excludes, so the two together are the real PLAIN size. BYTE_ARRAY columns only |
| `Records` | `rep[0]` |
| `Present values` | `def[maxDef]` |
| `Avg fan-out` | `sum(def) / rep[0]` — slots per record, the quantity `BatchSizing.valuesPerRow` computes today from `num_values / num_rows` |
| `Avg list length` | `(sum(def) − def[0] − def[1]) / (rep[0] − def[0] − def[1])` |
| level rows | count, share of `sum(def)`, bar |

The label column is the part that earns the screen. Raw `def[1] = 104,857` means nothing; `tags empty` does, and it is derivable from the schema path alone — walk the optional/repeated nodes from root to leaf, and level *i* names the node the value bottoms out at. Same for rep: level 0 is always "new record", level *i* is the *i*-th repeated ancestor. Contrast `tags null` (52,428 records have no `tags` field) with `tags empty` (104,857 have an empty list): that distinction exists nowhere else in the metadata, `ColumnIndex.null_counts` lumps them together.

The screen is self-checking: `Values` must equal `sum(def)` and `sum(rep)`, and `Nulls` must equal `Values − def[maxDef]`. A writer that disagrees should be painted as an error rather than silently rendered — that is exactly the class of bug people open `dive` to find. (Overlaps with the broader audit in #506.)

## Degraded forms

Flat BYTE_ARRAY column, both histograms legitimately omitted by the spec (`max_rep == 0`, `max_def <= 1`), unencoded size still the interesting number:

```
╭ order.description (RG #0) ───────────────────────────────╮
│ Values                12,582,912                         │
│ Nulls                 0                                  │
│ Uncompressed          184.2 MiB                          │
│ Compressed            61.7 MiB                           │
│ Size statistics       chunk + 240 pages                  │
│ Unencoded             420.1 MiB  (+48.0 MiB lengths)     │
│ Avg value size        35 B                               │
│ Def levels            — (max 1, see Nulls)               │
│ Rep levels            — (not repeated)                   │
╰──────────────────────────────────────────────────────────╯
```

`Avg value size` is the same quantity that would replace the hardcoded 16-byte BYTE_ARRAY estimate in `BatchSizing.columnByteWidth` — screen and sizing code read one field.

A file written before parquet-format 2.10 collapses to a single dim row, with no empty scaffolding for the rows that cannot be filled:

```
│ Size statistics       — (not written)                    │
```

## Non-interactive CLI

The same facts must be reachable without entering the TUI — `dive` and the `inspect` commands are two entry points to the same metadata, not two products.

`hardwood inspect columns` gains an `Unencoded` column in the ranked table, populated for BYTE_ARRAY columns and `-` elsewhere:

```
Rank  Column                   Type        Compressed  Uncompressed  Unencoded  Ratio  # Pages
   1  order.description        BYTE_ARRAY    61.7 MiB     184.2 MiB  420.1 MiB  33.5%      240
   2  order.tags.list.element  BYTE_ARRAY    12.4 MiB      31.8 MiB   66.0 MiB  39.0%       96
   3  order.total_cents        INT64         38.4 MiB      96.0 MiB          -  40.0%      120
```

A new `--column <path>` option prints the per-chunk detail — the same content as the dive facts pane, one row per row group, followed by the level histograms:

```
$ hardwood inspect columns orders.parquet --column order.tags.list.element

order.tags.list.element  BYTE_ARRAY / String  max def 3  max rep 1

RG  Values     Nulls    Records    Present    Fan-out  Unencoded  Size stats
 0  6,488,062  196,606  1,048,576  6,291,456     6.19   66.0 MiB  chunk + 96 pages
 1  6,502,110  201,004  1,048,576  6,301,106     6.20   66.1 MiB  chunk + 96 pages

Definition levels (RG #0, max 3)
 0  tags null            52,428   0.8%  ▏
 1  tags empty          104,857   1.6%  ▏
 2  element null         39,321   0.6%  ▏
 3  element present   6,291,456  97.0%  ███████████▋

Repetition levels (RG #0, max 1)
 0  new record        1,048,576  16.2%  █▉
 1  tags.list         5,439,486  83.8%  ██████████▏
```

Bars are plain characters, not ANSI, so they render identically in both surfaces. The `inspect` commands emit no colour today and this should not change that.

## Scope

- `ColumnChunkDetailScreen` facts pane: the rows above, with the schema-derived level labels and the degraded forms.
- `ColumnChunkDetailScreen` menu hints: `present · levels` / `present · unencoded` when the page-level fields are there, so the user knows whether drilling into the column index will show per-page histograms before going.
- `hardwood inspect columns`: `Unencoded` table column and the `--column <path>` detail mode above.
- Consistency check between `Values` / `Nulls` and the histograms, rendered as an error in both surfaces.
- Docs update under `docs/content/` for both the screen and the command option.

Out of scope, tracked separately: per-page columns on `ColumnIndexScreen` and unencoded bytes on `PagesScreen` / `inspect pages`.

## Open questions

1. **Pane height.** The facts pane is a plain `Paragraph` with no scroll, and this adds ~14 lines (~35 total). Fine at 40 rows, cut off at 24. Either gate the level blocks behind a toggle key (`[l] levels`, default off) alongside the existing `[t]`, or make the facts pane scrollable. The toggle is cheaper and needs no new scroll state in `ScreenState.ColumnChunkDetail`; the derived rows (`Records`, `Present values`, `Avg fan-out`) stay visible at a glance while the raw histograms become a drill-down.
2. **Narrow terminals.** At 100 cols the bars fit; at 80 the left pane is 48 wide and they clip. Suggest dropping the bar below ~56 inner width and the percentage below ~44, keeping level / label / count. Relates to the general policy in #332.

## Styling

Per the decision tree in `_designs/DIVE_THEME.md`, the bars and level counts are tier 5 — content the user came to read — so `Style.EMPTY`, no `Theme` call. `accent()` is reserved for structural anchors and would make the bars read as headings. The `— (not written)` and `— (max 1, see Nulls)` advisories are tier 4, `Theme.dim()`.

No new colours. Bar length already encodes magnitude, so colouring the bars by magnitude would be a redundant second encoding of the same variable, and colouring them by category (present vs absent) would need palette slots the theme deliberately does not have — dive ships one theme with no startup probe, tuned against Solarized dark/light and macOS Terminal. The semantic split is already carried by the label column; if the absence levels need more separation, `Theme.dim()` on those rows is the in-palette way to get it.

One case may genuinely warrant colour: the `Declared vs actual` mismatch marker, which is an error state. The theme has no error tier, so that needs a `Theme` extension rather than an ad-hoc `Color.*` (a smell per CLAUDE.md) — flagging rather than assuming.

Good-first-issue candidate once #607 has landed; blocked until then.

nitrogen404:
Closes #870

The `SizeStatistics` and rep/def level histograms that #607 parsed are displayed on the two CLI surfaces that address the questions “why is this column so large” and “where did these null values originate”.

The level histograms are the only place in the metadata where “the field was absent” and “the list was empty” can be distinguished — `null_counts` combines them. A raw histogram of `[52428, 104857, 39321, 6291456]` doesn’t provide any information on its own, so each bucket is named by traversing the column’s path through the schema: `websites null`, `websites empty`, `element null`, `element present`.

Design: [`_designs/DIVE_SIZE_STATISTICS.md`](_designs/DIVE_SIZE_STATISTICS.md) 
The first commit is the design doc, so it can be reviewed on its own before the implementation that follows it. 

## What changed

**`dive` column chunk detail** now includes the derived rows — unencoded size with its length-prefix parenthetical, records, present values, average fan-out, average list length, and average value size. The `l` toggle switches on the raw level histograms, following the existing `t`. It defaults off because the facts pane doesn’t scroll and the derived rows are the summary worth seeing first. The drill-into menu displays `present · levels` or `present · unencoded` when the page index contains per-page copies, so the reader knows before typing a keystroke.

**`hardwood inspect columns`** introduces an `Unencoded` column and a `—column PATH` detail mode that prints one row per row group followed by the named level histograms. The `—row-group N` option narrows both the row groups and the histograms. Histograms are summed element-wise across row groups instead of sampling the first, ensuring an exact representation of the file-wide block.

Both surfaces share `cli/internal/LevelSummary`, which computes and renders nothing else. The `dive` surface uses its own `fact()` helper to paint it, so labels retain `Theme.primary()`. The `inspect` surface uses `RowTable`. Bars are plain eighth characters, so the two surfaces agree character-for-character.

## Notable decisions

**A fifth `Theme` tier.** The screen is self-checking — `num_values` must equal both histogram totals, and `null_count` must equal the values below `maxDef`. A writer that disagrees is a defect worth surfacing, and none of the four existing roles mark content as *wrong*. `DIVE_THEME.md` declares its decision tree exhaustive, so `error()` is added there rather than applied ad hoc; it is numbered tier 0 so the existing tiers keep their numbers.

**A malformed histogram is reported, not thrown.** When a histogram’s length doesn’t match the column’s maximum level, it’s routed to the consistency check instead of being indexed. This ensures that the files worth opening `dive` on are precisely the malformed ones, and throwing would cause the screen to crash instead of displaying the defect.

**Rows are dropped, not scaffolded.** For columns that cannot repeat, `Records` and `Avg fan-out` are withheld, as they would simply restate `Values`. However, the present-value count still defaults to `num_values` for a required column, which maintains the availability of `Avg value size` on a flat required `BYTE_ARRAY`.

**No boxing.** Optional quantities are represented as `has…()` or value pairs derived from the rows, rather than boxed components. Consequently, no allocation is made per level, adhering to the `List<Long>` → `long[]` amendment from #835.

**No fixture regeneration.** The `dive_screenshots_fixture.parquet` already contains chunk-level `SizeStatistics`. PyArrow writes field 16 by default, and `write_page_index` only governs the copies per page. This approach covers every branch, including LIST, MAP, flat required, `max def 1`, and the `metric_*` columns, which do not have size statistics.

## Screenshots

image

`tags.list.element` from `size_statistics_test.parquet`, where the underlying data is `[[1, 2], [], None, [3]]`:
`l` toggles the histograms; they're collapsed by default since the facts pane doesn't scroll. The labels are the point — `tags null` (row 2, field absent) and `tags empty` (row 1, empty list) are separate buckets, and nothing else in the metadata tells them apart; `Nulls 2` lumps both in.
`Offset index` reads plain `present` here while `Column index` says `present · levels`, because this is an INT32 column with no unencoded byte size — the hint keys on the fields themselves, not on the index existing.

Some facts without the TUI:
image

One note on the docs screenshots in the PR: they're captured from `dive_screenshots_fixture.parquet`, which has no nulls or empty lists, so they show `chunk only` and a single full bar rather than the split above.

## Checklist


nitrogen404:
@gunnarmorling sorry but this one might be a little long one to review

gunnarmorling:
Thanks a lot, @nitrogen404; could you attach a screen shot for a quick first look?

nitrogen404:
@gunnarmorling I've included them in the PR description :) I would like to now explore #510

gunnarmorling (approved):
Hey there, thanks a lot. I reworked a few things to make the reported numbers a bit more actionable, rebased and merged. Good stuff!

PR #960, opened 2026-08-19
#9 write support stage 15: parseable created_by

### LINKED ISSUE #959: Writer: emit a parseable created_by identifier
Hardwood writes the bare string `hardwood` as the footer's `created_by`. parquet-java's `VersionParser` expects `<app> version <version> (build <hash>)` and cannot parse the bare form, so every parquet-java read of a hardwood-written file logs a `CorruptStatistics` parse warning, and under the PARQUET-251 heuristic parquet-java discards the deprecated `min`/`max` of a `BINARY` / `FIXED_LEN_BYTE_ARRAY` column whose writer it cannot identify.

Hardwood's bounds survive today only because it writes solely the modern `min_value`/`max_value`, which that heuristic does not gate — the correctness rests on a field the writer happens not to emit rather than on the identifier being right.

Surfaced by the write-path interop gate (#907). This is increment 15 of the writer plan in `_designs/WRITER_SUPPORT.md`.

**Scope**

- `WriterConfig.DEFAULT_CREATED_BY` becomes `hardwood version <project-version> (build <short-sha>[-dirty])`.
- `hardwood-core` gains the build-info plumbing it needs for that (filtered resource fed by the existing `capture-git-info` step); `hardwood-cli`'s `Version` delegates to it instead of carrying a second copy.
- The interop gate's footer assertions gain that parquet-java's `VersionParser` accepts the identifier.

gunnarmorling:
Increment 15 of the writer plan (`_designs/WRITER_SUPPORT.md`). Closes #959.

Hardwood wrote the bare string `hardwood` as its footer `created_by`. A reader that cannot parse that field cannot tell which implementation produced the file, so it assumes the worst and applies its writer-specific correctness workarounds — parquet-java's PARQUET-251 heuristic discards the deprecated `min`/`max` of a `BINARY` or `FIXED_LEN_BYTE_ARRAY` column outright. Hardwood's bounds survived only because it writes solely the modern `min_value`/`max_value`, which that heuristic does not gate; the outcome rested on a field the writer happens not to emit. The stage 14 interop gate surfaced it.

The default is now `hardwood version <project-version> (build <short-sha>[-dirty])`, e.g. `hardwood version 1.1.0-SNAPSHOT (build 899b129-dirty)`.

### Build plumbing

Identifying the build needs the version and commit at runtime, which `hardwood-core` had no way to obtain. `hardwood-cli` already solved exactly that with a filtered resource fed by the parent POM's `capture-git-info` step, so rather than stand up a second copy the plumbing moves down into core as `dev.hardwood.internal.BuildInfo`, and the CLI's `Version` becomes a formatter over it — identical output (`hardwood 1.1.0-SNAPSHOT (899b129-dirty)`), one source of truth. The resource is registered in a `resource-config.json` for native-image, which the CLI's copy never was. A build that cannot determine a component reports `unknown` in its place, which stays parseable.

### Gate extension

Per the extension contract in `_designs/WRITER_INTEROP_GATE.md`, the footer assertions gain `ParquetJavaReader.assertParseableCreatedBy`, run over the whole 84-case flat sweep. It asserts both halves of the defect:

- `VersionParser.parse` accepts the string, yielding application `hardwood` with a semantic version;
- `CorruptStatistics.shouldIgnoreStatistics` consequently returns false for `BINARY` and `FIXED_LEN_BYTE_ARRAY`.

Verified the assertion bites — the old identifier fails both:

```
created_by=hardwood
   PARSE FAILED: VersionParseException
   PARQUET-251 ignores BINARY stats: true
created_by=hardwood version 1.1.0-SNAPSHOT (build 899b129-dirty)
   app=hardwood ver=1.1.0-SNAPSHOT build=899b129-dirty semver=true
   PARQUET-251 ignores BINARY stats: false
```

`WriterConfigTest` additionally pins the shape offline against parquet-java's `VersionParser.FORMAT` regex, so a core-only build still catches a regression.

No `docs/content/` update: writer user docs are deliberately deferred to stage 17, and this changes the default value of a `WriterConfig` option that page will document along with the rest.

`./mvnw verify` green across all 13 modules.


gunnarmorling:
Also carries the fix for #961, reported locally while reviewing this branch.

`WriterDifferentialTest.duckDbReadsWrittenLogicalTypes` asserted a `TimestampType(isAdjustedToUTC = false)` column as an `Instant`. That column is a wall clock with no zone, so `ResultSet.getTimestamp` builds its `java.sql.Timestamp` in the JVM default zone and `toInstant()` recovers the written value only under UTC — the same rule `LocalTimestampTest` states for the read path (#568). CI and the dev container both run UTC, which is why it never went red here.

Pre-existing on `main`, unrelated to the writer stage — confirmed by reproducing it on `main` at `TZ=Europe/Berlin`. A sweep of the full `core` suite (2100 tests) under `Europe/Berlin` and `America/Los_Angeles` found it to be the only zone-dependent assertion. The fix compares the wall clock, and is verified green under UTC, `Europe/Berlin`, `America/Los_Angeles`, `Asia/Kolkata` (UTC+5:30) and `Pacific/Chatham` (UTC+12:45).

PR #968, opened 2026-08-19
#904 Fix group predicate resolution

### LINKED ISSUE #904: FilterPredicate on a group name silently resolves to one of its leaves and returns wrong rows
`FilterPredicate.isNull("address")` on a group (struct) column does not fail — it silently resolves to one of the group's leaf columns and answers a different question. The wrong rows reach the caller.

## Reproduction

Fixture: `id INT32`, `address` an optional group of `{street STRING, city STRING, zip INT32}`, four rows chosen to separate the cases.

```python
ty = pa.struct([("street", pa.string()), ("city", pa.string()), ("zip", pa.int32())])
rows = [
    {"street": "123 Main St", "city": "New York", "zip": 10001},  # 1 all present
    {"street": None, "city": "Los Angeles", "zip": 90001},        # 2 present, street null
    {"street": None, "city": None, "zip": None},                  # 3 present, ALL fields null
    None,                                                         # 4 struct itself null
]
t = pa.table({"id": pa.array([1, 2, 3, 4], pa.int32()), "address": pa.array(rows, ty)})
pq.write_table(t, "struct_null_semantics.parquet", write_page_index=True)
```

Reading it back through PyArrow confirms the encoding keeps rows 3 and 4 distinct (`is_valid == [True, True, True, False]`), so this is not a fixture artifact.

```java
for (String name : new String[]{ "address", "address.street", "address.zip" }) {
    try (ParquetFileReader reader = ParquetFileReader.open(InputFile.of(FIXTURE));
            RowReader rows = reader.buildRowReader()
                    .filter(FilterPredicate.isNull(name)).build()) {
        while (rows.hasNext()) { rows.next(); System.out.print(rows.getInt("id") + " "); }
    }
}
```

Observed:

```
isNull("address")        -> [ 2 3 4 ]     <-- wrong, expected [ 4 ]
isNull("address.street") -> [ 2 3 4 ]
isNull("address.zip")    -> [ 3 4 ]
```

`FilterPredicateResolver.resolve(isNull("address"), schema)` yields `IsNullPredicate[columnIndex=1]`, and leaf 1 is `address.street` — the predicate has become `address.street IS NULL`.

## Mechanism

`FileSchema` indexes leaf columns only: `columnPathToIndex` is built from leaf field paths, so `"address"` misses. `FilterPredicateResolver.resolveColumn` then falls back to the first column whose `fieldPath().topLevelName()` matches, which is whichever leaf happens to come first in schema order.

The fallback has no correct use case:

- Its comment says it exists so `"scores"` matches `"scores.list.element"`. But every leaf below a `LIST` has `maxRepetitionLevel > 0`, and `rejectRepeated` throws two lines later — so for the case it was written for, it resolves and then immediately fails.
- The surviving case is a group name, which is exactly the silently-wrong one. This is not limited to multi-leaf groups: for `optional group point { optional int32 x }`, `point` null and `point` present with `x` null are still different rows.

## Correct semantics, for reference

Parquet has no per-group null marker; group nullity lives in the definition levels of the group's leaves. For a leaf under `optional group address`, `def 0` means `address` is absent, `def 1` means `address` present and the leaf null. Because a leaf's presence implies its ancestors', **every** leaf under `address` necessarily agrees on whether `address` is absent — so one child's definition levels answer the question completely, and no values need to be read at all.

Note the read side already gets this right: `RowReader.isNull("address")` correctly reports struct nullity (`PqRowApiTest:153`). Only predicate resolution is wrong.

## Fixing it is not a pure deletion

Removing the fallback outright breaks `PredicatePushDownTest.testFilterOnRepeatedColumnIsRejected`, which asserts `IllegalArgumentException` with a message containing `"repeated"`. That message comes from `rejectRepeated` — reached *through* the fallback, since `"scores"` is not a leaf path. Without the fallback the error becomes `Column 'scores' not found in schema`, which is both a test failure and a worse message for a column that does exist.

Suggested shape: on an exact-path miss, consult `FileSchema.rootNode` (the `SchemaNode.GroupNode` tree, which does model non-leaf nodes) and throw a message that names what was found — a repeated group, or a group where a leaf is required — instead of guessing a leaf. That keeps both tests' intent and improves both messages.

## Scope

This issue is about not returning wrong answers. Actually supporting `address IS NULL` — resolving a predicate to a group node and evaluating it from one child's definition levels — is the feature, and it is also the predicate surface the `SizeStatistics` level histograms from #607 need before they can drive nested pushdown; see the note on #222.

Per CLAUDE.md the fixture should be generated from `tools/simple-datagen.py` rather than checked in by hand.

chlgusrbs0:
Fixes #904

## Authorship

Check this box to confirm:

## Checklist

## Verification

Docs were not updated because group predicates are not a supported or documented predicate surface.

All hardwood-core tests pass locally. The full `./mvnw clean verify` could not complete because Docker/Testcontainers is unavailable in the local environment.

### DIFF
--- a/core/src/main/java/dev/hardwood/internal/predicate/FilterPredicateResolver.java
+++ b/core/src/main/java/dev/hardwood/internal/predicate/FilterPredicateResolver.java
@@ -13,6 +13,7 @@ import java.time.Instant;
 import java.time.LocalTime;
 import java.util.List;
 
+import dev.hardwood.internal.schema.SchemaPathResolver;
 import dev.hardwood.metadata.ColumnOrder;
 import dev.hardwood.metadata.LogicalType;
 import dev.hardwood.metadata.PhysicalType;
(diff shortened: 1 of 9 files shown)


gunnarmorling:
Awesome, thanks a lot, @chlgusrbs0! Pushed one more commit for a few tweaks, merging now.

PR #969, opened 2026-08-19
#9 Write support stage 16: row-oriented RowWriter

gunnarmorling:
Stage 16 of the writer milestone (#9): the record-shaped entry point on top of the columnar core, mirroring the read side's `rowReader()`.

```java
try (ParquetFileWriter writer = ParquetFileWriter.create(out, schema)) {
    RowWriter rows = writer.rowWriter();
    for (Person person : people) {
        rows.writeRow(row -> row
                .setLong("id", person.id())
                .setString("name", person.name())
                .setDate("hired", person.hired())
                .setStruct("address", address -> address.setString("city", person.city()))
                .setList("phones", phones -> person.phones().forEach(phones::addString)));
    }
}
```

Design: [_designs/WRITER_ROW_API.md](https://github.com/hardwood-hq/hardwood/blob/9-write-support-stage-16-row-writer/_designs/WRITER_ROW_API.md).

## What it adds

`RowWriter`, obtained from `ParquetFileWriter.rowWriter()`, plus the `StructBuilder` / `ListBuilder` / `MapBuilder` interfaces. Fields are addressed by their user-visible name — the caller never spells the synthetic `list.element` / `key_value` segments the columnar API exposes. Setters mirror the reader's accessors one for one and convert logical-type values through `PhysicalValueConverter`, the inverse of `LogicalTypeConverter`.

Internally it is an adapter: records are staged into a `ColumnBatch` and submitted through `writeBatch`, so paging, dictionary encoding, compression, statistics and row-group cadence remain exactly those of stages 1–15.

## Rules

- A field left unset is null if `OPTIONAL`, and fails the record if `REQUIRED`. `setNull(name)` and a `null` value on any object-typed setter say the same thing.
- Conversion is exact: a sub-millisecond `Instant` into `TIMESTAMP(MILLIS)`, a `BigDecimal` whose rescale would round, an out-of-range `INT(8)`, a wrong-length `FIXED_LEN_BYTE_ARRAY` are all rejected rather than truncated. `setTimestamp` and `setLocalTimestamp` follow the column's `isAdjustedToUTC`, mirroring the reader's split.
- A record is staged in full or not at all — staging is checkpointed and rolled back if the filler throws, so a rejected value leaves the batch as it was and the caller can carry on. Without this a validation error would strand a half-populated batch whose columns disagree on their record count.
- A file is latched to one write API on first use. Interleaving rows and batches would merge two independent staging states into one row group.
- Shapes the core cannot produce (a nullable struct enclosing a repeated field, a legacy two-level list) are rejected at `rowWriter()`, naming the path, rather than later inside the shredder.

## Verification

- **Equivalence**: the same data written through `writeBatch` and through `writeRow` produces byte-identical files — flat with nulls, nested struct/list/map, and 2500 records spanning several staged batches. The columnar side deliberately supplies junk in the slots the levels mark absent, so identical bytes also prove no placeholder reaches the encoder.
- **Round trips** through `RowReader` for every physical type, every nesting shape, absent/empty/null-entry lists, and both batch-submission triggers.
- **Interop gate** (stage 14) grew a row-written axis — the single-entry-dictionary floor in every repetition shape plus a page- and row-group-crossing case per type — so parquet-java verifies this layer's output too. 119 cases, all green.
- **Rules and conversion** tests for each rejection above.

`./mvnw verify` green across all modules.

User documentation follows in stage 17, as planned.


gunnarmorling:
Added `WriterConfig.precisionLossPolicy(...)` (c0223f2) after review discussion of whether a higher-resolution value into a lower-resolution column should be an error.

`PrecisionLossPolicy.REJECT` (default) / `TRUNCATE`. The knob is scoped to **precision**, not value loss in general: a value the column cannot represent at all — a decimal wider than the declared precision, an `INT(8)` out of range, a wrong-length `FIXED_LEN_BYTE_ARRAY` — is rejected under either setting, since no narrowing preserves it. `TRUNCATE` drops digits rather than rounding: timestamps floor (what `Instant.toEpochMilli()` does with the same value), decimals rescale with `RoundingMode.DOWN`.

`REJECT` stays the default because it is the direction that keeps the option open — relaxing a rejection later is compatible, tightening one silently changes what past files meant — and because it matches PyArrow, the only comparable typed API, which raises `Casting from timestamp[us] to timestamp[ms] would lose data` unless `allow_truncated_timestamps=True`. The Java precedents that truncate silently (Avro conversions, Spark) sit above a Parquet API taking a raw `long`, where the decision had already been made.

The default's real cost is that `Instant.now()` carries microseconds on a modern JDK, so the rejection message now names all three ways out: `truncatedTo` at the call site, a finer column unit, or the policy.

PR #972, opened 2026-08-19
859 Dictionary-space predicate evaluation

shril:
## Authorship

Check this box to confirm:

## Checklist


shril:
@gunnarmorling this took sometime, and I have taken AI assistance to understand/write some code pieces.
Please let me know your review.

core/src/main/java/dev/hardwood/internal/predicate/BatchFilterCompiler.java
```diff
@@ -239,6 +240,7 @@ private static int leafColumnIndex(ResolvedPredicate leaf) {
             case ResolvedPredicate.IntPredicate p -> p.columnIndex();
             case ResolvedPredicate.FloatPredicate p -> p.columnIndex();
             case ResolvedPredicate.BooleanPredicate p -> p.columnIndex();
+            case ResolvedPredicate.BinaryPredicate p -> p.columnIndex();
             case ResolvedPredicate.IntInPredicate p -> p.columnIndex();
             case ResolvedPredicate.LongInPredicate p -> p.columnIndex();
             case ResolvedPredicate.IsNullPredicate p -> p.columnIndex();
```
kogupta: cant we collapse all these `case`s: `case ..., ..., ... -> <common operation>`? No Intellij warning in these sections?

shril: @kogupta I have collapsed it using `ResolvedPredicate.leafColumnIndex()`.

gunnarmorling (commented):
Hey @shril, thanks for that. Quick question upfront: is there any specific reason for limiting this to the flat reader?

shril:
Hi @gunnarmorling, this was mainly a scope decision rather than a fundamental limitation. The flat reader has a simple one-value-per-row batch, so dictionary IDs fit the existing matcher directly. The nested reader uses a different batch structure where values do not always map one-to-one to rows, so it needs additional wiring.

If the current scope looks good, I can add nested-reader support as a follow-up. If you’d prefer it in this PR, I’m happy to extend it here. Let me know which approach you prefer.

shril:
@gunnarmorling I have rebased with main branch. Let me know if you need me to do anything else.

core/src/main/java/dev/hardwood/internal/predicate/matcher/binary/BinaryEqBatchMatcher.java
```diff
@@ -0,0 +1,128 @@
...
+                }
+                else {
+                    if (matches == null) {
+                        matches = new long[(entries.length + 63) >>> 6];
+                        matches[firstMatch >>> 6] |= 1L << firstMatch;
+                    }
+                    matches[i >>> 6] |= 1L << i;
+                }
+            }
+        }
```
iifawzi, line 124: This scans every dictionary entry the first time a batch brings a new dictionary; however, little of the chunk is read afterwards. I looked at this while working on #1210, and I was leaning towards deciding entries lazily, from the values the batches actually reference.

Reading only part of a chunk is common. Page statistics can leave a handful of pages, including when a different column's index rules the rows out. Row masks narrow the pages that survive, and head(n) stops the read early. Take a dictionary of thousands of values: we compare every one of them, and the pages that survive reference only a few.

The lazy version: each batch decides only the entries its rows reference that aren't decided yet. Those answers stay for the rest of the chunk, and the extra pass stops once every entry is decided. Comparisons are then bounded by what the read touches rather than by the dictionary size. A small dictionary is fully decided within the first batch or two, so it costs the same as today.

The trade-off shows up on full scans of large dictionaries. Each entry is still compared only once, but until every entry is decided, each batch walks its entry indices once more before the row loop; it should still be faster than row by row comparisons without dictionaries, though

What do you think?

core/src/main/java/dev/hardwood/internal/predicate/matcher/binary/BinaryEqBatchMatcher.java
```diff
@@ -0,0 +1,128 @@
...
+
+import dev.hardwood.internal.predicate.BinaryBatchMatcher;
+import dev.hardwood.internal.reader.BatchExchange;
+import dev.hardwood.internal.reader.BinaryBatchValues;
+import dev.hardwood.internal.reader.Dictionary;
+
+/// Equality matcher for binary columns. Dictionary-encoded rows compare their
+/// integer entry IDs; plain rows compare their packed byte ranges without
+/// materializing a per-row `byte[]`.
+public final class BinaryEqBatchMatcher implements BinaryBatchMatcher {
```
iifawzi, line 20: Once #1210 lands, it would be great to extend dictionary support to all binary operators, not just equality. That needs an outcome per dictionary entry instead of a single entry ID, because a range, IN or NOT_EQ can match many entries. Each entry could be decided by the matcher's own per-value test, and each row would then check states[entryId]

iifawzi (commented):
Hi @shril, thanks for being consistent in taking care of that! 

I've been working on #1210, which brings binary predicates to the drain side on their own: EQ, NOT_EQ, the four orderings, IN and NOT IN, across all three Comparison modes (strings, fixed- and variable-width decimals). Once it lands, this PR can focus on what's specific to dictionary support: the index-retention contract so columns other than strings keep their entry indices, the column reader wiring, and evaluation in dictionary space on top of those matchers.

left a few comments on the current shape. One suggests deciding dictionary entries lazily, so a read that skips pages or stops early doesn't compare entries no surviving row references. The other is about extending dictionary support to the other operators once #1210 lands.

It would also be great to see RecordFilterBenchmarkTest results before and after. in #1210 I added couple of binary column cases to it, so it can be enough to run it against your implementation before and after to see the improvement, but If this PR lands first, extending the benchmark with a few binary cases would be nice to show it

In case it helps sparking ideas, here's the approach I explored before seeing your PR. A wrapper matcher takes a binary matcher as its delegate:
- A batch without a dictionary goes straight to the delegate, so the delegate's byte-comparison loop only ever sees plain batches and stays well optimised.
- For a batch with a dictionary, each entry its rows reference is decided once, using the delegate's per-value test, and stored per entry for the rest of the chunk.
- Each row then reads its entry's outcome, and a row without an entry, such as one from a page written without the dictionary, falls back to the delegate's per-value test.

Because entries are decided by the delegate's own test, this should work for every operator. 

That was the shape I had, but you may come up with a better approach, being always backed by RecordFilterBenchmarkTest results should decide which approach could fit better

core/src/main/java/dev/hardwood/internal/predicate/matcher/binary/BinaryEqBatchMatcher.java
shril: @iffawzi your approach is definitely better. The lazy version will surely give us some performance gains. I'll start working on this.

shril: Hi @iifawzi, Agreed. The eager scan does unnecessary work when pruning, row masks, or an early stop means only part of the chunk is read.
I’ll change this to cache a tri-state outcome per dictionary entry and resolve only previously unseen IDs referenced by each batch. The outcomes will remain cached for the rest of the column chunk. Plain-encoded and cross-chunk fallback rows will continue through the normal per-value matcher.
I’ll benchmark both sparse/early-stop and full-scan cases to verify the trade-off.

core/src/main/java/dev/hardwood/internal/predicate/matcher/binary/BinaryEqBatchMatcher.java
shril: @iifawzi Agreed again. Now that #1237 has landed, I’ll build dictionary evaluation as a wrapper around its binary matchers rather than maintaining a separate equality matcher. This will support EQ, NOT_EQ, the range operators, and IN while reusing each matcher’s comparison semantics. NOT IN will benefit through its existing NOT_EQ composition.

Since #1237 supersedes a substantial part of this PR and the implementation will be largely rewritten, I plan to open a fresh PR from current main and close #972 as superseded. I’ll include before/after RecordFilterBenchmarkTest results and link the replacement PR here. Please let me know if you would prefer that I update this PR instead.

shril:
I am closing this after going through the PR review.
New PR with suggested changes - https://github.com/hardwood-hq/hardwood/pull/1257
cc: @gunnarmorling @iifawzi

Issue #872, opened 2026-08-09
cli: consistent machine-readable output (`--format json`) for all introspective commands


gunnarmorling:
## Motivation

`hardwood dive` is for interactive exploration by a human. Every other command serves a different audience: quick ad-hoc use at a terminal, but primarily automated use — scripts, pipelines, CI checks, and AI coding agents. Today all of them emit only human-formatted text, so that second audience has to scrape ASCII tables.

The agent case is already explicit in the project. `docs/content/reference/cli.md` has a "Use with AI coding agents" section, and `skills/hardwood-cli/SKILL.md` ships as a published plugin. A large share of those 278 lines exists to teach a model how to read table layouts ("you get a table of `(RG, Index, Length, Value)`"). That is a compatibility shim for terminal formatting, paid for in tokens on every invocation.

Scripting pays a similar tax. `Sizes.format` emits `12.4 MB`, so anything wanting a byte count has to parse a human string and re-multiply. And the contract is accidental: the three most recent commits on this branch are table-formatting fixes (bottom border, min-content column floor), each of which would have broken a text-scraping consumer. The contract exists either way — it is just undeclared and coupled to layout.

## Current state

| Command | Flags (beyond `-f/--file`) | Output |
|---|---|---|
| `info` | — | 6 aligned `Key: value` lines |
| `schema` | `-F NATIVE\|AVRO\|PROTO` | native text / hand-built JSON / proto3 IDL |
| `footer` | — | 5 aligned `Key: value` lines |
| `print` | `-n -c -s -w -t --transpose -i -d` | `StreamedTable` ASCII, or transposed table per row |
| `convert` | `-F csv\|json` (required), `-o`, `-c`, `-n` | CSV, or a JSON array |
| `inspect columns` | — | ranked ASCII table |
| `inspect rowgroups` | — | heading line + ASCII table per row group |
| `inspect pages` | `-c`, `--no-stats` | label line + ASCII table + `Total` row |
| `inspect dictionary` | `-c` (required), `--limit` | label line + message lines + ASCII table |
| `dive` | `--max-dict-bytes`, `--log-file` | TUI |

Exit codes are 0 / 1 / 127 (`Main.java:22-26`); failures print plain text to stderr.

## JSON cannot be a serialization of the existing rows

The rendered rows are presentational and lossy, so a machine mode has to be built from the metadata objects, not from the `List<String[]>` each command already assembles:

[...]

PR #987, opened 2026-08-20
#877 Surface key-value metadata in hardwood info

### LINKED ISSUE #877: cli: surface file key-value metadata from `hardwood info`
## Problem

A Parquet file's key-value metadata is not reachable from any non-interactive command. `hardwood dive` shows it — `OverviewScreen` renders a `key/value meta (N)` section with per-entry selection and a detail modal (`OverviewScreen.java:216-240`, `KvMetadataFormatter`) — but `info` prints only format version, created-by, row-group count, row count, and the two aggregate sizes. A grep of `cli/command/` and `cli/internal/` for `keyValueMetadata` returns nothing.

This is the widest parity gap between `dive` and the introspective commands, and it lands on the most load-bearing content. Key-value metadata is where Arrow, Spark, and pandas store their own schemas (`ARROW:schema`, `org.apache.spark.sql.parquet.row.metadata`, `pandas`), so it is the first thing to look at when a file reads differently in two engines. That is precisely the debugging case `skills/hardwood-cli/SKILL.md` exists for, and today the skill has no way to get at it.

## Proposal

Surface key-value metadata from `hardwood info`.

Human output needs a length policy: these values are routinely kilobytes of embedded JSON, so printing them raw would bury the rest of the command's output. Suggested default is to list keys with truncated values and the byte length, with a flag to print one key in full:

```
Format Version:    2
Created By:        parquet-mr version 1.13.1
Row Groups:        12
Total Rows:        12582912
Uncompressed Size: 1.2 GiB
Compressed Size:   410.3 MiB

Key/Value Metadata (3):
  ARROW:schema                                  4.1 KiB  /////5AEAABAAAAAAAAKAAwABgAFAAgACgAAAAABBAAM…
  org.apache.spark.sql.parquet.row.metadata     1.8 KiB  {"type":"struct","fields":[{"name":"order_id"…
  pandas                                          912 B  {"index_columns":["__index_level_0__"],"colu…
```

Under `--format json` (#872) the whole map is emitted untruncated as an object — no policy needed, which is the point of that issue.

## Scope

- Key-value metadata section on `hardwood info`, truncated by default.
- A way to print a single entry in full (`--kv-key <name>` or similar).
- Full untruncated map under `--format json`.
- Docs under `docs/content/`.
- Worth a line in `skills/hardwood-cli/SKILL.md`: engine-specific schema mismatches start here.

Part of the `dive` / CLI parity pass, alongside #878 (aggregates in `footer`) and #879 (remaining `inspect` gaps). Related: #872 (machine-readable output).

nitrogen404:
Closes #877

- Surfaces key-value metadata (ARROW:schema, Spark/pandas row-metadata JSON, etc.) in hardwood info. Long values are truncated with byte length shown 
- --kv-key <name> prints one entry in full for piping elsewhere 
- Full untruncated JSON output under `--format json` is deferred to #872 (designing that flag consistently across all commands).
## 
image

`{"type":"struct","fields":[{"name":"order_id","type":"long","nullable":false,"metadata":{}}
  {"name":"customer","type":"string","nullable":true,"metadata":{}} {"name":"amount","type":"double","nullable":true,"metadata":{}}]}`

- [x] ./mvnw clean verify passes
- [x] Commit prefixed with #877
- [x] Covered by tests (6 local + 5 S3)
- [x] Docs updated (cli.md, SKILL.md)


gunnarmorling:
Thanks a lot, @nitrogen404! I'm wondering how much truncation makes sense; I know it's stated in the issue, and we do it in dive, but perhaps that's not the right thing? Have you explored with not truncating? How is usability?

nitrogen404:
@gunnarmorling I tested this locally on a 14' macbook vscode terminal.  Without truncation ARROW:schema alone wraps 10+ lines vs 6 for the whole summary. It doesn't hide the summary in a single run, but it dominates the screen

I think the 60-char default is more better usabilitywise; --kv-key covers the "I need the full value" case.

image
image

gunnarmorling:
Ok, I see; yeah, let's go that way then. Ultimately, I think we should apply some smarts to the contents and e.g. render JSON indented and also decode the Arrow schema (i.e. some special support for well-known keys like `ARROW:schema` would make sense.

docs/content/reference/cli.md
```diff
@@ -113,6 +116,26 @@ and as a comment in Protobuf:
...
+Key/Value Metadata (2):
+  ARROW:schema                               4.1 KiB  /////5AEAABAAAAAAAAKAAwABgAFAAgACgAAAAABBAAM…
+  org.apache.spark.sql.parquet.row.metadata  1.8 KiB  {"type":"struct","fields":[{"name":"order_id"…
+```
+
+Pass `--kv-key <name>` to print one entry's value in full, untruncated, with no
+other output — safe to pipe into another tool:
+
+```shell
+hardwood info -f data.parquet --kv-key ARROW:schema | base64 -d | xxd | head
```
gunnarmorling, line 136: Nice!

gunnarmorling:
LGTM, merging. Thanks a lot, @nitrogen404!

gunnarmorling:
Applied.

PR #988, opened 2026-08-21
#976 Codec and encoding breadth: the remaining codecs, and the four optional encodings behind one per-column policy

### LINKED ISSUE #976: Remaining write codecs and optional delta / byte-stream-split encoders
Delivery stage 19 of `_designs/WRITER_SUPPORT.md` (part of #9), targeting `1.1.0.Beta1`.

## Scope

The writer produces `UNCOMPRESSED` and `ZSTD` today; `WriterConfig.codec` accepts the rest of
`CompressionCodec` only to reject it at writer creation. This stage completes the codec
breadth (Snappy, GZIP, LZ4_RAW, Brotli, LZO as availability allows), each gated on its library
being present the way ZSTD already is, and adds the optional `DELTA_*` and `BYTE_STREAM_SPLIT`
encoders as an opt-in encoding choice rather than an automatic one.

## Why it sits inside the gated series

Every codec is a new axis value on the stage 14 write-path interop gate, and the optional
encodings are ones that gate's readers (parquet-java, PyArrow) must read back. Shipping this
breadth before the gate is declared done means each new codec and encoding arrives proven
against a strict reader rather than after the fact. It also lands before the writer's user
documentation (stage 20), so the codec and encoding options in `WriterConfig` are documented
once, in their final form.

## Notes

- The gate's codec axis currently has two values; it grows with this stage, and the interop
  matrix grows with it.
- Encoder selection should slot into the per-chunk decision point established by #975
  (row-group-global dictionary selection), which is why that stage comes first.

## Done when

- Every codec `WriterConfig.codec` accepts is either produced or rejected with a clear reason at
  writer creation, with no silent gaps.
- The optional encodings are selectable, and the interop gate covers each new codec and encoding
  across the shapes it already sweeps.

gunnarmorling:
Delivery stage 19 of [_designs/WRITER_SUPPORT.md](https://github.com/hardwood-hq/hardwood/blob/main/_designs/WRITER_SUPPORT.md), in the two increments it was planned as. Closes #976.

Settled in [`_designs/WRITER_CODECS_AND_ENCODINGS.md`](https://github.com/hardwood-hq/hardwood/blob/main/_designs/WRITER_CODECS_AND_ENCODINGS.md), which the first commit adds — reviewing that first will make the other two read faster.

## 19a — the remaining codecs

The writer produced two codecs where the reader decompressed six, and refused the rest with one message: "not yet supported". That was wrong about half of what it covered. Every member of `CompressionCodec` now resolves to a compressor or to a refusal of its own:

| Produced | Refused, permanently |
|---|---|
| `UNCOMPRESSED`, `GZIP` (JDK, no new dependency), `SNAPPY`, `ZSTD`, `LZ4_RAW`, `BROTLI` | `LZ4` — the Hadoop framing the format deprecated in favour of `LZ4_RAW`; still read, since files written before the deprecation exist. `LZO` — no maintained JVM implementation; refused in both directions. |

`SNAPPY`, `LZ4_RAW` and `BROTLI` compress in the raw block form Parquet specifies — the page header already carries the uncompressed size their framed variants would repeat. The switch is exhaustive, so a codec added to the enum is a compile error here rather than a silent fall-through.

Framing is the whole risk, and framing is invisible to a round trip: a wrapped block decodes perfectly against the compressor that wrote it while being the wrong bytes for every other reader. The gate's codec axis therefore carries four codecs it did not have, and the differential suite reads all six back through DuckDB.

## 19b — the optional encodings

`internal/encoding/` held decoders for `DELTA_BINARY_PACKED`, `DELTA_LENGTH_BYTE_ARRAY`, `DELTA_BYTE_ARRAY` and `BYTE_STREAM_SPLIT` and encoders for none of them. Each now has its inverse, selected by a `ColumnEncoding` set file-wide or per dotted leaf path:

```java
WriterConfig config = WriterConfig.builder()
        .encoding("ts", ColumnEncoding.DELTA_BINARY_PACKED)
        .encoding("readings.list.element", ColumnEncoding.BYTE_STREAM_SPLIT)
        .build();
```

**This replaces `WriterConfig.enableDictionary` rather than sitting beside it.** Dictionary-or-`PLAIN` is the whole of what `AUTO` decides, so the boolean was a two-valued encoding policy named before there were encodings to name it with, and `enableDictionary(false)` is a file-wide `PLAIN` exactly. Nothing has shipped that depends on it — the writer's first release is `1.1.0.Beta1`, which this precedes — and stage 20 then documents one concept instead of two.

The encodings are **selected, never inferred**. `AUTO`'s comparison works because both its candidates' sizes fall out of what the buffer already retains; a delta encoding's size is a property of value order that nothing measures, and byte-stream-split changes no page's size at all — its payoff belongs to the codec that follows it.

Validation runs at writer creation, before the output is touched: a path matching no leaf column, a policy illegal for a column's physical type, or a file-wide default no column of the schema can carry. `ColumnMetaData.encodings` also becomes what the chunk actually uses — it listed `PLAIN` unconditionally, true while a chunk was either `PLAIN` or a dictionary whose body is, and false of one written with an optional encoding.

## Two invariants worth reviewing closely

Both encode *perfectly* while being wrong, which is why each has an assertion beyond symmetry:

1. **Deltas wrap at the column's width.** `DeltaBinaryPackedDecoder` recovers the value only because it accumulates in a `long` and narrows back. The existing decoder needed no change for this — the open question the design left — which the alternating `MIN_VALUE`/`MAX_VALUE` cases pin for both `INT32` and `INT64`.
2. **The block minimum is signed; the residues are unsigned.** Comparing the minimum unsigned makes every negative delta look enormous and drives the bit width to the full type. That round-trips exactly and costs an order of magnitude more — 100 identical values encoded to 584 bytes instead of ~120. Only a size assertion sees it, so the suite has them.

## Independent readers

- **Interop gate** — 152 tests, +10 for every legal (encoding, type) pair, all read back through parquet-java. Now asserts **PARQUET-246** alongside PARQUET-251: `DELTA_BYTE_ARRAY` is the second path parquet-java gates on who wrote the file, and a parseable `created_by` is what keeps our output off it.
- **DuckDB** — all six codecs, and the delta encodings.

Two gaps in the *readers*, recorded rather than papered over:

- parquet-java cannot read `BROTLI` at all: `CompressionCodecName.BROTLI` names `org.apache.hadoop.io.compress.BrotliCodec`, which is in neither parquet-java 1.17.1 nor Hadoop, only in the unmaintained `com.github.rdblue:brotli-codec` (native binaries for a few platforms). Putting it on the gate's classpath would make the verdict architecture-dependent. **This is a pre-existing hole**, not one the write path introduced — `large_string_map.brotli.parquet` is the corpus's only `BROTLI` file and was already skipped in the comparison suite, misfiled under "Avro reader decoding issues"; the skip list now names the real cause. DuckDB covers the codec instead, and the gate asserts the class's absence so a parquet-java that gains it says so.
- DuckDB 1.4.4 accepts `BYTE_STREAM_SPLIT` for `FLOAT`/`DOUBLE` only, rejecting the `INT32`, `INT64` and `FIXED_LEN_BYTE_ARRAY` forms parquet-format 2.10 added. Those three are covered on the gate through parquet-java, so every pair has an independent reader — what varies is which one.

## Also here

- The bit packer moves out of `RleBitPackingHybridEncoder` into a shared helper, a 32-value miniblock being the same layout as an eight-value group. It gains a `long` domain and widths to 64, which an `INT64` column with type-spanning deltas actually reaches.
- `FlatWriteBenchmark`'s codec dimension widens to the five both writers produce. That needed lz4-java on the benchmark module — `core` declares the codec libraries `optional`, so they do not reach a consumer that has not asked for them. A smoke run surfaced it; both writers land within ~0.03% on size for every codec.

## Not here

User documentation. Stage 20 documents the settled writer surface — `codec` and `encoding(path, …)` among it — in one `how-to` guide and one `reference` page, so the options are described once in their final form.

## Verification

`./mvnw clean verify` green across the reactor: 2367 core tests (+112 for the policy, +73 for the encoders and packer), 795 in the gate.


gunnarmorling:
Review addressed in cf30ab9c. All ten items, plus the decision.

**Decision — A.** `WriterConfig.encodingFor(String)` is package-private. `ParquetFileWriter` and the tests are both in `dev.hardwood.writer`, so nothing else moved; one keyword to promote if a real external caller turns up.

**The guard.** `BitPacker.pack(…)` rejects a group whose `count * bitWidth` is not a multiple of eight, and a `bitWidth` its value type cannot hold, in both overloads. You were right that the precondition was documentation only — neither current caller can trip it, which is exactly why the next one would have got a short value section rather than a failure.

**Coverage — the three gaps.**

- `ByteStreamSplitEncoderTest` is new. `NaN`, both signed zeros and the infinities asserted on their **raw bits** (a stream landing in the wrong place still yields a plausible-looking NaN or a zero of the wrong sign), every byte width including `FIXED_LEN_BYTE_ARRAY` at width 3, and the stream layout pinned directly rather than only round-tripped. `WriterEncodingPolicyTest`'s float and double fixtures now start with those bit patterns too.
- All-null pages: a sweep over every legal (policy, type) pair with no present value at all — the `count == 0` branch each encoder carries. Matches your local probe; it round-trips, so this is the tripwire rather than a fix.
- Nested: a repeated leaf inside a list, swept over every legal pair, **reading the values back** — that shape is also what puts a repetition-level stream ahead of the values, which the flat sweeps never produce.

`WriterEncodingPolicyTest` goes 112 → 160 tests.

**Documentation.** `ColumnChunkBuffer` no longer calls the non-dictionary value section `PLAIN`; the supersession notes in both dictionary designs now say complete; the stage 19 design drops the change-terms prose and the call-site list; and the rationale in the new public JavaDoc moves to `_designs/` before stage 20 copies that wording into `docs/content/`.

Nits fixed. `./mvnw clean verify` green — 2541 core, 795 gate.

PR #1020, opened 2026-08-23
#1017 Writer: emit nan_count in column chunk statistics for float, double and FLOAT16

### LINKED ISSUE #1017: Writer: emit nan_count in column chunk statistics for float, double and FLOAT16
The writer never emits `Statistics.nan_count` (field 9) for floating-point columns. The spec makes it mandatory under the type-defined order (`parquet.thrift`, `ColumnOrder.TYPE_ORDER`):

> When writing page or column chunk statistics for columns with TYPE_ORDER order, the following rules must be followed:
> - The nan_count field must be set for floating point types, even if it is zero.
> - If the nan_count field is set, min and max statistics fields, when present, must not contain NaN values and must be computed from non-NaN values only. This signals to readers that the min and max statistics are reliable for non-NaN values.

Beyond conformance, this is what makes floating-point pruning recoverable. `nan_count` is already parsed (#607) and #898 specifies how the filter should consume it; #1016 makes `gt` / `gtEq` / `notEq` on `FLOAT` / `DOUBLE` / `FLOAT16` stop pruning because an absent `nan_count` forces a reader to assume NaNs may be present. Until the writer records it, files Hardwood writes itself lose that pruning against Hardwood's own reader — a self-inflicted regression that #898 alone cannot undo.

## Current state

- `FloatStatisticsCollector`, `DoubleStatisticsCollector` and `Float16StatisticsCollector` skip NaN when accumulating bounds (correct — the second rule above already holds), but never count it.
- An all-NaN chunk leaves `hasValues` false, so no bounds are written — also already conformant, but indistinguishable from "this writer wrote no bounds at all" without `nan_count`.
- `ColumnChunkBuffer:419-421` forwards `statistics.nanCount()`, which is always `null`.
- `StatisticsWriter` emits fields 3–8; there is no field 9 arm.

## Scope

- Count NaN values in the three floating-point collectors and carry the count into the `Statistics` they produce.
- `StatisticsWriter`: emit field 9 as an `i64` when the statistics carry a count.
- Always set it for `FLOAT`, `DOUBLE` and `FLOAT16` chunks, including when it is zero — a recorded `0` is the only thing that proves a chunk is NaN-free. Never emit it for any other type.

Out of scope: `ColumnIndex.nan_counts` (field 8) and per-page statistics — the writer emits neither a page index nor `DataPageHeader.statistics` today; the per-page count belongs with that work. Writing the `IEEE_754_TOTAL_ORDER` column order (which also makes `nan_count` mandatory) is its own change.

## Tests

- Floating-point chunk with no NaN → `nanCount() == 0`, bounds unchanged.
- Chunk with NaN among finite values → exact count, bounds still spanning the non-NaN values only.
- Chunk mixing nulls and NaN → `nanCount` counts only NaN, `nullCount` only nulls.
- All-NaN chunk → `nanCount == numValues - nullCount`, no bounds written.
- Non-floating-point chunk → field absent.
- Round-trip through `StatisticsReader` and the public `Statistics` record, plus a `parquet-java-compat` check that an external reader accepts the field.

Related: #1016 (the pruning this restores), #898 (the reader-side consumer), #607 (parsing).

nitrogen404:
Fixes #1017.

The writer never set `Statistics.nan_count` (field 9) for FLOAT/DOUBLE/FLOAT16 columns even though the specification mandates it under TYPE_ORDER, even when zero, since a recorded zero is the only evidence that a chunk contains no `NaN` values. Without this setting, Hardwood’s own reader assumes the presence of NaNs and disables gt/gtEq/notEq pruning on files that Hardwood has written itself.

The three floating-point collectors already correctly excluded NaN from min/max calculations but they simply never counted it. Added the count to each collector and a field-9 arm to StatisticsWriter. Verified forward compatibility by adding a footer with `nan_count` set which decodes cleanly against a pre-nan_count parquet-format-structures version (an unknown field that is safely skipped).

The following are out of scope (as per the issue): `ColumnIndex.nan_counts` (page-level) and IEEE_754_TOTAL_ORDER.


gunnarmorling:
LGTM, merged. Thanks a lot, @nitrogen404!

Issue #257, opened 2026-04-18
ColumnWorker.close() must join VThreads and in-flight decode tasks before returning


gunnarmorling:
## Summary
`ColumnWorker.close()` (`core/src/main/java/dev/hardwood/internal/reader/ColumnWorker.java`) currently sets the `done` flag, finishes the exchange, unparks the retriever and drain VThreads, and returns immediately. It does **not** join the VThreads, and it does **not** track or await in-flight decode tasks submitted to `decodeExecutor` via `CompletableFuture.runAsync(...)`. As a result, after `close()` returns, asynchronous work can still be touching `PageInfo.pageData()` slices over chunk byte buffers held by the underlying `InputFile`.

## Why this isn't a crash today
For all current `InputFile` implementations the underlying memory is GC-rooted via slice parent references, so in-flight decodes pin the byte buffers they're reading from:

- `MappedInputFile.close()` is a no-op; `MappedByteBuffer` is released by GC.
- `S3InputFile.close()` is a no-op (the `HttpClient` is owned by `S3Source`); direct `ByteBuffer` instances are cleaned via `Cleaner` only after GC.
- `ByteBufferInputFile` doesn't release its caller-supplied buffer.

So no user can hit a JVM crash with the code as it ships in 1.0.0.CR1.

## Why it must be fixed before 1.0.0.Final
The invariant *"InputFile.close() must not eagerly release memory an in-flight decode might still be reading"* is undocumented, unenforced, and surprising. Once 1.0.0.Final solidifies the contract, anyone implementing a new `InputFile` (HDFS, GCS, Azure, encrypted local file, future eager-unmap optimization on `MappedInputFile` via `Cleaner.invokeCleaner`, etc.) will silently introduce a SIGSEGV the moment they wire up cleanup in `close()`.

Secondary motivations:

- Deterministic shutdown — tests cannot today assert "no in-flight work after close".
- JFR/profiling sees mystery decode tasks attributed to closed readers.
- Interaction with the new `RowGroupIterator` cache eviction (commit TBD) becomes cleaner: after `close()` returns, no decode task can still hold chunk references.

## Proposed implementation
Two changes inside `ColumnWorker`:

### 1. Track decode futures
```java
private final Set<CompletableFuture<Void>> inFlightDecodes =
        ConcurrentHashMap.newKeySet();
```

In the retriever's submission site (currently `CompletableFuture.runAsync(() -> decode(slot, pi, rdr), decodeExecutor);`):

```java
[...]

PR #1028, opened 2026-08-25
#1027 Release a parked drain thread on teardown instead of waiting out the poll interval

### LINKED ISSUE #1027: Reader close() blocks ~11 ms per projected column when a read is torn down with data outstanding
_Via Claude after my review_

Tearing a reader down while pages are still outstanding costs roughly 11 ms per projected column.
Measured on 1.0.0.Final and on current `main`, warm JVM, shared `HardwoodContext`, after consuming a
single small batch:

| projected columns | `ColumnReaders.close()` | `RowReader.close()` |
| --- | --- | --- |
| 1 | 0.04 ms | — |
| 2 | 11.4 ms | — |
| 7 | 70.5 ms | 1.4 ms |
| 16 | 195.1 ms | 122.3 ms |

For a long scan this amortises away. For short-lived reads — a small batch, a `head(n)`, a point
lookup touching one page per column — teardown dominates total latency by one to two orders of
magnitude: single-digit milliseconds of useful work followed by 60–195 ms of `close()`.

Holding the `ParquetFileReader` open does not help, and neither does sharing a `HardwoodContext`.

Two preconditions, both of which make this easy to miss:

- **Data must still be outstanding.** If the file fits in the first batch, each drain thread reaches
  EOF and calls `finish()` itself before `close()` arrives, nothing is parked, and there is no delay.
- **It does not reproduce below about 8 columns**, because auto batch sizing decides whether a drain
  is parked at all when `close()` lands. Note the 7-column `RowReader` row above: 1.4 ms, which looks
  immune and is not.

### Mechanism

`BatchExchange` observes its `finished` flag only after a 10 ms timed wait expires — `takeBatch()` uses
`freeQueue.poll(10, MILLISECONDS)` (`BatchExchange.java:139`), `publish()` uses
`readyQueue.offer(batch, 10, MILLISECONDS)` (`:155`), and the consumer `poll()` uses
`readyQueue.poll(10, MILLISECONDS)` (`:194`). `READY_QUEUE_CAPACITY` is 2 (`:34`).

`finish()` (`:163-165`) only sets the flag. Nothing releases a drain thread already parked in one of
those waits, so it cannot observe the flag until its current window elapses. `ColumnWorker.close()`
(`ColumnWorker.java:208-216`) calls `finish()` and then joins the retriever and drain threads, so it
inherits that wait — and `ColumnReaders.close()` (`ColumnReaders.java:272-279`) closes the per-column
readers sequentially, so the windows add rather than overlap.

**Both reader families are affected, at different lines.** This is the part worth being careful about:

- **Detaching mode** (`ColumnReader.java:865, 879`), used by the columnar path: `freeQueue` is null, so
  `takeBatch()` allocates and never blocks, and the drain parks in `publish()`.
- **Recycling mode** (`FlatRowReader.java:270`, `NestedRowReader.java:139`), used by the row path: the
  drain parks in `takeBatch()` waiting for a holder from the pool. `FlatRowReader.close()` does call
  `recycle(...)` + `drainReady()`, which would refill that pool — but only at `:935-941`, *after*
  `:930-934` has already joined the workers. Too late to release anything.

### Relationship to existing issues

- **#659** (*Performance regression for reader close() methods in v1.0.0.Beta2*) is a sibling, and this
  is outside its scope. That thread was about *repeated* `close()` on the row reader, with first-close
  measured at 15–63 µs, addressed by the idempotency guard in #661. This is the *first* close, three
  orders of magnitude larger, and the idempotency guard does not touch it.
- **#257** (*ColumnWorker.close() must join VThreads and in-flight decode tasks before returning*) is
  the behaviour to preserve. The join exists so a decode task cannot still be reading a buffer the
  caller is about to free. Any fix must keep it and remove only the waiting.
- **#375** (*ColumnWorker.close() returns early when the calling thread is interrupted*) is related but
  distinct, and is neither fixed nor worsened by what is proposed below.

### Suggested fix

Give `BatchExchange` a teardown-only `abort()` that sets `finished` and then releases whichever
blocking point applies — returning `readyQueue` leftovers to `freeQueue` in recycling mode (which
frees ready capacity *and* makes the free pool non-empty) and clearing `readyQueue` in detaching mode
— and call it from `ColumnWorker.close()` in place of `finish()`. `finish()` stays as it is, so the
"any batch published before `finish()` remains visible" guarantee documented on `poll()` is untouched;
`abort()` is reachable only from teardown, where no consumer remains. The #257 join is kept.

Measured with that change: `ColumnReaders.close()` at 16 columns 195.1 → 0.60 ms, at 7 columns
70.5 → 0.19 ms; `RowReader.close()` at 16 columns 122.3 → 0.81 ms. Scan throughput unchanged.

**Clearing `readyQueue` alone is not sufficient**, which is worth stating because it is the obvious
first attempt. It merely moves the row path's wait from `publish()` to `takeBatch()`: the recycling
pool holds only `READY_QUEUE_CAPACITY + 1` = 3 holders, so discarding them leaves `freeQueue`
permanently empty. Measured `RowReader.close()` at 16 columns: 122.3 ms unpatched, **122.3 ms with the
clear-only version**, 0.81 ms with both branches.

I have opened a PR with the fix plus tests — a `RowReader`-path latency test (the pre-existing tests only
cover `ColumnReaders`, so they pass even with the insufficient version) and deterministic
`BatchExchange` unit tests for both modes. Full `core` suite is green at 2650 tests.

That said, `abort()` is arguably a workaround for a design smell: the three timed loops exist purely to
observe a `volatile boolean`, which is polling in place of a wake-up. Replacing the two
`ArrayBlockingQueue`s with a `ReentrantLock` + two `Condition`s and having `finish()`/`signalError()`
`signalAll()` would delete all three windows, remove the need for two divergent stop methods, and make
`poll()`'s guarantee structural rather than a comment. That is a bigger change and per CONTRIBUTING
probably wants a `_designs_/` note first — I'd rather follow your preference than guess.

ZhaoMJ:
Fixes #1027.

_The following PR description is generated by Claude after my review._

Reader teardown blocked roughly 11 ms per projected column whenever a read was torn down with pages
still outstanding — 70 ms at 7 columns, 195 ms at 16. The cause was not the teardown work but the
wake-up: `BatchExchange` only checks its `finished` flag after a 10 ms timed wait expires, and
`finish()` did nothing to release a drain thread already parked in one of those waits.
`ColumnWorker.close()` joins that thread, so it inherited the wait, and `ColumnReaders.close()` closes
columns one at a time, so the waits added instead of overlapping.

Both reader families are affected, at different lines, and that turned out to be the crux. In
**detaching** mode — the columnar path — `freeQueue` is null, so `takeBatch()` never blocks and the
drain parks in `publish()` on `readyQueue.offer(...)`. In **recycling** mode — the row path — the drain
parks in `takeBatch()` waiting for a holder from the pool. `FlatRowReader.close()` does call
`recycle(...)`/`drainReady()`, which would refill that pool, but only after it has already joined the
workers, so it cannot release them.

`abort()` sets `finished` and then clears whichever blocking point applies: it returns `readyQueue`
leftovers to `freeQueue` in recycling mode, freeing ready capacity *and* making the free pool
non-empty, and clears `readyQueue` in detaching mode. `finish()` is left exactly as it was, so the
guarantee documented on `poll()` — that a batch published before `finish()` stays visible — is
unaffected; `abort()` is reachable only from teardown, where no consumer remains. The join introduced
by #257 is kept, since a decode task must not still be reading a buffer the caller is about to release.
Only the waiting is removed.

| | before | after |
| --- | --- | --- |
| `ColumnReaders.close()`, 7 columns | 70.5 ms | 0.19 ms |
| `ColumnReaders.close()`, 16 columns | 195.1 ms | 0.60 ms |
| `RowReader.close()`, 16 columns | 122.3 ms | 0.81 ms |

Scan throughput is unchanged within run-to-run noise.

Worth recording, because it is the obvious first attempt: **clearing `readyQueue` alone does not fix
the row path.** It moves the wait from `publish()` to `takeBatch()` rather than removing it — the
recycling pool holds only `READY_QUEUE_CAPACITY + 1` = 3 holders, so discarding them leaves `freeQueue`
permanently empty. `RowReader.close()` at 16 columns measured 122.3 ms unpatched and 122.3 ms with the
clear-only version. Both branches are needed.

One benign race, noted rather than fixed: in recycling mode, if the drain's timed poll expires at the
same instant as `abort()`, it exits via the `finished` check and returns `null` rather than taking the
recycled holder. Both exits are immediate, so latency is unaffected — it is why the threaded tests
assert release rather than the returned value.

Investigated and patched with Claude.

## Authorship

## Checklist

- [x] If this PR adds or changes user-facing behaviour, `docs/` has been updated — n/a, `abort()` is in
      an `internal` package and no user-facing behaviour changes


gunnarmorling:
Thanks a lot, @ZhaoMJ! Gonna review this one asap; in the meantime, could you rework the commit so it's using an email associated to your GH account as the author (i.e. no octocat commit). Thx!

ZhaoMJ:
@gunnarmorling recommited using my personal email address.

gunnarmorling:
Thanks a lot, @ZhaoMJ! I've pushed one more commit for reworking and simplifying this, just interrupting the drain thread does the trick (I verified shut-down still is fast, your tests still pass). Gonna squash everything and merge. Thx!

ZhaoMJ:
@gunnarmorling Nice! I've verified the fix too! Thanks!

gunnarmorling:
Perfect, thank you. This will ship as part of the upcoming 1.1.0.Beta1 release, either later this week, or early next week.

PR #1030, opened 2026-08-25
#977 Support IS NULL / IS NOT NULL predicates on group columns

### LINKED ISSUE #977: Support IS NULL / IS NOT NULL predicates on group columns
## Summary

`FilterPredicate.isNull("address")` on a struct column is rejected (#904, #968): `FileSchema` indexes leaf columns only, so a group has nothing to attach a predicate to. The information is on disk and the read side already exposes it — `RowReader.isNull("address")` reports struct nullity correctly (`PqRowApiTest:153`). Only the predicate path cannot ask the question.

## Why this needs no value decoding

Parquet has no per-group null marker; group nullity lives in the definition levels of the group's leaves. Let `d` be the definition level at which the group is defined. For any leaf below it, `def < d` means the group is absent (the group itself is null, or an ancestor is) and `def >= d` means it is present. A leaf's presence implies its ancestors', so **every** leaf under the group agrees — one child's definition levels answer the question completely, and no values are decoded.

## What has to change

**1. Predicate surface.** `columnPathToIndex` is built from leaf field paths (`FileSchema:47`), so there is no `ColumnSchema`, and no column index, for a group node. `ResolvedPredicate.IsNullPredicate` / `IsNotNullPredicate` carry a bare `int columnIndex` (`ResolvedPredicate:68-69`) that the projected-column and multi-file remappings rewrite. A group predicate has to resolve to a *(leaf column index, definition level `d`)* pair instead, the leaf being any non-repeated leaf below the group — preferably the cheapest one to decode.

**2. Record-level evaluation.** `RecordFilterCompiler` / `NullBatchMatcher` test a leaf's own validity. Group nullity is `def < d` on the chosen leaf: a different test over the same level stream.

**3. Pushdown, where the statistics are present.** `Statistics.nullCount` cannot answer this — on a leaf under an optional struct it conflates "the struct is null" with "this field is null". The definition-level histograms parsed by #607 can, and nothing consumes them today:

- per chunk: `SizeStatistics.definitionLevelHistogram()`
- per page: `ColumnIndex.definitionLevelHistogram(int pageIndex)`

For a chunk or page, `sum(hist[0..d-1])` is the number of rows where the group is absent and `sum(hist[d..maxDefinitionLevel])` the number where it is present. `IS NULL` drops the unit when the first sum is zero, `IS NOT NULL` when the second is. Both accessors return `null` when the writer omitted the structure; pushdown is then skipped and record-level evaluation still yields correct rows, exactly as the existing statistics paths degrade.

## Scope

- **Non-repeated groups only.** The histograms count values per level, not records, so the arithmetic above is exact only when the chosen leaf has `maxRepetitionLevel() == 0`. Groups below a `LIST` or `MAP` stay rejected, consistent with `rejectRepeated`.
- **`isNull` / `isNotNull` only.** Comparison predicates on a group stay rejected — a node with no value has nothing to compare.
- A group whose entire ancestor chain is `REQUIRED` makes `IS NULL` constant-false and `IS NOT NULL` constant-true, which is worth folding rather than evaluating (see #923).

## Acceptance criteria

- `FilterPredicate.isNull("address")` and `isNotNull("address")` resolve on a non-repeated group and return the rows where the struct itself is / is not null, distinguishing "struct present, all fields null" from "struct null".
- Row groups and pages are dropped from the footer when the definition-level histograms are present and prove no match; the same query returns the same rows when they are absent.
- The "Predicates apply to leaf columns only" limitation in `docs/content/how-to/query-controls.md` is updated to record what is supported.
- Per CLAUDE.md, the test fixture separating the two null cases is generated from `tools/simple-datagen.py` rather than checked in by hand.

## Context

- #904 / #968 — the rejection this replaces. Before it, a group name silently resolved to its first leaf and returned wrong rows.
- #607 (closed) parsed the histograms; this would be their first consumer. Fields documented in `_designs/SIZE_STATISTICS_AND_NAN_COUNTS.md`.
- #222 is the neighbouring layer — record-level filtering on nested *leaf* columns such as `address.zip`. The two compose; neither blocks the other. See [the cross-reference comment there](https://github.com/hardwood-hq/hardwood/issues/222#issuecomment-5253509763) for the layering.

MovinduJay:
## Summary

Adds support for `IS NULL` and `IS NOT NULL` predicates on non-repeated struct group columns.

The implementation:
- resolves group predicates using a non-repeated leaf column and the group's definition level
- distinguishes a null struct from a present struct whose child fields are null
- uses definition-level histograms for row-group and page-level pruning
- falls back conservatively when histogram data is unavailable or invalid
- adds tests and documentation for the new behavior

Fixes #977

## Authorship

Check this box to confirm:

## Checklist

### Testing

Relevant predicate resolver, record filtering, row-group filtering, page filtering, and predicate pushdown tests pass.

`git diff --check` also passes.


gunnarmorling:
Hey @MovinduJay, there are lots of unrelated formatting changes, which make it very hard to review this change. Can you revert those and update the PR accordingly? Thx!

MovinduJay:
Thanks! @gunnarmorling  I’ve reverted the unrelated formatting changes and updated the PR with a cleaner, focused diff. The relevant test suites are passing as well.

gunnarmorling:
Hey again, could you please rebase this to current main. Thx!

gunnarmorling:
Rebased to current main

gunnarmorling:
Reworked a few things and applied. Thanks, @MovinduJay!

PR #1033, opened 2026-08-25
#1026 Writer: allow a nullable struct enclosing a repeated field

### LINKED ISSUE #1026: Writer rejects a nullable struct enclosing a repeated field — a shape the reader accepts
Both write APIs refuse a schema in which an `OPTIONAL` struct group directly encloses a repeated field:

```
optional group s {
  optional group phones (LIST) { repeated group list { optional int32 element } }
}
```

```
UnsupportedOperationException: A nullable struct enclosing a repeated field (s) is not yet supported by the writer
```

The guard is in `RecordShredder.impliedRecordCount` (`core/src/main/java/dev/hardwood/internal/writer/RecordShredder.java:308`), with the row layer raising the same at `rowWriter()` time (`RowPlan.java:236`). Walking a leaf's layers from leaf to root, each `REPEATED` layer resets the running record count to `offsets.length - 1`; a nullable `STRUCT` layer above one is refused rather than levelled.

### Why it matters

Only this direction fails. A nullable struct *inside* a repeated scope — `struct("people.list.element", nulls)` — round-trips today.

The refused direction is what PyArrow produces by default, since it marks struct fields nullable: any Arrow or Spark struct column holding a list has this shape. This repository's own fixture `core/src/test/resources/nested_list_struct_test.parquet` is an instance of it:

```
optional group element {
  optional binary name (String);
  optional group sections (List) { repeated group list { ... } }
}
```

Hardwood reads that file in its test suite and cannot write it, so:

- the writer cannot round-trip a shape the reader accepts;
- copying such a file `ColumnReader` → `ColumnBatch` fails on a schema neither side considers unusual;
- the only advice available to a caller is to change what their data means (`REQUIRED` instead of `OPTIONAL`) or to restructure a schema that is usually handed down from upstream.

### Why it looks solvable

The columnar input contract already distinguishes every state the definition levels need for `optional s { optional phones (LIST) { repeated list { optional element } } }`:

| State | Input | Definition level |
|---|---|---|
| `s` absent | `struct("s", validity)` | 0 |
| `s` present, `phones` null | `list("s.phones", offsets, validity)` | 1 |
| `s` present, `phones` empty | zero-delta offsets | 2 |
| element null | element `Validity` | 3 |
| element present | values | 4 |

The guard therefore protects against emitting wrong levels rather than marking something the contract cannot express. Lifting it means teaching the leaf-to-root walk how an absent ancestor maps onto the offsets below it, plus the placeholder handling the shredder already applies for absent struct ancestors elsewhere.

### Acceptance

- A round trip through both write APIs for `optional struct { optional list }`, asserting all five states above read back.
- The same for a nullable struct enclosing a `MAP`, and for one nested inside a list (`chapters.list.element.sections`), matching the fixture.
- `nested_list_struct_test.parquet` copies `ColumnReader` → `ColumnBatch` and reads back equal.
- The interop gate covers the shape, so parquet-java agrees on the levels.

nitrogen404:
fixes #1026 
Both write APIs refused an OPTIONAL struct group that directly encloses a LIST or MAP, even though the columnar input contract already carries enough information to level it correctly and the reader has no trouble with the shape. It is what Arrow-based writers (PyArrow, Spark) produce by default for any struct column holding a list, so the restriction blocked a common, well-formed input and made this repository's own nested_list_struct_test.parquet fixture unwritable even though it reads fine.

RecordShredder.impliedRecordCount's leaf-to-root walk reset the record count at every REPEATED layer and threw the moment a nullable STRUCT layer sat above one, but a STRUCT layer never remaps its item scope, nullable or not, so it already preserved the count unchanged in the non-repeated case - the guard was stricter than the invariant it was protecting. RowPlan's nullableAncestor gate mirrored the same refusal for the row-oriented layer, but nothing else in RowPlan consulted it; RowStructNode/RowRepeatedNode's scope-cascading machinery was already fully generic. Both guards came out with no replacement logic needed.

The still-unsupported shape is unrelated: a legacy two-level list (an unannotated REPEATED field with no LIST/MAP wrapper), refused earlier at FileSchema.Builder / RowPlan construction time, never part of this guard.

## Checklist


gunnarmorling:
Applied; thanks, @nitrogen404!

Issue #1044, opened 2026-08-27
ColumnWorker drain and retriever VThreads: unbounded park() can wedge the pipeline permanently


ZhaoMJ:
_via Claude_

## Failure

Under concurrent read load, one request thread blocks permanently in `BatchExchange.poll()` while all other request threads complete normally. The thread is `TIMED_WAITING`. The stalled thread is waiting for a batch from a specific column's pipeline; that pipeline has stopped making progress. No timeout, no error, no interrupt: the column's `ColumnWorker` is simply parked.

## Root cause

`ColumnWorker` runs two virtual threads — a retriever that fetches pages and submits decode tasks, and a drain that assembles decoded pages into batches. Each waits on a condition the other side makes true and then unparks it:

- The **drain** waits for `reorderBuffer[consumePosition]` to hold a decoded page.
- The **retriever** waits for `consumePosition` to advance past `MAX_INFLIGHT_PAGES`.

Both waits were `LockSupport.park()` — unbounded, with no timeout.

The conditions are monotone: once true they remain true until the waiter itself acts. But `park()` can return prematurely (a known property of `LockSupport`), and it can also *fail to park* when an unpark arrives before the thread enters `park()`. When that race occurs the unpark is consumed without the thread noticing, and no second unpark will ever arrive because nothing else is left to send one. The pipeline wedges permanently.

## Diagnostic evidence

Observed at concurrency 64. Thread state of the stalled column worker:

```
col_X   ready=0  consumePos=100  submitted=108  drained=100  inFlight=0
        done=false  error=null  finished=false  slots=PPPPPPPP
```

Interpretation:
- Every reorder-buffer slot is full of decoded pages (`slots=PPPPPPPP`), including slot `100 % 8 = 4`, which is exactly the one the drain is parked waiting for.
- `inFlight=0`: every decode task had already completed and called `unpark`.
- The retriever is throttled at `submitted − consumePosition = 108 − 100 = 8 = MAX_INFLIGHT_PAGES` — it cannot advance until the drain does.
- `done=false`, `error=null`, `finished=false`: no shutdown in progress, no error path reached.

This state is unreachable if any of the three `unpark` calls had taken effect. It is unrecoverable once none of them did.

## Reproduction

The hang reproduced on **6 of 10** sweep points at concurrency 64. After the fix described below, **0 of 28** sweep points exhibited a hang under similar load.

## Fix

[...]

PR #1046, opened 2026-08-27
#1044 reader: bound the drain and retriever parks to survive the JDK 25 lost-unpark regression

ZhaoMJ:
_Prepared via Claude_

## Summary

Closes issue #1044.

Both virtual threads in `ColumnWorker` used `LockSupport.park()` for conditions the other side makes true and then unparks them for. A lost unpark — an unpark that arrives before the thread enters `park()` — wedged the column's pipeline permanently, because no second notification would ever arrive. The consumer then waited forever in `BatchExchange.poll()`, presenting as one request thread stuck in `TIMED_WAITING` while all others complete.

This PR replaces both unbounded parks with `parkNanos(10 ms)`.

## Changes

Single file: `core/src/main/java/dev/hardwood/internal/reader/ColumnWorker.java`.

Three call sites replaced:

| Site | Before | After |
|------|--------|-------|
| Retriever throttle (line ~298) | `LockSupport.park()` | `LockSupport.parkNanos(WAKE_CHECK_NANOS)` |
| Retriever throttle in sentinel path (line ~324) | `LockSupport.park()` | `LockSupport.parkNanos(WAKE_CHECK_NANOS)` |
| Drain waiting for decode (line ~383) | `LockSupport.park()` | `LockSupport.parkNanos(WAKE_CHECK_NANOS)` |

`WAKE_CHECK_NANOS = 10L * 1_000_000L` is a new package-private constant with a Javadoc block explaining the liveness argument. The `unpark` calls at every notification point are kept unchanged.

## Why it is safe

Both wait conditions are monotone: once true they remain true until the waiter itself acts. Re-checking after a spurious or timeout wake always produces a correct decision. The 10 ms bound is only a fallback; the common path — condition already satisfied — exits on the first `unpark` as before. A spurious wake costs one integer comparison or one `AtomicReferenceArray` read.

## Consistency with BatchExchange

`BatchExchange` already uses `parkNanos` for both of its queue waits, with the same justification. Before this change, the two parks in `ColumnWorker` were the only unbounded waits in the pipeline.

## Test evidence

- Hang reproduced on **6 of 10** sweep points at concurrency 64 before the fix.
- **0 of 28** sweep points exhibited a hang after the fix under similar load.
- Hardwood core test suite on this branch: **2,668 run, 0 failures, 1 skipped** (`./mvnw -pl core -am test`).

### Deterministic regression test

`ColumnWorkerTest.drainWakesOnTimeoutWhenUnparkWasMissed` demonstrates the exact failure class without requiring high concurrency.

The test uses a stalled executor — one that captures decode `Runnable`s submitted by the retriever but never executes them. This means `reorderBuffer[slot]` is never written by `decode()` and `LockSupport.unpark(drainThread)` is never called from it. Once the drain is parked waiting on slot 0, the test injects the `EMPTY_SENTINEL` directly into slot 0 via reflection, with no corresponding `unpark`. This reproduces the invariant violation from the production hang: the slot is full but the drain has received no notification and has no outstanding permit.

With `parkNanos(WAKE_CHECK_NANOS)` (the fix) the drain re-checks the buffer at the next timeout wake (~10 ms), reads the sentinel, calls `finishDrain()`, and sets `done = true`. The test polls `worker.done` for up to 200 ms and asserts it became true — this passes within ~20 ms.

With the unfixed `LockSupport.park()` the drain stays parked indefinitely (the stalled executor never calls `unpark`). The assertion fires after the 200 ms polling window, failing the test within ~250 ms — well before the 5-second `@Timeout` ceiling. The failure was verified by temporarily reverting `parkNanos` to `park()` and re-running the test.

## Authorship

Check this box to confirm:

## Checklist

### DIFF
--- a/core/src/main/java/dev/hardwood/internal/reader/ColumnWorker.java
+++ b/core/src/main/java/dev/hardwood/internal/reader/ColumnWorker.java
@@ -217,7 +217,7 @@ public abstract class ColumnWorker<B> implements AutoCloseable {
         // is not enough on its own: ArrayBlockingQueue's timed operations go through
         // AQS.ConditionObject.awaitNanos, which treats a bare unpark as spurious and re-parks
         // for the remainder of the window. The unpark above is still needed for the drain's
-        // other wait, the LockSupport.park() in runDrain that waits on a decode task.
+        // other wait, the LockSupport.parkNanos in runDrain that waits on a decode task.
         //
         // Only the drain is interrupted, and only because it does no I/O: every InputFile
         // access happens on the retriever (via PageSource.next) or on a decode task. That
@@ -263,7 +263,7 @@ public abstract class ColumnWorker<B> implements AutoCloseable {
     private long sourceNanos;
     private long throttleNanos;
     private int totalPagesSubmitted;
-    private int throttleParks;
+    private int throttleWakes;
 
     private void runRetriever() {
         try {
@@ -297,7 +297,7 @@ public abstract class ColumnWorker<B> implements AutoCloseable {
                 // Throttle: park while too many pages are in flight
                 t0 = System.nanoTime();
                 while (!done && nextSeq - consumePosition >= MAX_INFLIGHT_PAGES) {
-                    throttleParks++;
+                    throttleWakes++;
                     LockSupport.parkNanos(WAKE_CHECK_NANOS);
                 }
                 throttleNanos += System.nanoTime() - t0;
@@ -335,9 +335,9 @@ public abstract class ColumnWorker<B> implements AutoCloseable {
 
             LOG.log(System.Logger.Level.DEBUG,
                     "[{0}] Retriever finished: {1} pages submitted. "
-                    + "source={2,number,0.0}ms, throttle={3,number,0.0}ms ({4} parks)",
+                    + "source={2,number,0.0}ms, throttle={3,number,0.0}ms ({4} wakes)",
                     column.name(), totalPagesSubmitted,
-                    sourceNanos / 1_000_000.0, throttleNanos / 1_000_000.0, throttleParks);
+                    sourceNanos / 1_000_000.0, throttleNanos / 1_000_000.0, throttleWakes);
         }
         catch (Throwable t) {
             signalError(enrichWithFileName(t, pageSource.getCurrentFileName()));
@@ -366,7 +366,7 @@ public abstract class ColumnWorker<B> implements AutoCloseable {
     private long assemblyNanos;
     private long decodeWaitNanos;
     private int totalPagesDrained;
-    private int decodeWaitParks;
+    private int decodeWaitWakes;
 
     private void runDrain() {
         try {
@@ -382,7 +382,7 @@ public abstract class ColumnWorker<B> implements AutoCloseable {
                     // No pages were ready — wait for a decode task to complete, but re-check rather than
                     // rely on being told. See WAKE_CHECK_NANOS.
                     long parkStart = System.nanoTime();
-                    decodeWaitParks++;
+                    decodeWaitWakes++;
                     LockSupport.parkNanos(WAKE_CHECK_NANOS);
                     decodeWaitNanos += System.nanoTime() - parkStart;
                 }
@@ -394,10 +394,10 @@ public abstract class ColumnWorker<B> implements AutoCloseable {
 
             LOG.log(System.Logger.Level.DEBUG,
                     "[{0}] Drain finished: {1} pages drained, {2} batches. "
-                    + "assembly={3,number,0.0}ms, decodeWait={4,number,0.0}ms ({5} parks), "
+                    + "assembly={3,number,0.0}ms, decodeWait={4,number,0.0}ms ({5} wakes), "
                     + "publishBlock={6,number,0.0}ms",
                     column.name(), totalPagesDrained, batchesPublished,
-                    pureAssembly / 1_000_000.0, decodeWaitNanos / 1_000_000.0, decodeWaitParks,
+                    pureAssembly / 1_000_000.0, decodeWaitNanos / 1_000_000.0, decodeWaitWakes,
                     publishBlockNanos / 1_000_000.0);
         }
         catch (Throwable t) {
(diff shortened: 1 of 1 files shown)


gunnarmorling:
Hey @ZhaoMJ, any chance you provide a reproducer for this? I tried to recreate this situation but have failed to do so so far.

ZhaoMJ:
@gunnarmorling I've taken a closer look. The PR works but the root-cause analysis is incorrect. It appears to be a known JDK bug in the version I was using (25.0.0), and this patch only happens to work around it. The regression test as written is also incorrect and circular, so I’ll remove it.

## The explanation in #1044 is wrong

I claimed an `unpark` arriving before the thread enters `park()` is lost. However, that is not how `LockSupport` works: the permit makes exactly that case safe. According to the spec:

> `unpark`: "Makes available the permit for the given thread... If the thread was blocked on `park` then it will unblock. **Otherwise, its next call to `park` is guaranteed not to block.**"
>
> class doc: "Races between one thread invoking `park` and another thread trying to `unpark` it will preserve liveness, due to the permit."

## What actually happened: JDK-8369227, a JDK 25 regression

[JDK-8369227](https://bugs.openjdk.org/browse/JDK-8369227) — "Virtual thread stuck in PARKED state". A regression from JDK-8351927 ("Change VirtualThread implementation to use FJP delayed task handling") in **JDK 25**, fixed in **25.0.3, 26.0.1 and 27**. Alan Bateman's summary: *"a virtual thread doing untimed-park immediately after a timed-park when racing with an async unpark and the timeout task ... executing before it is cancelled."*

In `VirtualThread` as shipped in 25 GA:

1. `parkNanos(T)` → `afterYield()` arms `timeoutTask = schedule(this::parkTimeoutExpired, T)`; state `TIMED_PARKED`.
2. An async `unpark()` lands just before `T` expires → permit `true`, `TIMED_PARKED → UNPARKED`, continuation submitted.
3. `runContinuation()` → `cancelTimeoutTask()` → `timeoutTask.cancel(false)`, whose own comment concedes *"The timeout task may be executing, or may have already completed."* If `parkTimeoutExpired()` is already past the cancel check it survives.
4. The thread resumes, and then does an **untimed** `park()` → state `PARKED`.
5. The stale task finally runs:
   ```java
   if (!getAndSetParkPermit(true) && (state() == TIMED_PARKED) && compareAndSetState(TIMED_PARKED, UNPARKED))
       lazySubmitRunContinuation();
   ```
   It **sets the permit to true**, but the state is `PARKED`, not `TIMED_PARKED` — so it does not resubmit the continuation.
6. The real `unpark()` arrives, finds `getAndSetParkPermit(true) == true`, and returns without submitting anything. The virtual thread is `PARKED` forever and **every** later unpark is swallowed.

This has been fixed by making `parkTimeoutExpired()` go through the full `unpark()` path so the `PARKED` case is handled too.

## Why `ColumnWorker`'s drain is precisely this shape

The drain alternates a timed park with an untimed one, and is unparked asynchronously from decode tasks on other threads:

- timed park: `publishCurrentBatch()` → `BatchExchange.publish()` → `readyQueue.offer(batch, 10, MILLISECONDS)`, and `takeBatch()` → `freeQueue.poll(10, MILLISECONDS)`. Both park only under back-pressure — which is the state the stalled reader was in (35 of 36 workers had a full ready queue).
- untimed park: `runDrain()`'s `LockSupport.park()` waiting for `reorderBuffer[consumePosition]`, unparked by `decode()` from the decode executor.

So the 10 ms timeouts inside `BatchExchange` are firing continuously, an async unpark from a decode task can land microseconds either side of one of them, and the very next thing the drain does is an untimed park. That is the trigger, per iteration, on every column.

This also resolves what looked like a contradiction in the #1044 dump. `slots=PPPPPPPP` with `inFlight=0`, `done=false`, `error=null` and the retriever throttled at exactly `MAX_INFLIGHT_PAGES` is unreachable if the notification had been delivered — and it wasn't: the JDK dropped it and neutered every unpark after it.

## Independent reproducer produced by Claude

Standalone, no hardwood, no JDK test library: [LostUnpark3.java](https://github.com/user-attachments/files/31655989/LostUnpark3.java). It aims an async unpark a few hundred microseconds *before* a timed park's expiry, so the timeout task is already dispatching when `cancel(false)` arrives, keeps the carriers busy so the stale task executes late, then does the untimed park and delivers the notification that must wake it.

```
java LostUnpark3.java <seconds> <cells> <burners> <victim|control|fix>
```

On an M1 Pro (10 carriers), `java LostUnpark3.java 60 8 2 victim`:

| Arm | JVM | Result |
|---|---|---|
| A `victim` — timed park, then untimed park | Corretto 25+36 (25.0.0) | **hang** at attempt #7,097 — `proceed=true` and `unpark()` called, thread still `WAITING` at `VirtualThread.park:738` |
| B `control` — untimed park, no preceding timed park | Corretto 25+36 | no hang in 28,017 attempts |
| C `fix` — timed park, then `parkNanos(10 ms)` (this PR) | Corretto 25+36 | no hang |
| D `victim` — identical to A | Corretto 25.0.4.1 | no hang in 23,910 attempts |

A vs. B isolates the preceding timed park as the trigger. A vs. D confirms it is the runtime, not the test. Arm C matters less as an experiment than as a code-path argument: a stale `parkTimeoutExpired()` firing against `TIMED_PARKED` satisfies its own guard and *does* resubmit, and a timed waiter is bounded by its own fresh timeout in any case. Only untimed parks get stranded, which is why bounding these two waits sidesteps the bug.

## What I think should happen to this PR

I’ll push a commit to update the comments and drop the test. However, since this is a JDK bug, any untimed park following a timed park on a virtual thread is affected, not just these three sites. Therefore, I think it’s totally reasonable to not carry a workaround for a bug fixed in the JDK, and close the PR and instead document a minimum runtime requirement of 25.0.3 for JDK 25.

gunnarmorling:
Hey @ZhaoMJ, thanks for the analysis! I think there's value in having this change, despite this being fixed in the JDK now. It will help folks currently running on the affected versions, and there's no harm in us doing so.

PR #1050, opened 2026-08-27
#1049 Assert Parquet column orders through interop

### LINKED ISSUE #1049: Nothing asserts the footer's column_orders, which is what makes min_value/max_value trustworthy
`ParquetFileWriter.columnOrders()` returns one `TYPE_DEFINED_ORDER` per leaf column and `FileMetaDataWriter` writes the list when it is non-empty, so produced files carry it. Nothing checks that they do.

It is worth pinning because of what depends on it rather than because it looks fragile. A reader only trusts `min_value` / `max_value` under the order the footer declares; with `column_orders` absent, parquet-java falls back to the deprecated `min` / `max` semantics, under which a `BYTE_ARRAY` bound is compared **signed**. Hardwood writes `STRING` bounds unsigned and `DECIMAL` bounds signed, both correct under `TYPE_DEFINED_ORDER` — so a file that lost the list would have its string statistics silently misread by every other implementation, and every existing statistics test would still pass, since they all read the file back through Hardwood's own reader, which does not need the list to know its own conventions.

That is the shape of bug the interop gate exists to catch and currently cannot.

## What to do

Two assertions, either in `WriterFooterMetadataTest` (that the writer emits one `TYPE_DEFINED_ORDER` per leaf column, for a schema of several columns) or better in `WriterFooterMetadataInteropTest`, where parquet-java's `ParquetMetadata` exposes the decoded orders — asserting there proves the bytes are what another implementation reads, not merely what we wrote.

The stronger version, if it is cheap: a `STRING` column whose values straddle the signed/unsigned boundary — one value with a high bit set, one without — written by Hardwood and read back through parquet-java, whose `min`/`max` then differ depending on whether the order was declared. That fails loudly if the list is ever dropped, rather than only asserting a field is present.

Fhatu12:
## Purpose

Closes #1049.

Hardwood already writes `column_orders`, but there was no independent interop regression proving the raw footer field survives the wire format and is seen as `TYPE_DEFINED_ORDER` by parquet-java.

## Implementation

- add raw Parquet footer reading support to the test helper using parquet-java/parquet-format
- assert raw `column_orders` cardinality for a multi-leaf schema including nested leaves
- assert each raw entry is `TYPE_DEFINED_ORDER`
- additionally assert parquet-java's high-level schema sees `TYPE_DEFINED_ORDER`
- add the complementary STRING statistics regression across the signed/unsigned byte boundary

The STRING bounds test confirms the expected unsigned STRING statistics behaviour, but it is not the independent proof of raw `column_orders` presence. The raw footer assertion provides that proof.

## Testing

- Focused `WriterFooterMetadataInteropTest`: 6 passed, 0 failures/errors/skips
- Adjacent `WriterFooterMetadataInteropTest` + `WriterLogicalTypeInteropTest`: 33 passed, 0 failures/errors/skips
- Relevant reactor:
  - error-prone: 10 passed
  - core: 2683 passed / 1 skipped
  - parquet-testing-runner: 935 passed / 68 skipped
  - write coverage: 5 passed
- Full `./mvnw clean verify`: `BUILD SUCCESS`
- Independent mutation QA: temporarily omitting writer `column_orders` caused the raw regression to fail as intended
- `git diff --check HEAD^ HEAD` passed

### Process-sources note

`./mvnw process-sources` was attempted but currently fails on both this candidate and pristine upstream parent because the `process-sources` lifecycle reaches `hardwood-core` without resolving `dev.hardwood:hardwood-test-support:jar:1.1.0-SNAPSHOT`.

This is classified as `PRE_EXISTING_REPOSITORY_LIFECYCLE_BEHAVIOUR`, not a candidate-caused regression. Full `./mvnw clean verify` succeeds.

## AI assistance

OpenAI Codex assisted with repository analysis, test implementation, validation, and publication preparation. The change was independently QA'd in a separate pass. ChatGPT assisted with planning, review coordination, and publication preparation.

The human contributor reviewed the purpose and evidence and is responsible for the submission and maintainer communication.

## Authorship

Check this box to confirm:

## Checklist

### DIFF
--- a/_designs/WRITER_INTEROP_GATE.md
+++ b/_designs/WRITER_INTEROP_GATE.md
@@ -76,10 +76,16 @@ Per file, four things:
    empty-versus-absent repeated values are compared as written.
 3. **The metadata agrees.** parquet-java parses the footer, and the column-chunk
    statistics it exposes — `min`, `max`, `null_count` — match the written data under
-   *its* comparator, which it derives from the annotation and the footer's
-   `column_orders`. Agreement here is the first cross-implementation check on stage 13b:
-   an order the two implementations disagree about produces bounds that would prune a row
-   group holding a matching row.
+   *its* comparator, which it derives from the column's annotation. Agreement here is the
+   first cross-implementation check on stage 13b: an order the two implementations
+   disagree about produces bounds that would prune a row group holding a matching row.
+
+   The `column_orders` those bounds take their meaning from is asserted separately, over
+   the **raw** footer. parquet-java defaults a missing column order to type-defined for
+   every type but `INT96` and `INTERVAL`, so its decoded schema reports
+   `TYPE_DEFINED_ORDER` whether the field reached the wire or not, and every other
+   assertion here would hold on a file that dropped the list — leaving bounds the format
+   calls undefined for a reader that holds it to its word.
 4. **The case covered what it claims to.** The encodings say the case actually produced
    the dictionary, `PLAIN` fallback or plain-only chunk it exists to cover, and the layout
    cases walk parquet-java's page readers to count the data pages they crossed. Without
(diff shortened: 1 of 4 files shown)


gunnarmorling:
Slightly reworked and merged. Thanks, @Fhatu12!

PR #1051, opened 2026-08-27
#1048 Sweep nullability in the optional-encoding interop axis

### LINKED ISSUE #1048: Interop gate covers the named encodings at one nullability out of four
`WriterInteropTest` has two axes covering the four encodings stage 19b added. Both derive their
`(encoding, physical type)` pairs from `EncodingSupport.supports` rather than from a hand-written
copy, so every legal pair is written and read back through parquet-java with values compared row by
row. That part is in good shape — a pair added to or dropped from the table moves the gate in the
same commit.

Both axes then pin `Nullability.OPTIONAL_SOME_NULL`, so each pair has only ever been shown to
another implementation in one of the four repetition shapes a flat column can take.

## What is covered

`Nullability` has four members. `optionalEncodings` sweeps the ten legal pairs against one of them:

| Encoding | Physical type | `REQUIRED` | `OPTIONAL` all-present | `OPTIONAL` some null | `OPTIONAL` all-null |
|---|---|---|---|---|---|
| `DELTA_BINARY_PACKED` | `INT32` | ✗ | ✗ | ✅ | ✗ |
| `DELTA_BINARY_PACKED` | `INT64` | ✗ | ✗ | ✅ | ✗ |
| `DELTA_LENGTH_BYTE_ARRAY` | `BYTE_ARRAY` | ✗ | ✗ | ✅ | ✗ |
| `DELTA_BYTE_ARRAY` | `BYTE_ARRAY` | ✗ | ✗ | ✅ | ✗ |
| `DELTA_BYTE_ARRAY` | `FIXED_LEN_BYTE_ARRAY` | ✗ | ✗ | ✅ | ✗ |
| `BYTE_STREAM_SPLIT` | `INT32` | ✗ | ✗ | ✅ | ✗ |
| `BYTE_STREAM_SPLIT` | `INT64` | ✗ | ✗ | ✅ | ✗ |
| `BYTE_STREAM_SPLIT` | `FLOAT` | ✗ | ✗ | ✅ | ✗ |
| `BYTE_STREAM_SPLIT` | `DOUBLE` | ✗ | ✗ | ✅ | ✗ |
| `BYTE_STREAM_SPLIT` | `FIXED_LEN_BYTE_ARRAY` | ✗ | ✗ | ✅ | ✗ |

10 of 40 cells.

`optionalEncodingCodecs` takes one representative type per encoding and crosses it with the codecs
parquet-java can read, again at `OPTIONAL_SOME_NULL` only:

| Encoding / type | `UNCOMPRESSED` | `GZIP` | `SNAPPY` | `ZSTD` | `LZ4_RAW` |
|---|---|---|---|---|---|
| `DELTA_BINARY_PACKED` / `INT64` | some null | some null | some null | some null | some null |
| `DELTA_LENGTH_BYTE_ARRAY` / `BYTE_ARRAY` | some null | some null | some null | some null | some null |
| `DELTA_BYTE_ARRAY` / `BYTE_ARRAY` | some null | some null | some null | some null | some null |
| `BYTE_STREAM_SPLIT` / `DOUBLE` | some null | some null | some null | some null | some null |

20 of 80 cells.

## Why the missing shapes are the interesting ones

- `REQUIRED` — no definition-level stream at all, so the encoder is fed the values through a
  different path than every covered case.
- `OPTIONAL_ALL_NULL` — a page with **zero** values to encode, and, on the codec axis, a
  near-empty page body handed to a compressor.
- `OPTIONAL_ALL_PRESENT` — a levelled column whose definition-level stream is a single run.

I probed the all-null end by hand against all four encodings and they write without complaint, so
this is a coverage gap rather than a known defect, and closing it is a matter of widening axes that
already exist.

## What to do

- [ ] Sweep `Nullability.values()` in `optionalEncodings` the way `singleEntryDictionary` already
      does, instead of pinning `OPTIONAL_SOME_NULL`.
- [ ] Do the same for `optionalEncodingCodecs`.

The assertions, the fixtures and the legality filter are all in place, so this is only the axis
definitions. Check the runtime after the second one: it multiplies the case count four-fold.

nitrogen404:
closes #1048
`WriterInteropTest.optionalEncodings` pinned every case at `OPTIONAL_SOME_NULL` so `DELTA_BINARY_PACKED`, `DELTA_LENGTH_BYTE_ARRAY`, `DELTA_BYTE_ARRAY` and `BYTE_STREAM_SPLIT` never went through the parquet-java gate as REQUIRED, all-present, or all-null columns. The last of those hands an encoder an empty range to encode and `REQUIRED` takes a different level path entirely so the pinned axis was covering only one of the four repetition shapes a legal pair can take. singleEntryDictionary already sweeps Nullability.values() the same way; this widens optionalEncodings to match using the InteropCase machinery that shape already exercises elsewhere.

## Checklist

### DIFF
--- a/_designs/WRITER_INTEROP_GATE.md
+++ b/_designs/WRITER_INTEROP_GATE.md
@@ -99,17 +99,22 @@ repetition shape.
 
 Beyond the floor, the matrix varies one axis at a time against a representative base
 rather than taking the full cross product, which would multiply to hundreds of files for
-coverage the axes already give independently. Every axis is swept across all seven
-writable physical types (`BOOLEAN`, `INT32`, `INT64`, `FLOAT`, `DOUBLE`, `BYTE_ARRAY`,
+coverage the axes already give independently. An axis is swept across all seven writable
+physical types (`BOOLEAN`, `INT32`, `INT64`, `FLOAT`, `DOUBLE`, `BYTE_ARRAY`,
 `FIXED_LEN_BYTE_ARRAY`), since the value encoders are per-type and that is where an
-encoding defect lives.
+encoding defect lives — except where the axis is defined over only some of them, as an
+encoding legal for two types is, and there the sweep is restricted to those.
 
 | Axis | Values |
 |------|--------|
 | Repetition | `REQUIRED`; `OPTIONAL` all-present; `OPTIONAL` with interleaved nulls; `OPTIONAL` all-null |
-| Encoding | single-entry dictionary; multi-entry dictionary; dictionary overflowing to mid-chunk `PLAIN`; dictionary disabled |
-| Codec | `UNCOMPRESSED`; `ZSTD` |
+| Encoding | single-entry dictionary; multi-entry dictionary; an all-distinct column the size comparison writes `PLAIN`; dictionary disabled |
+| Optional encoding | `DELTA_BINARY_PACKED`; `DELTA_LENGTH_BYTE_ARRAY`; `DELTA_BYTE_ARRAY`; `BYTE_STREAM_SPLIT` — one case per legal (encoding, physical type) pair, in every repetition shape |
+| Optional encoding × codec | each of those four encodings, on one type it is legal for, against every codec, in every repetition shape |
+| `FIXED_LEN_BYTE_ARRAY` length | 2, 8, 12 and 16 bytes against every encoding legal for the type |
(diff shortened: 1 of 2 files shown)


gunnarmorling:
LGTM, merging. Thanks, @nitrogen404!

PR #1057, opened 2026-08-28
#1019 Decode DELTA_BINARY_PACKED a miniblock at a time

### LINKED ISSUE #1019: Bulk-unpack DELTA_BINARY_PACKED miniblocks instead of decoding value by value
## Context

`DeltaBinaryPackedDecoder` reconstructs one value per call, and each call re-derives everything around it: a modulo against the block size to detect a block boundary, a miniblock-boundary check, and a bit-at-a-time unpack loop that recomputes byte and bit offsets on every iteration. `loadMiniblock` allocates a fresh `byte[]` and `System.arraycopy`s the body into it — once per 32 values.

The neighbouring `RleBitPackingHybridDecoder` already does this the other way: bulk reads, eight values per iteration, width-specialised paths. The delta decoder never got that treatment.

This matters beyond integer columns: `DELTA_LENGTH_BYTE_ARRAY` and `DELTA_BYTE_ARRAY` both run their length arrays through this decoder during `initialize()`, and `DELTA_BYTE_ARRAY` does it twice — once for prefix lengths, once for the suffix lengths of the nested `DELTA_LENGTH_BYTE_ARRAY`. Both call sites already have an `int[]` to fill and loop `readInt()` per value instead of using the bulk entry point.

## Proposal

Decode a whole miniblock at a time into a buffer and serve values from it:

- unpack the miniblock in bulk, with width-specialised readers (eight values of `bitWidth` bits occupy exactly `bitWidth` bytes, so widths up to 8 are one 8-byte load per group; widths 9–32 come from a 64-bit accumulator refilled four bytes at a time; wider widths stay bit-at-a-time)
- drop the per-miniblock allocation and copy — unpack in place from the page bytes
- drain in bulk from `readInts` / `readLongs` rather than one `readLongValue()` call per value
- point `DeltaLengthByteArrayDecoder` and `DeltaByteArrayDecoder` at the bulk `readInts`

## Measured

Decoder throughput, 4096 values per op, by the bits each delta needs (ops/ms):

| deltaBits | before | after |
|---|---|---|
| 1 | 83.3 | 300.4 (3.61x) |
| 5 | 63.2 | 316.3 (5.00x) |
| 12 | 49.8 | 179.8 (3.61x) |
| 21 | 27.1 | 143.8 (5.31x) |

End-to-end scan of a 2M-row file written at writer version 2.0 (ms/op, lower is better):

| column | before | after |
|---|---|---|
| INT32, DELTA_BINARY_PACKED | 5.343 | 2.602 (2.05x) |
| INT64, DELTA_BINARY_PACKED | 4.442 | 2.896 (1.53x) |
| BYTE_ARRAY, DELTA_BYTE_ARRAY | 40.852 | 39.916 (no change) |

## What not to try: shortening the prefix-sum dependency chain

Delta reconstruction is a prefix sum, which is serial by nature, and the obvious next move is the one FastLanes makes: lift `min_delta` out of the chain as a closed form (`(i + 1) * min_delta`) and split the miniblock into independent chains that are carried across afterwards.

This was implemented and measured. It is **9–22% slower at every bit width** and was reverted.

A 32-value miniblock is too short to be latency-bound: four chains save roughly 24 cycles of dependency latency, but the carry pass costs 32 more loads, stores and multiplies, which a 4-wide out-of-order core does not hide. The technique needs FastLanes' 1024-element chains and lanes that make the fixup free. Please do not re-propose it for Parquet's miniblock size without new evidence.

## Also in scope

The rewrite sizes its buffers from the header, so the header needs bounds it did not previously need: a block size that is positive, within a sane maximum, and divisible by the miniblock count. Separately, miniblock bit widths are read as an unsigned byte and were never validated — a corrupt width of 200 reaches `1L << 200`, which Java evaluates as `1L << 8`, yielding silently wrong values instead of an error.

## Coverage

`DeltaBinaryPackedDecoderTest` sweeps every bit width from 0 to 64 for both INT32 and INT64 against parquet-java as the encoding oracle, at a value count that leaves the last miniblock partially filled. The decoder previously had only fixture-based coverage over small value ranges, so widths above roughly 8 bits were never exercised — and the rewrite adds two path boundaries (8/9 and 32/33) that would otherwise be hit only incidentally.

ZhaoMJ:
_Prepared via Claude._

### What changed

`DeltaBinaryPackedDecoder` now decodes a whole miniblock at a time into a
pre-allocated `long[]` buffer, with three width-specialised unpack paths:

- **Width 0**: all residuals are zero; no bytes consumed, the loop writes
  `minDelta` directly.
- **Widths 1–8**: every 8 values occupy exactly `bitWidth` bytes. Those bytes
  are loaded into a 64-bit word (little-endian) and 8 values are extracted via
  shift-and-mask, one multi-byte load per group of 8.
- **Widths 9–32**: a 64-bit accumulator is refilled one byte at a time; max
  bits in the accumulator at any point is `bitWidth − 1 + 8 ≤ 39`, safely
  within a `long`.
- **Widths 33–64**: bit-at-a-time (the previous slow path, retained because
  widths 57+ would overflow a 64-bit accumulator on the 9–32 path).

The per-miniblock `byte[]` allocation is eliminated; `miniblockBuffer` is
allocated once in `readHeader()` and reused for every miniblock.

`readInts` and `readLongs` drain from the buffer in chunks via
`System.arraycopy` / a bulk int-cast loop rather than calling `readLongValue()`
per value.

`DeltaLengthByteArrayDecoder` and `DeltaByteArrayDecoder` are updated to call
`readInts(lengths, null, 0)` / `readInts(prefixLengths, null, 0)` in their
`initialize()` methods, replacing the per-element `readInt()` loop.

Header validation is added: `blockSize` must be positive and divisible by
`miniblockCount`; each bit width is validated as an unsigned byte in `[0, 64]`.

### New test

`DeltaBinaryPackedDecoderWidthSweepTest` is a parameterised JUnit 5 test that
sweeps every bit width 0–32 for INT32 and 0–64 for INT64. Each case encodes 50
values, which gives a full miniblock of 32 plus a partial last miniblock of 17,
exercising the partial-last-miniblock truncation. The width-64 and width-32
cases use alternating `Long.MIN_VALUE`/`Long.MAX_VALUE` and
`Integer.MIN_VALUE`/`Integer.MAX_VALUE` to cover wrap-around. All other cases
construct values so that the largest residual requires exactly `w` bits.

### Benchmark results

From the benchmark suite referenced in issue #1019
(4096 values/op; ops/ms, higher is better):

| deltaBits | Before | After  | Speedup |
|-----------|--------|--------|---------|
| 1         | 83.3   | 300.4  | 3.6×    |
| 5         | 63.2   | 316.3  | 5.0×    |
| 12        | 49.8   | 179.8  | 3.6×    |
| 21        | 27.1   | 143.8  | 5.3×    |

End-to-end column scan (2 M-row synthetic file, ms/op — lower is better):

| Column type                    | Before | After | Speedup |
|-------------------------------|--------|-------|---------|
| INT32, DELTA_BINARY_PACKED     | 5.343  | 2.602 | ~2.1×   |
| INT64, DELTA_BINARY_PACKED     | 4.442  | 2.896 | ~1.5×   |
| BYTE_ARRAY, DELTA_BYTE_ARRAY   | 40.852 | 39.916 | no change |

Gains are largest at narrow bit widths, where the 1–8 path processes 8 values
per `bitWidth`-byte load; the BYTE_ARRAY scan is unchanged because string
copying dominates, not the integer length decode.

### Why this is scalar, and not SIMD

`DeltaBinaryPackedDecoder` has never been on a vectorised path, on any JVM: it
does not reference `SimdOperations`, and `RleBitPackingHybridDecoder` — the only
class in `core/src/main/java` that does — calls it for `countNonNulls` and the
four `applyDictionary*` methods, never for bit unpacking. So this change speeds
up every JVM equally, whether or not `jdk.incubator.vector` resolves; there is
no faster path it is competing with.

`SimdOperations` does declare `unpackBitWidth1` and `unpackBitWidthN`, which
look like the primitives a miniblock body wants, and using them was tried. Two
things rule it out. They have no production caller today, and
`VectorOperations.unpackBitWidthN` is byte-for-byte identical to
`ScalarOperations.unpackBitWidthN` — the same little-endian word load and
unrolled shift-and-mask this patch's 1–8 path uses, with a comment saying scalar
unrolling was judged competitive for eight values. There is no vector code in
either. Their contract is also narrower than the delta path needs: they write
`int[]` and load `bitWidth` bytes into a `long`, so they are correct only up to
width 8, while a miniblock runs to 64.

Routing the delta path through them therefore costs rather than gains, because
it splits one fused pass into an unpack into `int[]` plus a separate prefix-sum
pass. Measured on a 4096-value sweep (ops/ms, higher is better):

| width | fused (this patch) | via `unpackBitWidthN` + prefix sum | real `LongVector` unpack + prefix sum |
|-------|-------|-------|-------|
| 1     | 462.1 | 310.3 | 335.5 |
| 4     | 317.1 | 255.8 | 273.3 |
| 8     | 217.8 | 180.8 |  58.1 |

The fused kernel wins at every width. The underlying reason is that the prefix
sum is serial and cannot be vectorised — isolating the two halves puts the
unpack at 247–581 ops/ms and the prefix sum alone at 509–574, so even a free
unpack would leave the prefix sum as the floor. Any formulation that vectorises
the unpack has to materialise residuals to memory for the serial sum to read
back, and that round-trip costs more than the unpack saves. This is the same
result as the FastLanes experiment below, reached from the other direction: at a
32-value miniblock there is not enough width to pay for breaking the dependency
chain. (Measurements are on 128-bit NEON, where `LongVector` has two lanes;
wider hardware would narrow the last column's gap but not change the floor.)

### FastLanes chain-shortening: implemented, measured, rejected

The FastLanes prefix-sum chain-shortening approach (interleaving 32 independent
sub-sums and merging at the end to break the serial dependency chain) was
implemented and benchmarked on 32-value miniblocks. It is **9–22% slower** at
every bit width at this miniblock size. The overhead of the merge pass exceeds
any out-of-order execution benefit at 32 values. Hardwood's miniblock size is
32; the FastLanes paper targets 1024-value lanes. 

### Notes

The partial-last-miniblock truncation in `loadMiniblock` limits `validValues`
to `totalValueCount − valuesRead`. The Parquet spec requires encoders to pad
the last miniblock with zero-delta values; this bounds the caller to the
declared count rather than the full miniblock size.

Closes #1019

## Authorship

Check this box to confirm:

## Checklist

### DIFF
--- a/core/src/main/java/dev/hardwood/internal/encoding/DeltaBinaryPackedDecoder.java
+++ b/core/src/main/java/dev/hardwood/internal/encoding/DeltaBinaryPackedDecoder.java
@@ -45,11 +45,11 @@ public class DeltaBinaryPackedDecoder implements ValueDecoder {
     private long minDelta;
     private int[] bitWidths;
     private int currentMiniblock;
-    private int valuesInCurrentMiniblock;
 
-    // Bit unpacking buffer
-    private byte[] miniblockData;
-    private int bitPosition;
+    // Pre-allocated miniblock decode buffer (size = valuesPerMiniblock)
+    private long[] miniblockBuffer;
+    private int bufferPos;
+    private int bufferFill;
 
     public DeltaBinaryPackedDecoder(byte[] data, int offset) {
         this.data = data;
@@ -78,9 +78,31 @@ public class DeltaBinaryPackedDecoder implements ValueDecoder {
     /// Read INT64 values directly into a primitive long array.
     @Override
     public void readLongs(long[] output, int[] definitionLevels, int maxDefLevel) throws IOException {
+        if (!headerRead) {
+            readHeader();
+            headerRead = true;
+        }
         if (definitionLevels == null) {
-            for (int i = 0; i < output.length; i++) {
-                output[i] = readLongValue();
+            int out = 0;
+            int len = output.length;
+            while (out < len) {
+                if (valuesRead == 0) {
+                    output[out++] = firstValue;
+                    lastValue = firstValue;
+                    valuesRead = 1;
+                    continue;
+                }
+                if (valuesRead >= totalValueCount) {
+                    throw new IOException("No more values to read");
+                }
+                if (bufferPos >= bufferFill) {
+                    loadNextMiniblockInternal();
+                }
+                int canDrain = Math.min(bufferFill - bufferPos, len - out);
+                System.arraycopy(miniblockBuffer, bufferPos, output, out, canDrain);
+                bufferPos += canDrain;
+                out += canDrain;
+                valuesRead += canDrain;
             }
         }
         else {
@@ -95,9 +117,31 @@ public class DeltaBinaryPackedDecoder implements ValueDecoder {
     /// Read INT32 values directly into a primitive int array.
     @Override
     public void readInts(int[] output, int[] definitionLevels, int maxDefLevel) throws IOException {
+        if (!headerRead) {
+            readHeader();
+            headerRead = true;
+        }
         if (definitionLevels == null) {
-            for (int i = 0; i < output.length; i++) {
-                output[i] = (int) readLongValue();
+            int out = 0;
+            int len = output.length;
+            while (out < len) {
+                if (valuesRead == 0) {
+                    output[out++] = (int) firstValue;
+                    lastValue = firstValue;
+                    valuesRead = 1;
+                    continue;
+                }
+                if (valuesRead >= totalValueCount) {
+                    throw new IOException("No more values to read");
+                }
+                if (bufferPos >= bufferFill) {
+                    loadNextMiniblockInternal();
+                }
+                int canDrain = Math.min(bufferFill - bufferPos, len - out);
+                for (int i = 0; i < canDrain; i++) {
+                    output[out++] = (int) miniblockBuffer[bufferPos++];
+                }
+                valuesRead += canDrain;
             }
         }
         else {
@@ -118,6 +162,7 @@ public class DeltaBinaryPackedDecoder implements ValueDecoder {
 
         if (valuesRead == 0) {
             valuesRead = 1;
+            lastValue = firstValue;
             return firstValue;
         }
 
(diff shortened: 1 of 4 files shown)


gunnarmorling:
Have you done some end-to-end measurements of this, @ZhaoMJ? Any results you could share?

gunnarmorling:
Thanks a lot, @ZhaoMJ. Closing this one in favor of #1063 which has your commit and a few further improvements.

Issue #837, opened 2026-08-06
Reusable footer handle: cache the parsed footer to skip the per-open footer read


gunnarmorling:
## Motivation

For point-query / online-lookup workloads, a serving process resolves a key to `(file, rowOrdinal)` via an external index and reads a few rows. When the process can keep a `ParquetFileReader` open per file, the parsed footer is already held (`getFileMetaData()`), so repeated lookups re-parse nothing. But a service indexing terabytes-to-petabytes of files **cannot keep readers (or even footers) resident for every file** — it caches footers for a working set behind an LRU and opens readers on demand. Today every such open pays a footer round-trip to object storage, which dominates the latency of an otherwise tiny read.

The fix is to let a consumer cache the footer once and hand it back on open, so the only I/O per lookup is the target row group's pages.

## API

Producer (instance — the reader already read these bytes at open):

```java
/// The raw footer bytes of this (single) file, as read at open. Suitable for
/// caching and later reuse via `open(InputFile, ByteBuffer, HardwoodContext)`.
/// Throws on a multi-file reader — see below.
public ByteBuffer getFooterBytes();
```

Consumer (static — skips the footer read entirely):

```java
/// Open a reader from an already-parsed/cached footer. No footer I/O is issued.
/// The caller guarantees `footer` was produced from `inputFile` (see Correctness).
public static ParquetFileReader open(InputFile inputFile, ByteBuffer footer,
                                     HardwoodContext context) throws IOException;
```

`open(file).getFooterBytes()` (open → get → close) also covers cache-warming, so no separate static `readFooterBytes(file)` is needed — one producer entry point.

## Why bytes, not a serialized object

The parsed `FileMetaData` came *from* bytes: the Thrift-compact footer at the tail of the file. That is already its canonical serialized form — native, spec-versioned by construction, identical to what other engines cache. Re-parsing on `open` is CPU-only (microseconds); the expensive thing we avoid is the object-storage round-trip, which caching bytes avoids just as well.

So we deliberately do **not** add a `Serializable FileMetaData` or a custom object codec — that would make Hardwood own and version a serialization contract to buy back only a trivial parse. The footer bytes are self-contained for planning: offset/column indexes live elsewhere in the file and are read lazily per row group.

## Correctness: staleness is the real hazard

[...]

PR #1058, opened 2026-08-28
#837 Install a MetadataSource on HardwoodContext to reuse parsed footers

ZhaoMJ:
_Via Claude_

## Why

Opening a reader on a warm file is mostly footer read plus Thrift parse: three `readRange`
calls and a parse linear in footer size, roughly 3 µs/KB plus 0.1 ms fixed — about 6 ms on a 2 MB
footer. A caller that cannot keep persistent readers (a serving process holding an index over more
files than it can hold open readers) pays this on every request, even for a read that touches one page.

The parsed metadata is already accessible through `getFileMetaData()`. This PR makes it possible
to supply it back to every subsequent open, through a single installation point on the context.

## What

Replaces the `open(InputFile, HardwoodContext, FileMetaData)` overload with a context-installed
`MetadataSource`, as proposed in the maintainer's review. The source handles single and multi-file
reads (`open` and `openAll`) uniformly, adds staleness detection, and is extensible to page indexes
and Bloom filters as additional defaulted methods.

### New API

```java
Map<String, ParsedFooter> cache = new ConcurrentHashMap<>();

MetadataSource source = file -> {
    try {
        return cache.computeIfAbsent(file.name(), k -> {
            try { return ParsedFooter.readFrom(file); }
            catch (IOException e) { throw new UncheckedIOException(e); }
        });
    } catch (UncheckedIOException e) { throw e.getCause(); }
};

try (HardwoodContext ctx = HardwoodContext.builder()
        .metadataSource(source)
        .build()) {
    ParquetFileReader.open(myFile, ctx);  // footer from cache on warm opens
}
```

Staleness handling:

```java
try { read(file); }
catch (StaleMetadataException e) {
    e.sourceIdentity().ifPresent(id -> cache.remove(file.name()));
    read(file);
}
```

### Types added

| Type | Role |
|---|---|
| `MetadataSource` | `@FunctionalInterface`: `footerOf(InputFile) → ParsedFooter` |
| `ParsedFooter` | Value: `FileMetaData` + `FileSchema` + lengths + `sourceIdentity`; `readFrom(InputFile)` |
| `StaleMetadataException` | Thrown when `sourceIdentity` disagrees with `InputFile.identity()` |
| `HardwoodContext.builder()` | New entry point with `.threads()` and `.metadataSource()` |
| `InputFile.identity()` | Default `Optional.empty()`; `MappedInputFile` overrides with `path:size:mtime` |

### Structural validation

When a `MetadataSource` is installed, the reader calls `ParquetMetadataReader.validateTrailer`
before trusting the cached offsets. It reads start and end magic, checks for an encrypted footer,
and verifies the trailer footer length matches `ParsedFooter.footerLength()`. This is ~12 bytes of
I/O on a file whose data pages will be read anyway, and it catches file replacement and truncation
even when no `sourceIdentity` is available. A size or footer-length mismatch throws
`StaleMetadataException`; a magic or encrypted-footer problem throws `ParquetReadException`.

## Tests

`ReuseFileMetadataTest`, five tests. The `metadataSourceSkipsFooterOnSubsequentOpens` test asserts
not just equal row counts but **exactly one read in the footer range** — the 8-byte trailer read from
`validateTrailer` — with zero reads of the full Thrift body. The staleness test verifies
`StaleMetadataException` is thrown when identity mismatches, and that a missing identity on either
side passes through silently. `contextWithoutSourceReadsNormally` confirms the no-source path is
unaffected.

Build: 13,962 tests, 0 failures, 0 errors, 2 skipped.

## Authorship

Check this box to confirm:

## Checklist

- [x] Build via `./mvnw clean verify` passes (excluding the pre-existing impsort parse failure on `ParquetFileReader.java`, which exists on `upstream/main` before this PR)

### DIFF
--- a/core/src/main/java/dev/hardwood/reader/ParquetFileReader.java
+++ b/core/src/main/java/dev/hardwood/reader/ParquetFileReader.java
@@ -189,28 +189,83 @@ public class ParquetFileReader implements AutoCloseable {
         return openAll(List.of(inputFile), context, readerConfig);
     }
 
+    /// Open a single Parquet file with a shared context, reusing metadata the
+    /// caller has already parsed.
+    ///
+    /// Equivalent to [#open(InputFile,HardwoodContext)] except that the footer is
+    /// not read or parsed: the supplied [FileMetaData] is used as-is, and the
+    /// [FileSchema] is derived from it exactly as the other `open` methods derive
+    /// it from the footer they read. Reading and parsing the footer is most of the
+    /// cost of opening a reader on a warm file - three `readRange` calls and a
+    /// Thrift parse, linear in footer size - so a caller that opens many readers
+    /// over one unchanging file can hoist it out of the request. The intended
+    /// shape is: open once, keep the metadata, open per read from it. Deriving the
+    /// [FileSchema] is the part that is *not* avoided, and it grows with the
+    /// number of columns rather than with footer size.
+    ///
+    /// The metadata comes from [#getFileMetaData()], which stays usable after its
+    /// reader is closed, so acquiring it needs no separate entry point:
+    ///
+    /// ```java
+    /// FileMetaData metaData;
+    /// try (ParquetFileReader reader = ParquetFileReader.open(InputFile.of(path), context)) {
+    ///     metaData = reader.getFileMetaData();
+    /// }
+    /// ```
+    ///
+    /// The metadata is treated as read-only by the reader, so one instance may
+    /// back any number of concurrent readers. It is not deeply immutable -
+    /// `Statistics` holds `byte[]` bounds - so a shared instance must not be
+    /// mutated once it has been handed over.
+    ///
+    /// **The caller is responsible for the metadata describing exactly these
+    /// bytes.** Every chunk offset and page index the reader uses comes from it,
+    /// so metadata from a different version of the file reads unrelated bytes
+    /// rather than failing cleanly. The reader's existing requirement that input
+    /// files are not modified while it is open therefore extends backwards to
+    /// whenever this metadata was read. A cache of parsed footers must key on a
+    /// content identity - an ETag or generation number, or [InputFile#length()]
+    /// paired with a modification time - and not on a path alone, because an
+    /// object store may replace the content behind one.
+    ///
+    /// [#getFileMetaData()] on a multi-file reader returns only the *first*
+    /// file's footer; reusing it for any other file of that reader is a misread,
+    /// not an error.
+    ///
+    /// @param fileMetaData the footer of `inputFile`, already parsed; must not be `null`
+    public static ParquetFileReader open(InputFile inputFile, HardwoodContext context,
+                                         FileMetaData fileMetaData) throws IOException {
+        if (fileMetaData == null) {
+            throw new IllegalArgumentException("fileMetaData must not be null");
+        }
+        return openInternal(List.of(inputFile), (HardwoodContextImpl) context, ReaderConfig.defaults(), false,
+                fileMetaData);
+    }
+
     /// Open multiple Parquet files with a dedicated context. The schema
     /// is read from the first file and is assumed to be common across all
     /// files. Files are opened on demand by the iterator; the first file is
     /// opened eagerly so any I/O or metadata error surfaces immediately.
     public static ParquetFileReader openAll(List<? extends InputFile> inputFiles) throws IOException {
-        return openInternal(inputFiles, HardwoodContextImpl.create(), ReaderConfig.defaults(), true);
+        return openInternal(inputFiles, HardwoodContextImpl.create(), ReaderConfig.defaults(), true, null);
     }
 
     /// Open multiple Parquet files with a shared context.
     public static ParquetFileReader openAll(List<? extends InputFile> inputFiles, HardwoodContext context) throws IOException {
-        return openInternal(inputFiles, (HardwoodContextImpl) context, ReaderConfig.defaults(), false);
+        return openInternal(inputFiles, (HardwoodContextImpl) context, ReaderConfig.defaults(), false, null);
     }
 
     /// Open multiple Parquet files with a shared context and an explicit
     /// [ReaderConfig].
     public static ParquetFileReader openAll(List<? extends InputFile> inputFiles, HardwoodContext context,
                                             ReaderConfig readerConfig) throws IOException {
-        return openInternal(inputFiles, (HardwoodContextImpl) context, readerConfig, false);
+        return openInternal(inputFiles, (HardwoodContextImpl) context, readerConfig, false, null);
     }
 
+    /// @param preParsedMetaData the first file's footer, already parsed, or `null` to read it
     private static ParquetFileReader openInternal(List<? extends InputFile> inputFiles, HardwoodContextImpl context,
-                                                   ReaderConfig readerConfig, boolean ownsContext) throws IOException {
+                                                   ReaderConfig readerConfig, boolean ownsContext,
+                                                   FileMetaData preParsedMetaData) throws IOException {
         if (inputFiles == null || inputFiles.isEmpty()) {
             throw new IllegalArgumentException("At least one file must be provided");
         }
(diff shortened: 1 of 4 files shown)


gunnarmorling:
Hey @ZhaoMJ, any plans to continue with this one?

ZhaoMJ:
Hey @gunnarmorling , yes, I’d like to see this land. I opened it as a draft PR because I was waiting for your feedback on https://github.com/hardwood-hq/hardwood/issues/837#issuecomment-5453388224, since this interface differs from what you originally proposed in that issue.

gunnarmorling:
Ah ok, gotcha. Will take a closer look and report back. Thanks for the quick reply!

gunnarmorling:
Hey @ZhaoMJ, finally getting back to this.

First, I agree, making `FileMetadata` reusable is valuable and the right approach. We also need to address multi-file reads though (i.e. passing a single file's metadata doesn't suffice), and staleness detection must be part of it. It would also be good to build this in a way, allowing to later on also pass in pre-read indexes and Bloom filters.

I've iterated with Claude for a bit along these lines; below is what we came up with. Would that make sense to you? If so, would you be up for building it? Happy to discuss it more as needed.

# Plan: Reusing Parsed Metadata Across Readers (#837)

**Status: Proposed.** Tracking issue: #837.

## Context

Opening a reader reads and parses the file's footer. The parse is Thrift-compact decoding linear in footer size — roughly 3 µs/KB, so about 0.3 ms on a 31 KB footer and 6.6 ms on a 2 MB one. Wide schemas, many row groups and per-chunk statistics all push footers into that range.

A caller that keeps one reader per file pays this once. A serving process resolving a key to `(file, rowOrdinal)` through an external index cannot: it holds an index over more files than it can hold readers for, so it opens a reader per request and pays the parse on every one, ahead of a data read that may be a few milliseconds. `FileMetadataCache` caches footers only for the lifetime of one reader, so nothing carries across.

The I/O is a smaller term than it appears. `S3InputFile.open()` already issues a suffix-range `GET` that discovers the file length and pre-fetches the last 64 KB in the same round-trip, so a footer within that window costs no additional request. Above it, one extra request is issued — and that is the same regime in which the parse dominates.

This design lets a caller supply metadata the reader would otherwise read, through a source installed on the context. It covers single- and multi-file reads through one installation point and adds no `open` overloads.

## Design

### `dev.hardwood.MetadataSource`

```java
@FunctionalInterface
public interface MetadataSource {

    /// The parsed footer of `file`.
    ParsedFooter footerOf(InputFile file) throws IOException;
}
```

```java
MetadataSource source = file -> cache.computeIfAbsent(key(file), k -> ParsedFooter.readFrom(file));
```

The footer is the only kind this interface carries today, and the one kind every reader acquires: it is read on every open, whatever the read then does, while the page index and bloom filters are read only when a filtered or page-level plan calls for them. `footerOf` is therefore the single abstract method. The interface is shaped to gain those further kinds — which the footer points to rather than contains (see [Later methods](#later-methods)) — as **defaulted** methods that perform the read the reader would otherwise have performed. An implementation then supplies only the kinds it holds, the rest behave as they do today, and because a default does not count as an abstract method the interface stays functional and existing lambdas keep compiling.

The source **owns acquisition**, rather than being a lookup the reader falls back from. `footerOf` must return a footer; on a miss the implementation calls `ParsedFooter.readFrom(file)` itself. Every policy question — what is worth retaining, eviction, sizing, and whether a failing backend degrades or propagates — therefore resolves in the implementation, in one place, with the caller's own key.

Population needs no separate entry point. The first read over a file fills whatever the implementation retains as a side effect of the miss branch, so there is no warm-up recipe and no requirement to harvest footers from a reader before closing it.

### `dev.hardwood.metadata.ParsedFooter`

```java
public final class ParsedFooter {

    public FileMetaData metaData();
    public FileSchema schema();
    public long footerLength();
    public long fileLength();
    public Optional<String> sourceIdentity();

    public static ParsedFooter readFrom(InputFile openFile) throws IOException;
}
```

A final class whose only entry point is `readFrom`: every instance originates from Hardwood reading a file, so the reader is never handed chunk offsets a caller composed by hand, and `fileLength` and `sourceIdentity` are always populated from the file the footer was read from.

It carries more than `FileMetaData`, which is the `parquet.thrift` `FileMetaData` struct (fields 1–7) and holds only what the file states. `footerLength` and `fileLength` are properties of the encoding and the file, `sourceIdentity` of the storage backend, and `FileSchema` is Hardwood's resolution of the schema elements. Carrying `FileSchema` matters on its own: deriving it grows with column count and happens per file, so a supplied footer skips it too.

A `ParsedFooter` is held in memory and is not a serialization format. Hardwood's Thrift readers skip every field they do not model, at each nesting level, so a footer parsed from a file and written back out is not the footer that was read: it is Hardwood's view of it, missing whatever this version does not decode. Persisting footers across processes therefore requires the file's **original** footer bytes rather than a re-serialization of the parse, and is out of scope here — see #837. This matches the stability policy already stated for `dev.hardwood.metadata`, under which those records may gain components between releases; a supported wire format over them would contradict it.

### `InputFile.identity()`

```java
/// An identifier that changes when the file's content changes, if the backend can
/// supply one cheaply. Empty when it cannot.
default Optional<String> identity() throws IOException { return Optional.empty(); }
```

Resolved during `open()` and fixed for the file's lifetime: it names the content the file was opened against, not the current remote state. That also makes the reader's existing requirement — input files are not modified while a reader is open — checkable rather than advisory.

| Backend | Identity | Cost |
|---|---|---|
| S3 | `ETag`, from the response to the suffix-range `GET` in `S3Fetcher.open()` | none; the response is already fetched |
| Local | `(size, mtime, fileKey)` from `BasicFileAttributes` | one stat during `open()` |
| In-memory | empty | — |

A default method, so existing implementations are unaffected.

### Installation

```java
try (HardwoodContext context = HardwoodContext.builder()
        .threads(16)
        .metadataSource(source)
        .build()) {
    // every open() and openAll() against this context consults the source
}
```

`HardwoodContext.create()` and `create(int)` remain, and a context without a source behaves exactly as today. The source is consulted at the two places the reader acquires a footer: `ParquetFileReader.openInternal` for the first file, and `FileMetadataCache.loadFile` for later files of a multi-file reader — lazily, one file at a time, so partial coverage needs no special handling.

### Validation

Metadata that does not describe the bytes being read is the hazard this design has to answer for: every chunk offset and page location comes from it, so a mismatched footer reads unrelated bytes and reports plausible values rather than failing. The reader validates on every load.

| Check | Cost | Catches |
|---|---|---|
| Structural: the maximum byte the metadata claims does not exceed `length()`; row-group row counts sum to `numRows` | none beyond `length()`, already called | truncation, replacement by anything shorter, incoherent metadata |
| Trailer: recorded `footerLength` and `fileLength` match the file's trailer | none on S3 (within the pre-fetched tail) or mmap | most rewrites |
| Identity: `sourceIdentity()` matches `identity()`, when both are present | none | content replacement, soundly |

The caller's key and the reader's check do different jobs and both are needed. Keying on identity makes a changed file a cache miss, which is the ordinary path. The reader's check makes a mis-keyed cache an error rather than a silent misread.

A failed check throws `StaleMetadataException`, naming the file, the check, and both identities. It does not extend `ParquetReadException`, which is documented for files whose own bytes are wrong and where retrying will not help; here the file is intact and retrying after eviction is the remedy:

```java
try {
    read(file);
}
catch (StaleMetadataException e) {
    cache.invalidate(e.sourceIdentity());
    read(file);
}
```

Where identity is absent on either side, the structural and trailer checks still run.

### Concurrency

`FileMetadataCache.loadFile` runs on the decode pool, so `footerOf` is called from pool threads for several files concurrently, and on the calling thread for the first file. Implementations must be thread-safe and must not block on a network: a remote store belongs behind an in-process tier.

## Scope

`footerOf` only. `ParsedFooter`, `InputFile.identity()`, `StaleMetadataException` and `HardwoodContext.builder()` ship with it.

Sources are in-process. Persisting parsed metadata beyond the lifetime of a JVM needs the original footer bytes, which no public API exposes; #837 covers that case.

### Later methods

The page index and bloom filters are pointed to by the footer, not contained in it, and are read lazily per row group. Their granularities differ from the footer's and from each other, so their signatures must carry enough context for an implementation to coalesce:

| Kind | Natural unit | Signature must carry |
|---|---|---|
| Page index | the file's `[all ColumnIndexes][all OffsetIndexes]` region, contiguous between the data and the footer | `FileMetaData`, so an implementation may fetch the file-wide region once or one row group's slice |
| Bloom filters | one per column chunk, position varying by writer | the row group, so sibling chunks can be fetched together |

Both would exchange types that do not exist publicly yet — a parsed page-index container, and `BloomFilter`, currently `dev.hardwood.internal.bloomfilter`. Neither is justified until measured.

## Documentation

A how-to page under `docs/content/how-to/` covering acquisition, installation on a context, and the caching key requirement. `ParsedFooter`, `MetadataSource`, `InputFile.identity()` and `StaleMetadataException` carry public JavaDoc; the release notes gain an entry.

gunnarmorling:
Hey @ZhaoMJ, just got reminded of this and thought I'd check in. How is it going, anything we should talk about?

ZhaoMJ:
Hey @gunnarmorling, sorry I didn't really get a chance to work on this until today. The design you proposed makes a lot of sense to me, and I really like it. I’ve pushed some changes accordingly. Please take a look when you get a chance. Thanks!

PR #1060, opened 2026-08-30
#1053 Add predicates to filter JFR events

### LINKED ISSUE #1053: Name the predicate in the Filter JFR events, with literals elided
## Summary

The four `Filter`-category JFR events report how much a filter skipped, but not **which** filter did
it. In a process that runs many different reads, the counts cannot be attributed:

| Event | Reports | Missing |
|---|---|---|
| `dev.hardwood.RowGroupFilter` | row groups dropped by statistics/bloom push-down | which predicate |
| `dev.hardwood.PageFilter` | pages dropped by the Column Index | which predicate |
| `dev.hardwood.RecordFilter` | records dropped by per-record evaluation | which predicate |
| `dev.hardwood.RowGroupByteRangeFilter` | row groups excluded by split selection | which range |

Thread and timestamp are the only correlation available today, and nothing else in the recording
names the predicate, so there is nothing to join them to.

Proposal: add a `predicate` field to each, carrying a **structural rendering with the literals
elided**:

```
and(gt(id, ?), eq(label, ?))
or(isNull(city), lt(zip, ?))
```

## Why literals are elided

Two reasons, and they point the same way.

**Sensitivity.** Every field these events carry today is schema or structure — file paths, column
names, counts, strategy strings. Predicate literals would be the first *data-domain* values to enter
a Hardwood recording. Recordings get attached to bug reports and uploaded to profiling services, and
`eq(email, 'someone@example.com')` sitting in a JFR chunk is an exfiltration path that nobody
auditing the reader would think to look for. Column names and operators add no class of information
the sibling events don't already expose (`PageFilter.column`, `RowGroupScanned.column`).

**Cardinality.** With literals, the distinct-string count is the number of query parameterizations —
unbounded in a service. Without them it is the number of distinct query *shapes*, which is small and
stable (the `pg_stat_statements` bound). The events themselves are low-volume — `RecordFilter` and
`RowGroupFilter` fire once per file per read — and the predicate is a property of the read, so one
rendered string covers every event of that read and JFR interns it once per chunk.

If literals are ever genuinely wanted, JFR's own mechanism fits: a custom `@SettingDefinition` on the
event, so an operator enables them per recording and the default stays off. Not part of this issue —
but shipping the shape now forecloses nothing.

## Implementation notes

- **Write an explicit renderer over the sealed `ResolvedPredicate`; do not use the records' generated
  `toString()`.** It leaks every literal by construction, and it is useless exactly where it matters:
  `BinaryPredicate.value` is a `byte[]`, so it renders `[B@1a2b3c`, and `IntInPredicate` /
  `LongInPredicate` / `BinaryInPredicate` print array identities. An exhaustive switch over the
  sealed interface also fails to compile when a predicate type is added, instead of silently
  rendering it wrong.
- **The records hold `columnIndex`, not names**, so the renderer needs the `FileSchema`. Both are in
  hand at reader-build time. Render once per read and hand the string down, rather than per event.
- **Render lazily, behind `isEnabled()`.** A read with no recording running should pay nothing; for
  `RecordFilter` the natural place is inside `RecordFilterTally.emit()`.
- `IsNull` / `IsNotNull` / geospatial predicates have no literal to elide, so they render in full.
- Cover the `*In` predicates explicitly: render arity (`in(id, ?×5)` or similar), not the values.

## `RowGroupByteRangeFilter` is the odd one out

Its predicate is a `RowGroupPredicate.ByteRange`, not a `FilterPredicate` — two file offsets, which
are neither user data nor schema, so the sensitivity argument does not apply and there is nothing to
elide. Cardinality is a different shape too: split boundaries vary per task, so `byteRange(0, 134217728)`
is effectively unique per split.

Suggested resolution: render it in full (`byteRange(0, 134217728)`, `and(...)` for compound
predicates) and accept the cardinality — the event fires once per read, and a split's offsets are
exactly what an operator debugging uneven split assignment wants to see. Worth deciding explicitly
rather than by default.

## Acceptance criteria

- A `predicate` field on `RowGroupFilterEvent`, `PageFilterEvent`, `RecordFilterEvent` and
  `RowGroupByteRangeFilterEvent`, populated on every emission where a predicate exists.
- A renderer with an exhaustive switch over `ResolvedPredicate`, covering every implementing record,
  with no literal reaching the output.
- A test asserting no literal value appears in the rendered string, for each predicate type that
  carries one — including the binary and `*In` shapes, where a naive `toString` would also be wrong.
- `docs/content/reference/configuration.md` field lists updated, stating that literals are elided.

Follows from the discussion on #1043, which wired these events up.

wyfryj:
## Authorship

- [] This change is coming from a human who understands what it does and why, and can discuss it in review

## Checklist

## Summary

- add exhaustive renderers for `ResolvedPredicate` and `RowGroupPredicate`
- elide predicate literals while preserving operators, column names, and IN arity
- add predicate information to the four filter-related JFR events
- render predicates lazily after `isEnabled()` and cache the resolved predicate per read
- add renderer and JFR integration tests
- document the new JFR fields

Fixes #1053

## Testing

- Error Prone checks: 10 tests passed
- Hardwood Core: 2771 tests run, 0 failures, 0 errors, 2 skipped
- Core integration tests: 14 passed
- `git diff --check`: passed

The full reactor reached `hardwood-s3`, where tests could not run because Docker was unavailable and the external `aws-c-auth` repository clone timed out. The affected core module completed successfully.

## AI assistance

AI assistance was used during implementation. I reviewed the changes, understand the predicate rendering, caching, and JFR event wiring, and can discuss them during review.

### DIFF
--- a/core/src/main/java/dev/hardwood/internal/predicate/ResolvedPredicateRenderer.java
+++ b/core/src/main/java/dev/hardwood/internal/predicate/ResolvedPredicateRenderer.java
@@ -1,0 +1,101 @@
+/*
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Copyright The original authors
+ *
+ *  Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package dev.hardwood.internal.predicate;
+
+import java.util.List;
+
+import dev.hardwood.reader.FilterPredicate;
+import dev.hardwood.schema.FileSchema;
+
+/// Renders an execution predicate for diagnostics without exposing its literal values.
+///
+/// Column names and predicate structure are retained so events from separate reads can be
+/// attributed to a query shape. Data-domain values are replaced with `?`; `IN` predicates
+/// additionally retain their arity. Geospatial bounds describe the queried region rather than
+/// values read from the file and are rendered in full.
+public final class ResolvedPredicateRenderer {
+
+    private ResolvedPredicateRenderer() {
+    }
+
+    /// Renders `predicate` using the unambiguous field paths from `schema`.
+    public static String render(ResolvedPredicate predicate, FileSchema schema) {
+        StringBuilder rendered = new StringBuilder();
+        append(rendered, predicate, schema);
+        return rendered.toString();
+    }
+
+    private static void append(StringBuilder rendered, ResolvedPredicate predicate, FileSchema schema) {
+        switch (predicate) {
+            case ResolvedPredicate.IntPredicate p -> appendComparison(rendered, p.op(), column(schema, p.columnIndex()));
+            case ResolvedPredicate.LongPredicate p -> appendComparison(rendered, p.op(), column(schema, p.columnIndex()));
+            case ResolvedPredicate.FloatPredicate p -> appendComparison(rendered, p.op(), column(schema, p.columnIndex()));
+            case ResolvedPredicate.Float16Predicate p -> appendComparison(rendered, p.op(), column(schema, p.columnIndex()));
+            case ResolvedPredicate.DoublePredicate p -> appendComparison(rendered, p.op(), column(schema, p.columnIndex()));
+            case ResolvedPredicate.BooleanPredicate p -> appendComparison(rendered, p.op(), column(schema, p.columnIndex()));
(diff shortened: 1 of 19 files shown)


gunnarmorling:
Thanks, @wyfryj. Closing this one for now as it has been in Draft state for more than a week. Feel free to reopen once it's complete (and rebased to current main).

PR #1064, opened 2026-08-30
#965 Validate fixed-length byte array width before allocation

### LINKED ISSUE #965: Decoding FIXED_LEN_BYTE_ARRAY column with no width throws unboxed NullPointerException
### Summary

`SchemaElement.type_length` is an optional Thrift field, but a `FIXED_LEN_BYTE_ARRAY` column requires it to decode values. Hardwood currently accepts this incomplete schema and later fails through Java null unboxing.

`FileSchema.fromSchemaElements` accepts such a schema. The column then holds `null` for `typeLength()`.

The reader converts that `null` to an `int`. Java throws a bare `NullPointerException`. The message names no column and no file.

### Problem

`BatchExchange.allocateArray` allocates the batch buffer from the column width:

```java
case FIXED_LEN_BYTE_ARRAY -> {
    int width = column.typeLength(); // null unboxing throws NullPointerException
    byte[] bytes = new byte[Math.multiplyExact(width, capacity)];
    ...
}
```

`NestedColumnWorker.growBinaryBatchValues` contains the same unchecked conversion when a nested fixed-width buffer grows. Normal nested allocation already calls `BatchExchange.allocateArray`, so the growth-site check is a second defensive validation site rather than an independent initial failure path.

The writer does better. `ParquetFileWriter.requireTypeLength` and `ValueEncoder.requireTypeLength` throw an `IllegalArgumentException` that names the column.

`hardwood schema` prints such a file without error, because it reads the schema only. The failure starts when the reader touches the column data.

### Affected code
`dev.hardwood.internal.reader.{BatchExchange#allocateArray, NestedColumnWorker#growBinaryBatchValues}`

### Fix

1. Keep `FileSchema.fromSchemaElements` tolerant. Metadata tools must read incomplete metadata.
2. Check `column.typeLength()` for `null` in both allocation paths, before the conversion to `int`.
3. If the width is absent, throw an `IllegalArgumentException` that names the column field path.
4. Keep the reader exception handling that adds the file name.

### Reproducer

```java
SchemaElement root = new SchemaElement("schema", null, null, null, 
                                       1, null, null, null, 
                                       null, null);
SchemaElement digest = new SchemaElement("digest", PhysicalType.FIXED_LEN_BYTE_ARRAY, null, RepetitionType.REQUIRED,
                                         null, null, null, null, 
                                         null, null);
FileSchema schema = FileSchema.fromSchemaElements(List.of(root, digest));
ColumnSchema column = schema.getColumn("digest");

assertThat(column.typeLength()).isNull();
assertThatThrownBy(() -> BatchExchange.allocateArray(column, 1024))
        .isInstanceOf(NullPointerException.class);
```

chlgusrbs0:
Fixes #965

## Authorship

Check this box to confirm:

## Checklist

### DIFF
--- a/core/src/main/java/dev/hardwood/internal/reader/BatchExchange.java
+++ b/core/src/main/java/dev/hardwood/internal/reader/BatchExchange.java
@@ -282,7 +282,11 @@ public class BatchExchange<B> {
                 yield bbv;
             }
             case FIXED_LEN_BYTE_ARRAY -> {
-                int width = column.typeLength();
+                Integer width = column.typeLength();
+                if (width == null) {
+                    throw new IllegalArgumentException(
+                            "FIXED_LEN_BYTE_ARRAY column " + column.fieldPath() + " has no type length");
+                }
                 byte[] bytes = new byte[Math.multiplyExact(width, capacity)];
                 int[] offsets = new int[capacity + 1];
                 for (int i = 0; i <= capacity; i++) {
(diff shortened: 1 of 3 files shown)


gunnarmorling:
Thanks, @chlgusrbs0. Gonna close this in favor of #1081, as validation can and should happen earlier.

PR #1075, opened 2026-09-03
#1047 Discard output on close when a write has failed

### LINKED ISSUE #1047: close() after a failed write publishes the rows that got through, and there is no way to abandon a writer
A `writeBatch` that throws leaves the writer open. `ParquetFileWriter` has no `abort()`, so try-with-resources — the idiom every how-to shows — calls `close()`, which *finishes the file*.

```java
try (ParquetFileWriter w = ParquetFileWriter.create(OutputFile.of(target), schema)) {
    ColumnWriter cols = w.columnWriter();
    cols.writeBatch(b -> b.ints("v", new int[] { 1, 2, 3 }));
    try {
        cols.writeBatch(b -> b.ints("v", new int[] { 4, 5 }).ints("v", new int[] { 6 }));
    }
    catch (RuntimeException expected) { /* caller notices and stops */ }
}
```

Measured:

```
writeBatch threw: IllegalArgumentException
published=true size=191 leftovers=[out.parquet]
readable, rows=3
```

The caller gets a valid, readable, published file holding the first three rows. No temporary artefact is left and `close()` reports nothing.

This is not corruption — the file is well-formed. It is silent truncation of intent, and it contradicts the contract stated in `docs/content/concepts/write-model.md`:

> **A failure leaves nothing behind.** When the writer cannot finish, it discards what it has written.

That holds for a failure inside `create` (#1035) and for one inside `close`, which discards rather than letting `out.close()` publish a truncated file. It does not hold for a failure between them, which is the window a caller is actually in.

## Why it matters before 1.1.0.Beta1 rather than after

Adding `abort()` later is additive and harmless. Changing what `close()` does after a failed write is a **behaviour break**, so which of the two is the default wants deciding while the surface is still unfrozen.

The options, as I see them:

- **Remember the failure and discard on close.** A writer that has thrown once refuses to publish; a caller who wants the good prefix asks for it explicitly. Matches the documented contract and the project's fail-early principle, and makes try-with-resources safe by default.
- **Keep publishing, and add `abort()`.** The batches that succeeded are legitimately in the file; a caller who does not want them says so. Needs the contract in `write-model.md` reworded, because it currently promises the opposite.
- **Distinguish the two failure kinds.** A caller error before anything was appended (a ragged batch, a type mismatch) leaves the file consistent and could publish; an `IOException` part way through a flush cannot. This is the most faithful and the most machinery.

I have no strong preference between the first two; the third is only worth it if the distinction turns out to be cheap to draw.

mohamedibrahim54:
Going with the first option, ParquetFileWriter now tracks a failed flag. Under the default WriteFailurePolicy.DISCARD, close() calls OutputFile.discard() instead of finishing the file, so a try-with-resources block never silently publishes a truncated prefix. WriteFailurePolicy.COMMIT_PREFIX opts back into the old behaviour for callers that want whatever arrived before the break.

A failed writer also rejects further writeBatch calls with IllegalStateException, making the poisoned state visible immediately rather than at close time.

### Decisions:
1. After the writer is poisoned, if the caller attempts another writeBatch call, Reject it with IllegalStateException — the writer's internal state may be corrupted, so allowing more data would compound the problem and could produce a silently wrong file.
2. for RowWriter.writeRow(), only the columnar I/O path can poison the writer; a pure validation failure in the row-planning layer cannot, check test WriterFailurePolicyTest.rowWriterRecoveryDoesNotPoison().
3. Metadata setters remain usable: no exception after the writer is failed also the file will discard eventually. should keep this behavior or block them too for consistency?

Closes #1047
 
## Authorship

Check this box to confirm:

## Checklist

### DIFF
--- a/core/src/main/java/dev/hardwood/writer/ColumnWriter.java
+++ b/core/src/main/java/dev/hardwood/writer/ColumnWriter.java
@@ -46,13 +46,18 @@ public final class ColumnWriter {
     /// populated (columns addressed by index or name), then submits it. There is no
     /// separate build or submit step to forget.
     ///
+    /// A failure anywhere in the batch — validation or I/O — marks the writer as failed.
+    /// Under the default [WriteFailurePolicy#DISCARD], a failed writer refuses subsequent writes
+    /// and discards the output on [ParquetFileWriter#close()] rather than publishing a
+    /// truncated file.
+    ///
     /// @param filler populates the batch's columns; must cover every column exactly once
     /// @throws IOException if the write fails
     /// @throws IllegalArgumentException if the batch does not cover every column, or its
     ///         per-layer inputs do not agree on a record count
     /// @throws UnsupportedOperationException if the schema has a shape the writer cannot
     ///         produce
-    /// @throws IllegalStateException if the writer is closed
+    /// @throws IllegalStateException if the writer is closed, or a previous write has failed
     public void writeBatch(Consumer<ColumnBatch> filler) throws IOException {
         writer.ensureOpen();
         writer.writeStagedBatch(filler);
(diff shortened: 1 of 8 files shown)


gunnarmorling:
Closing in favour of #1251, which is based on this work, rebasing to current main, and doing some tweaks.

PR #1097, opened 2026-09-05
#1093 Say where a read failed, once, from the scope it failed in

### LINKED ISSUE #1093: Reader exceptions should carry the row group and column the read failed in
A read failure currently reports the file and, sometimes, the row group: `[data.parquet] Failed to fetch metadata for row group 0`, or just `Unknown field type: 15` from the Thrift layer. Neither tells the reader *where* in the file to look, so diagnosing a corrupt file means bisecting it by hand.

`dev.hardwood.internal.ExceptionContext` already centralises file-name enrichment and `ColumnWorker` applies it at the retrieve / decode / drain boundaries. Extending that context to carry the column path and the byte offset of the region being read would give messages like:

```
[data.parquet] row group 0, column `id`, page header at byte 41104: Unknown field type: 15
```

This is what makes #1092's error overlay actionable, and it improves the non-interactive commands (`print`, `convert`, `inspect`) at the same time.

**Constraint: no cost to a read that does not fail.** The enrichment itself only runs inside `catch` blocks, so it is already free in the steady state. The care is needed for the context the enrichment needs to *have available*:

- The column path is fixed per `ColumnWorker`, so it is a final field, not per-page state.
- A byte offset is per page. `ColumnWorker` already keeps per-page state in `fileNameBuffer[slot]`, written once per page in the retriever — a `long` alongside it is the same shape and the same granularity.
- Nothing may be added inside the per-value decode loops.

Acceptance: a before/after benchmark on the read path showing no regression.

gunnarmorling:
Closes #1093. **Stacked on #1134** (`1093-thrift-field-names`), which names the field and struct a Thrift parse stopped on. This branch sits on that one, so until #1134 merges the diff below shows both. The base still reads `main` because #1097 is already the bottom of the stack carrying #1098; it becomes correct when #1134 merges. On its own this change is 57 files, +1744/−394.

A read that fails names the file and nothing else, which is the wrong grain for a format where the interesting question is always *which part*. `Failed to fetch metadata for row group 0` does not say which column, and a corrupt page arrives as `Unexpected EOF while reading fixed-length byte array` naming nothing at all.

## How

Where a failure happened is a property of the work in progress, not of the exception: the code that knows it is reading a bloom filter is the code running, and by the time a failure reaches a frame that can report it, that knowledge is gone. Every catch site that named a region was restating what its own call stack had already said.

So a region is *entered* rather than passed, and a failure takes its place at the moment it is raised — nearer the failure than any frame catching it can be — and renders it when the message is read. Raise sites throw plain exceptions carrying only what is wrong:

```java
try (ReadScope.Scope page = ReadScope.region(Region.DICTIONARY_PAGE, chunkStart)) {
    ...
    throw new ParquetReadException("Malformed Parquet metadata: the dictionary page lies"
            + " after the first data page at offset " + dataPageOffset);
}
```

No site states a position. 41 scope entries stand in for the 54 places that named a file, row group, column, region or offset in message text; the 24 that remain name a file and nothing more.

## What a caller sees

Every file in `parquet-testing`'s `bad_data`, before and after:

| fixture | main | this PR |
|---|---|---|
| ARROW-GH-47662 | `Unexpected EOF while reading fixed-length byte array` | `row group 0, column flba_field, data page — …` |
| PARQUET-1481 | `Invalid or corrupt physical type value: -7` | `footer at byte 289 (0x000121) — …` |
| ARROW-GH-41317 | `…ColumnMetaData.encodings declares…` | `footer at byte 35530 (0x008aca) — …` |
| ARROW-GH-41321 | `Invalid dictionary index bit width: 254` | `row group 0, column int64, data page — …` |
| ARROW-GH-45185 | `Invalid column chunk for 'element': …` | `row group 0, column x.list.element, batch assembly — …` |
| ARROW-RS-6229-DICTHEADER | `readRange(4, 2589) out of bounds (533 bytes)` | `row group 0, column nation_key, chunk fetch at byte 4 (0x000004) — Malformed Parquet metadata: the region of 2589 bytes at offset 4 runs past the end of the file` |
| ARROW-GH-43605 | `Cannot read ZSTD-compressed…` | unchanged — deliberately, see below |
| ARROW-RS-6229-LEVELS | `Cannot read SNAPPY-compressed…` | unchanged — deliberately |

The DICTHEADER line is the notable one: a footer pointing a chunk past EOF reached the input file's range check and arrived naming a byte range and nothing else. `InputFile.readRange` still raises `IndexOutOfBoundsException` — from outside the reader that is a bad offset, not a bad file — and the fetch, which knows the offset came from the footer, says which it is.

With #1134 underneath, a corrupt footer field gets both halves of the message the issue asks for:

```
[<memory>] footer at byte 195 (0x0000c3) — SchemaElement.type — Unknown field type: 13
```

`CorruptFooterNamingTest` asserts exactly that, deriving the expected byte from the byte it damaged.

## Properties that hold by construction

These are the reason for the shape, not side effects of it.

**A message is described once.** The place is a `final` field taken at construction, so two frames cannot describe the same failure. The check for whether message text had already been prefixed is deleted rather than reimplemented.

**A correct file we cannot read is named but never positioned.** `UnsupportedFileException` carries no place, because the file is not at fault and there is nothing at any byte to go and look at — the remedy is the same whichever column met it first. That is why the two codec messages above are untouched: not a judgement at a raise site, but a consequence of the type.

**Being described keeps the exact class.** The place is rendered into the message rather than wrapped around the exception. Rebuilding turned `SchemaIncompatibleException` into a plain `ParquetReadException`, which callers catch on. Nothing wraps to describe, so every cause chain is one layer shallower.

**The byte is the region's own address**, which is the answer everywhere but two. A region entered with no address names no byte — `DATA_PAGE`, where a decode fails an arbitrary distance in, so pointing at the page's first byte would send a reader to bytes that are intact; the row group and column already say which page it was. And a read that knows how far it got narrows the place before raising, through `ReadScope.stoppedAt`. `ThriftCompactReader` is its only caller, and reports the field header it stepped onto rather than the byte it came to rest on, which is one further along.

## Cost

`ExceptionContext`'s enrichment pass and its reflective rebuild, the `Exceptions` builder and its six entry points, three `Contextual*` carriers and `ReadContextAware` are all deleted. `ReadScope`, `PlacedIOException` and `UnsupportedFileException` replace them. The change is in how many places have to know about position, not in volume.

A full column scan of a 197-page chunk measures the same as before — p50 15.5 ms either way over three runs of 400 iterations — which bounds the per-page cost of entering a scope below the noise.

## Also fixed

`ColumnWorker`'s three outermost catches are the only thing that reports a failure at all: the decode task's own failure is discarded by the future that ran it, and each thread's dies with the thread. A throw while describing one left nothing to set the error and nothing to wake the drain, so the read hung rather than failing — with no exception and no log line to say why. Describing is now guarded, and `ColumnWorkerTest` pins that a failure while describing cannot take the original down with it.

## Split out

Rejecting a bloom filter or dictionary chunk that the footer places outside the file is #1135. Both checks were written here and taken back out: they change *which* exception a corrupt footer produces, which is a different question from where a failure happened. The chunk-fetch path keeps its check, because that is what fixes DICTHEADER above.

## Docs

`_designs/EXCEPTION_MODEL.md` gains the position layer; `docs/content/reference/error-handling.md` gains the message shape, part by part.

### DIFF
--- a/_designs/EXCEPTION_MODEL.md
+++ b/_designs/EXCEPTION_MODEL.md
@@ -196,6 +196,60 @@ in a separate file, a row group whose page-index region spans more than
 cache. Each of those files is correct and another reader will open it. Whether the
 last two limits are worth keeping is tracked as #1113.
 
+## Where a failure happened
+
+The type says what kind of failure it is; the message says where the read was when
+it happened. Both are properties of the failure rather than of any frame that
+catches it, so both are taken at the moment it is raised.
+
+**The place is entered, not passed.** `ReadScope` holds a `ThreadLocal<Place>` —
+file name, row group, column, region, and the region's first byte — that a frame
+narrows with try-with-resources:
+
+```java
+try (ReadScope.Scope footer = ReadScope.region(Region.FOOTER, footerStart)) {
+    return FileMetaDataReader.read(new ThriftCompactReader(buffer));
+}
+```
+
+The code that knows it is reading a bloom filter is the code running. By the time a
+failure has travelled to a frame that could report it, that knowledge is gone, and
+every catch site that names a region is restating what its own call stack already
+said. Scopes nest and the innermost wins, which is the rule the reader wants: a
+page header read inside a chunk fetch is a page-header failure.
+
+`ParquetReadException` and `PlacedIOException` read the scope in their constructors
+and render it in `getMessage()`. No site states a position, nothing enriches a
+failure on the way out, and nothing rebuilds one.
+
+Three properties follow from that shape rather than from care at any one site:
+
+- **A failure is described once.** A place is taken once, at construction, so a
+  failure travelling out through further scopes cannot collect a second prefix.
+  Nothing inspects a message to find out whether it has one already.
+- **A correct file is named but never positioned.** `UnsupportedFileException`
+  carries the file name and no place, because it is not the type that carries one.
+  There is nothing at any byte for a caller to go and look at.
+- **Describing a failure keeps its exact type.** The place is rendered into the
+  message, not wrapped around the exception, so a `SchemaIncompatibleException`
+  that a caller catches on is still one after it has been described.
+
+Work that crosses a thread boundary carries its place as data. A page is read on the
+retriever thread and decoded on a pool thread, which inherits nothing; the decode
+task re-enters the page's place with `ReadScope.resume` before it runs, so a failure
+inside it is built exactly as one on the retriever thread would be. `ColumnWorker`'s
+outermost catches do the same before restating what a decoder threw.
+
+**Which byte.** The region's own first byte is the answer for almost every failure:
+`RowGroupDictionaryFilterSource` enters one scope and has six throws that all want
+the chunk's start. Two cases differ. A region that has no address of its own —
+`DATA_PAGE`, where a decode fails an arbitrary distance into the page — is entered
+without one and names no byte, since the row group and column already say which page
+it was. A read that knows how far it got says so through `ReadScope.stoppedAt`,
+which narrows the place to that byte before building the failure;
+`ThriftCompactReader` is the only caller, and reports the field header it stepped
+onto rather than the byte it came to rest on, which is one further along.
+
 ## A truncated buffer
 
 `ThriftTruncatedException` is a `ParquetReadException` and reads as one everywhere
--- a/cli/src/main/java/dev/hardwood/cli/command/InspectDictionaryCommand.java
+++ b/cli/src/main/java/dev/hardwood/cli/command/InspectDictionaryCommand.java
@@ -182,7 +182,8 @@ public class InspectDictionaryCommand implements Command<CommandInvocation> {
             ColumnChunk chunk, HardwoodContextImpl context, String fileName, int rowGroupIndex,
             long dictionaryOffset) {
         try (ReadScope.Scope scope = ReadScope.file(fileName)
-                .column(rowGroupIndex, columnSchema.name())
+                .rowGroup(rowGroupIndex)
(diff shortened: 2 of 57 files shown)


gunnarmorling:
Closing this in favour of a much smaller change.

The premise this grew from was that a read failure cannot name the row group and column without a scope mechanism to carry them. That is only true of the *byte offset*. A byte offset is known deep inside a parse — the Thrift reader position, where a page header starts, where a chunk fetch began — and there is no catch boundary down there to hand it to, which is what justified a `ThreadLocal` scope, the `Region` enum, `PageInfo.fileOffset()`, and region entries across the footer reader, both fetch plans, the page decoder, the index reader and the bloom-filter and dictionary sources. Roughly 15 of the 29 main files this touches exist for that half alone and never see a column.

The row group and column need none of it. `ColumnWorker` already holds the column as a final field, already writes `fileNameBuffer[slot]` once per page in the retriever, and already routes all three of its catches through one helper. Widening that helper and adding an `int[]` beside the existing `String[]` gets `[data.parquet] row group 0, column category — CRC mismatch: …`, which is most of the diagnostic value on a wide multi-row-group file, without a new mechanism and without touching the concurrent machinery.

#1093 stays open, scoped to the row group and the column. The byte offset is dropped rather than deferred: if it comes back it should come back with a consumer asking for it.

Note for #1098, which is stacked on this branch: its overlay builds its own `ReadContext` in `ParquetModel` for the four regions dive reads itself, so nothing it renders depends on this. It does inherit the `ReadContext` record and the `Region` enum from here, and will need to carry them itself once it rebases onto `main`.

Issue #1095, opened 2026-09-05
Corrupt page data is read back as values when the writer emitted no page CRCs


gunnarmorling:
Corrupt page data is read back as values, with no error and a zero exit code. Flipping eight bytes inside the first data page of `core/src/test/resources/column_index_pushdown.parquet` (the page body, not its header):

```
$ hardwood print -n 3 -f silent.parquet
+---------------------+-------+
| id                  | value |
+---------------------+-------+
| 9223372036854775679 | 1000  |
| 1                   | 1001  |
| 2                   | 1002  |
+---------------------+-------+
$ echo $?
0
```

`9223372036854775679` is presented exactly as the two real values beside it. This is the failure mode CLAUDE.md rules out — "silent failures are never an option" — and it is arguably worse than a crash, because a crash at least says something is wrong.

### Why nothing catches it

**`PLAIN` has no redundancy.** Every bit pattern is a valid `INT64`. Nothing about the corrupted value is detectable from the value alone.

**An RLE/bit-packing run header cannot be invalid.** The header is a ULEB128 varint: bit 0 selects bit-packed from RLE and the rest is a count. Every bit pattern is a syntactically valid header, so corruption changes the mode and the run length rather than producing something rejectable. `RleBitPackingHybridDecoder` has one consistency check:

```java
if (remaining > 0) {
    throw new IllegalStateException("Insufficient RLE/Bit-Packing data: decoded "
            + (count - remaining) + " of " + count + " requested values");
}
```

That fires only when corruption makes a run come out **short**. A run that comes out long is consumed without complaint.

**Decoded levels are never bounds-checked**, and the two predicates in use disagree about an out-of-range one. For `maxDef == 1`, a corrupt level of `5` reads as *null* through `countNonNulls`, which tests `defLevels[i] == maxDef`, and as *present* through `NestedLevelComputer`, which tests `def >= maxDefLevel`. Whichever a given read takes, the other interpretation is also reachable.

### The mechanism that would catch it

Per-page CRCs. The reader already validates them, at three call sites (`PageDecoder`, `DictionaryParser`, `SequentialFetchPlan`), and does so correctly — a deliberately corrupted page in a file that carries CRCs is caught with `CRC mismatch for column id: expected e99fb73a but computed 9b62d6e2`.

But validation is conditional on the writer having emitted one:

```java
if (pageHeader.crc() != null) {
    CrcValidator.assertCorrectCrc(pageHeader.crc(), pageData, column.name());
}
[...]

Issue #1094, opened 2026-09-05
CLI: a failure raised mid-read prints a raw stack trace and leaves truncated output in place


gunnarmorling:
Every command catches the checked `IOException`, but the read path converts to `UncheckedIOException` once reading is under way — deliberately, so the failure can cross the virtual-thread boundary in `ColumnWorker`. Nothing catches that, so any failure raised *during* a read escapes to aesh, which prints a stack trace.

The messages themselves are fine. It is only the presentation that falls off a cliff once the failure happens mid-read rather than at open time:

```
$ hardwood print -n 3 -f nope.parquet
Error reading file: nope.parquet

$ hardwood print -n 3 -c nosuch -f data.parquet
Column not found: nosuch

$ hardwood footer -f encrypted_footer.parquet
Encrypted Parquet files are not supported (Parquet Modular Encryption).

$ hardwood print -n 3 -f crc-bad.parquet
Exception in thread "main" java.lang.RuntimeException: java.lang.RuntimeException: java.io.UncheckedIOException: [crc-bad.parquet] CRC mismatch for column id: expected e99fb73a but computed 9b62d6e2
	at org.aesh.command.impl.AeshCommandRuntime.executeSingle(AeshCommandRuntime.java:284)
	at org.aesh.command.impl.AeshCommandRuntime.runExecutor(AeshCommandRuntime.java:245)
	... 20 more frames
```

All four exit 1. The fourth wraps a perfectly good message — file, column and both checksums — in three layers of `RuntimeException` and twenty frames of aesh internals.

This is not specific to CRC or to `print`. Of the commands that catch `IOException`, **none** handle `UncheckedIOException`:

| Command | `catch (IOException)` | handles `UncheckedIOException` |
|---|---|---|
| `ConvertCommand` | 1 | no |
| `DiveCommand` | 2 | no |
| `FileMixin` | 1 | no |
| `FooterCommand` | 1 | no |
| `InfoCommand` | 1 | no |
| `InspectColumnsCommand` | 2 | no |
| `InspectDictionaryCommand` | 3 | no |
| `InspectPagesCommand` | 5 | no |
| `InspectRowGroupsCommand` | 1 | no |
| `PrintCommand` | 1 | no |
| `SchemaCommand` | 1 | no |

So every mid-read failure — a corrupt page, a truncated chunk, an object-store error part-way through a scan — reports this way.

[...]

PR #1078, opened 2026-09-04
#342 Surface IOExceptions in DataPreviewScreen as an error overlay

muhannd2004:
#342 Surface IOExceptions in DataPreviewScreen as an error overlay 

## Authorship

Hardwood is built with AI, not by AI. LLM assistance is welcome; submitting raw LLM output without understanding is not.
Neither are unattended agents opening pull requests.
See the [contribution guide](https://github.com/hardwood-hq/hardwood/blob/main/CONTRIBUTING.md) for more details.
Check this box to confirm:

- [x] This change is coming from a human who understands what it does and why, and can discuss it in review

## Checklist

- [x] Build via `./mvnw clean verify` passes
- [x] The commit message is prefixed with the issue key ("#123 Adding support for...")
- [x] Code changes are covered by tests
- [ ] If this PR adds or changes user-facing behaviour, `docs/` has been updated

---
## Note regarding local tests: 
<del>
While running the required `./mvnw clean verify` locally, my build failed with the following error:
IllegalState UUID can only annotate FIXED_LEN_BYTE_ARRAY(16)

I tested a clean checkout of the main branch without any of my changes, and the exact same error occurred, so this appears to be a pre-existing issue upstream.

</del>

---

[io_error_test.webm](https://github.com/user-attachments/assets/ec3a549c-9da5-4c0d-8a8a-661b39feeac0)

Issue #342, opened 2026-04-26
dive: surface IOExceptions in DataPreviewScreen.loadPage as an error overlay


gunnarmorling:
\`DataPreviewScreen.loadPage\`'s exception path wraps \`IOException\` in \`UncheckedIOException\` and lets it propagate. The TUI loop will then crash (TamboUI doesn't recover from exceptions in the render callback). For a corrupt-page error mid-session, the UX is \"your dive session dies.\"

**Fix:** catch in \`DiveApp.render\` (or in the screen's render itself) and surface as an error overlay — same shape as the help overlay, with the exception message and a hint to press Esc. The session keeps running and the user can navigate away from the bad state.

gunnarmorling:
Superseded by #1092.

PR #1098, opened 2026-09-05
#1092 Keep the dive session when a screen cannot read what it is showing

### LINKED ISSUE #1092: dive: a file read failure inside the render or event loop kills the session
Every `hardwood dive` screen that reads from the file does so inside the render callback or the key handler. Neither is guarded, so an unreadable region anywhere in the file terminates the session: the exception escapes `DiveApp.render`, TamboUI paints its crash screen over the un-cleared buffer (so the stack trace is interleaved with the previous screen's labels), and `Esc` exits the process because the event loop is already dead.

Six screens reach the file from a render or keybar path:

| Screen | Call |
|---|---|
| `PagesScreen` | `pageHeaders`, `columnIndex`, `offsetIndex` |
| `ColumnIndexScreen` | `columnIndex` |
| `OffsetIndexScreen` | `offsetIndex` |
| `ColumnAcrossRowGroupsScreen` | `offsetIndex` |
| `DictionaryScreen` | `dictionaryChunkBytes`, `loadDictionary` |
| `DataPreviewScreen` | `readPreviewPage` |

**Repro.** Take a file with a page index, clobber one data page header so the footer and the earlier pages stay intact:

```bash
cp core/src/test/resources/column_index_pushdown.parquet /tmp/corrupt-page.parquet
perl -e 'print "\xff" x 24' | dd of=/tmp/corrupt-page.parquet bs=1 seek=41104 conv=notrunc
hardwood dive -f /tmp/corrupt-page.parquet
```

Offset 41104 is where page 5 of the `id` column starts. `hardwood footer` and `hardwood print -n 3` still work. In dive, Schema -> `id` -> Pages kills the session with `UncheckedIOException: java.io.IOException: Unknown field type: 15` thrown from `PagesScreen.keybarKeys` -> `ParquetModel.pageHeaders`.

**Fix.** Handle this centrally rather than per screen: guard the render and key-dispatch paths in `DiveApp` so a read failure becomes an error overlay the reader can navigate away from, and the session survives. #342 asked for the same thing scoped to `DataPreviewScreen` and offered this option ("catch in `DiveApp.render`"); this issue supersedes it.

The overlay says where in the file the read failed. #1093 has shipped, so the reader's own message already names the file, the row group, the column and the page, and the overlay shows it unchanged rather than inventing wording of its own:

```
╭ Read failed ─────────────────────────────────────────────╮
│ [data.parquet: row group 0, column 'id'] PageHeader      │
│ field 15 — Unknown field type: 15                        │
│ [Esc] back                                               │
╰──────────────────────────────────────────────────────────╯
```

No byte offset: it is known only deep inside a parse, and carrying it out was what made #1097 too large to justify. Without one there is nothing to centre a hex window on, so the overlay shows the message alone.

Dive is not the only place that parses the file directly. The page index, the page-header walk and the dictionary page are read by `ParquetModel` and by `InspectPagesCommand` outside the read pipeline, so nothing was placing their failures. Both now do it themselves, which is what puts the file, row group and column into the box above and into `hardwood inspect pages`.

gunnarmorling:
Closes #1092. Supersedes #342 and its PR #1078.

**Stacked on #1097** — target that branch, not `main`. The diff shown here includes #1097's commits until it merges.

Every `hardwood dive` screen that reads from the file did so inside the render callback or the key handler, and neither was guarded. An unreadable region anywhere ended the session: the exception escaped, TamboUI painted its crash screen over a buffer it does not clear — so the stack trace interleaved with the previous screen's labels — and `Esc` exited the process, because there was no longer a loop to press it in. Which screen died depended only on which region was damaged.

```
╭ Read failed ─────────────────────────────────────────────╮
│ [data.parquet] column id, page header (row group 0) at   │
│ byte 41104 (0x00a090) — PageHeader field 15 — Unknown    │
│ field type: 15                                           │
│ 00a088  ff 13 00 00 00 00 00 00  ........                │
│ 00a090  ff ff ff ff ff ff ff ff  ........                │
│ 00a098  ff ff ff ff ff ff ff ff  ........                │
│ [Esc] back                                               │
╰──────────────────────────────────────────────────────────╯
```

## Why it is in `DiveApp` and not in the screens

Six screens reach the file from a render or keybar path — `PagesScreen`, `ColumnIndexScreen`, `OffsetIndexScreen`, `ColumnAcrossRowGroupsScreen`, `DictionaryScreen`, `DataPreviewScreen`. A per-screen error state solves one and leaves the shape of the bug in the other five, and every screen added later has to remember to opt in. **No screen changes in this PR.** `ScreenState` stays a description of what a screen is showing, and the handlers stay pure functions of `(event, model, stack)`.

Keys still reach the screen underneath a failure, which is what lets a reader move off a damaged region rather than only back out of it: the stack holds the last state that rendered, so paging from it loads a different window.

`Esc` is the exception, and a test found it. A screen normally gets first refusal on `Esc` so it can claim it for something of its own, but a screen that cannot read cannot honour that — and its handler reads too, so the one key out of the overlay was failing on its way to the door.

## The bytes

The overlay reads a bounded window at the offset rather than carrying a payload on the exception: a parse that failed over bytes that read cleanly will read them cleanly again, and a read that failed outright has nothing to show and fails here too, so the overlay catches its own read and falls back to the message.

Bytes are shown **only where the offset is the byte the read stopped on**, which in practice means a Thrift structure that would not parse — a page header, column index, offset index or dictionary page header. That line matters twice over:

- Thrift structures are the part of a Parquet file that is never compressed, so the window can never present compressed bytes as though they were readable.
- A failure inside a page body knows the page and nothing finer. Centring a window on the page's first bytes, which are usually fine, under an authoritative-looking gutter would be worse than showing nothing. Those report the region and no bytes.

So this shows bytes for damaged *structure*, not for wrong *data*. Wrong data is the larger class; what this covers is the class a reader has no other tool for while inside dive, since the screens that browse structure are the ones it fires on.

Eight bytes to a row, not sixteen: a sixteen-byte row needs about 78 columns and the overlay is capped at 60, so sixteen clips the ASCII gutter off the right edge. The key/value modal grows to 120 and keeps sixteen. The dump moved to a shared `HexDump` and the Arrow schema dump in `KvMetadataFormatter` is rewritten in terms of it.

Nothing is marked in the gutter: `▶` means "Enter acts on this" everywhere else in dive, so the failing byte and its row's offset carry `Theme.error()` instead.

## Tests

`DiveReadFailureTest` covers a damaged page header, column index, offset index and dictionary page, each deriving its offset from the file's own metadata rather than hardcoding one that stops being checked the moment the fixture changes. Which screen a damaged file takes down depends on which region is damaged, so one region proves nothing about the rest.

`HexDumpTest` asserts the two properties that were broken during development — the window starts on a row and contains the byte it was asked about — across the range rather than at the offsets that happened to be tried. Writing it found a third: trimming the hex column also trimmed the padding standing in for bytes a short final row does not have, so the ASCII column shifted left on the one row a reader is most likely to be reading.

`./mvnw verify` green, including the parquet-testing corpus.

## Not closed by this

- Bytes for a failure inside a page body would need the decompressed window carried on the exception. Deliberately out; the coordinate-space rule says why.
- The offset in a multi-page failure names *a* damaged page rather than deterministically the first: pages decode concurrently and the failure that reports is the one that won the race. Recorded in the design.
- #1094 (the non-interactive CLI prints a stack trace and leaves truncated output) and #1095 (corrupt page data is accepted silently) are untouched and independent.

Design: [_designs/DIVE_READ_FAILURE_HANDLING.md](_designs/DIVE_READ_FAILURE_HANDLING.md), Part 2. Part 1 is #1097.


gunnarmorling:
Reopened as a PR against `main`: GitHub will not retarget the base of a PR that is part of a stack, and the branch this one was stacked on (`1093-read-context-on-exceptions`, PR #1097) is gone. Same branch, rebuilt off main.

Issue #1124, opened 2026-09-07
Nothing in the build runs in parallel


gunnarmorling:
`./mvnw verify -DskipITs` takes 60 seconds on an 8-core machine, and 42 of those seconds are two modules running their tests in a single JVM, one class at a time.

| Module | Time | Of which Surefire |
| --- | --- | --- |
| `parquet-testing-runner` | 31.7s | 30.2s |
| `core` | 12.9s | 12.0s |
| everything else | ~15s | ~5s |

Compilation, Error Prone, the licence check, jar packaging and the enforcer come to under 5 seconds combined. The build is its tests.

There is no Surefire configuration in the parent POM, so it runs at the default `forkCount=1`, and no `-T`, so the reactor is serial as well. Setting both:

| Configuration | Wall clock |
| --- | --- |
| as-is | 59.9s |
| `-T 1C` | 48.1s |
| `-T 1C -DforkCount=4` | 35.9s |

Measured again after touching a `core` source file, so the incremental compiler is not doing the work: 58.9s becomes 44.1s.

Four forks rather than one per core: `parquet-testing-runner` runs 29.9s / 19.6s / 18.6s / 20.3s at one, two, four and eight forks, and `core` 12.9s / 10.0s / 10.5s at one, four and eight. Core barely moves at any setting because its tests already spawn reader worker threads, so a single fork saturates the machine.

Two approaches that measured worse and are not part of this:

- JUnit 5 parallel execution fails 26 `core` tests — the JFR event tests, which share a process-wide recording, and `SequentialFetchPlanChunkSizeTest`, which sets a system property. Scoped to `parquet-testing-runner` alone, on top of four forks, it took the module from 23s to 32s through thread over-subscription.
- `-Dsurefire.runOrder=balanced` took the same module from 23.1s to 26.6s.

CI is unaffected either way: every job builds a single module with `-pl`, so `-T` is a no-op there and the fork count applies to one module at a time.

A floor remains at roughly 16 seconds. `ParquetComparisonTest` is one class holding 15 of `parquet-testing-runner`'s 30 seconds, and no fork count splits a single class. #1123 covers what makes it expensive.

gunnarmorling:
The `-T 1C` added here races the Error Prone checks, and a fresh-clone `./mvnw verify` fails because of it.

[...]

Issue #1107, opened 2026-09-06
Time to first row is proportional to file count in a multi-file read


gunnarmorling:
`ParquetFileReader.openAll(files)` opens one file. `reader.rowReader()` opens the other N-1, before returning a single row.

Files opened before the first row, by dataset size:

| Files | Opened before first row |
|---|---|
| 8 | 8 |
| 32 | 32 |
| 128 | 128 |
| 512 | 512 |

On object storage that is roughly three round trips per footer, so a 512-file dataset spends about 1,500 round trips before the first row — most of them for files the caller may never reach. `head(N)` is already bounded, because the file loop stops when the row budget is spent; a plain `openAll(...).rowReader()` is not, and that is the common shape.

## What is misplaced

Pruning at reader-build time is not the problem: the caller said `.filter(...)`, so building the reader consults bloom filters and dictionaries. Opening *other files* is the part that belongs elsewhere. Neither method's name predicts the split — a caller who says `openAll(files)` assumes it opened the files, and a caller who says `rowReader()` assumes it wired a reader over what is already open.

Planning a file when the read reaches it makes the count constant instead. Files opened by a read abandoned after one row:

| Files | Before | Unfiltered | Filtered |
|---|---|---|---|
| 8 | 8 | 5 | 4 |
| 32 | 32 | 4 | 4 |
| 128 | 128 | 4 | 4 |
| 512 | 512 | 4 | 4 |

A filtered `head(5)` opens 2–3 whatever the dataset size. The remainder is the retriever prefetching until backpressure stops it, which is the prefetch working rather than speculation. Nothing is skipped: reading every row still opens every file.

One thing planning lazily must not lose: `prefetchNextRowGroup` starts the next row group's first chunk read, and it has to keep doing that *across* a file boundary. A work list that stops at the end of the current file leaves nothing for it to look at there, and the boundary becomes the one step of a read that stalls — the retriever plans the next file and only then fetches its first chunk, cold. The next row group therefore has to be asked for in a way that plans the next file when the read is about to cross into it.

## What forces the whole plan

Planning one file at a time is not enough on its own, because two decisions about a filtered read are made once, at construction, from facts that belong to individual row groups inside individual files:

- whether `maxRows` may cap the drain, and
- whether the predicate needs evaluating at all (`RowGroupIterator.isFilterSatisfiedByStatistics`).

[...]

PR #1125, opened 2026-09-07
#1096 Parse a dictionary page's header once, where the page is first met

### LINKED ISSUE #1096: Dictionary page headers are parsed twice, and their CRCs validated twice, on the sequential read path
`SequentialFetchPlan.initialize()` parses the dictionary page header, slices its compressed bytes, validates the CRC, and then hands the whole region to `DictionaryParser.parse`, which does all three again:

```java
ParsedHeader parsed = readPageHeader(position);          // header parsed
PageHeader header = parsed.header();
int headerSize = parsed.headerSize();

if (header.type() == PageType.DICTIONARY_PAGE) {
    int compressedSize = header.compressedPageSize();
    int numValues = header.dictionaryPageHeader().numValues();
    ...
    ByteBuffer dictRegion = readBytes(position, headerSize + compressedSize);
    ByteBuffer compressedData = dictRegion.slice(headerSize, compressedSize);
    if (header.crc() != null) {
        CrcValidator.assertCorrectCrc(header.crc(), compressedData, columnSchema.name());
    }
    dictionary = DictionaryParser.parse(dictRegion, columnSchema, metaData, context);   // <-- here
}
```

`dictRegion` includes the header, so `DictionaryParser.parse` opens a fresh `ThriftCompactReader` at offset 0, re-reads the same header, slices `(headerSize, compressedSize)` to the same bytes, and re-validates the same CRC.

Not a correctness bug: the second computation agrees with the first by construction, which is why it has gone unnoticed. It is duplicated work on the read path.

## What it costs

- **The dictionary page header is parsed twice on every dictionary-encoded column chunk taking the sequential path** — regardless of CRCs, so this is the part that bites on ordinary files.
- **The CRC-32 is computed twice over the whole compressed dictionary**, which for a large dictionary is not trivial — but only on files whose writer emitted page CRCs, and per #1095 almost none do. Small today, larger if #1095 ever leads to CRCs being expected.

There is no third site: `PageDecoder` rejects dictionary pages outright (`default -> throw new IOException("Unexpected page type for single-page decode: " + pageHeader.type())`), so it only ever validates data pages.

## Fix

`SequentialFetchPlan` cannot stop parsing the header — it needs `headerSize` for its position arithmetic and the `numValues` sanity check — so the parse stays there and `DictionaryParser` gains an entry point that takes the already-parsed header and the already-sliced compressed bytes. The existing `parse(ByteBuffer dictRegion, ...)` stays for callers that genuinely start from an unparsed region (the CLI reaches it that way).

Ownership of the CRC check should end up in one place rather than both.

## Acceptance

A before/after benchmark on a dictionary-encoded read showing the duplicated work gone and no regression elsewhere. Deliberately kept out of #1093, which is the change that must prove it does *not* affect read performance; measuring both at once would muddy that.

gunnarmorling:
Closes #1096. Carries a build fix belonging to #1124, which was blocking this branch.

`SequentialFetchPlan.initialize()` has to parse the dictionary page header itself — it needs the header size to know where the page ends and the next one begins. It then handed the whole region, header included, to `DictionaryParser.parse`, which opened a reader at offset 0 and read that same header back out, cut the same body, and checked the same checksum. The second answer agreed with the first by construction, which is why nothing ever caught it.

`DictionaryParser` gains `parsePage`, taking the already-parsed header and the body it delimits. The region-based `parse` stays for callers that genuinely start from an unparsed region — `IndexedFetchPlan`, the dictionary filter source, and both CLI entry points — and now reaches the same code once it has found a dictionary page to hand over.

Ownership of the checksum moves with it, as the issue asked. Repo-wide there are now exactly two `CrcValidator.assertCorrectCrc` call sites: `DictionaryParser` for dictionary pages and `PageDecoder` for data pages. The value count is validated in the same place, so the two entry points cannot drift on what counts as a valid page — and a dictionary page whose header omits `dictionary_page_header`, which used to reach the decoder and dereference null, now fails with a message.

## What it is worth

The end-to-end scan cannot measure this, and it took two wrong answers to establish that. On a fixture built to foreground the dictionary — 100 row groups, each a megabyte of dictionary feeding twelve kilobytes of indices — before and after are indistinguishable: 281.0 ± 20.1 ms against 278.5 ± 18.2 ms. `perfnorm` first appeared to show −74M instructions/op, but the check that mattered killed it: the wide fixture carries four times the dictionary bytes of the narrow one, so a checksum-driven delta has to be about four times larger there, and instead it was smaller (74M against 111M). Repeating with five forks explained why — the error bars are ±120–230M. A read that spans a thread pool and decodes several hundred thousand values per operation, with its GC and JIT threads counted in, has no resolution at this scale.

One page on one thread does. Setting up a single dictionary page, against the decode floor both entry points end in:

| dictionary | decode floor | header handed over | region handed over | duplicated work |
| --- | --- | --- | --- | --- |
| 256 KB | 94.17 ± 0.57 | 94.42 ± 0.82 | 117.49 ± 0.38 | 23.1 µs (+24.5%) |
| 1 MB | 427.56 ± 1.30 | 427.81 ± 1.18 | 514.90 ± 1.56 | 87.1 µs (+20.4%) |

Four times the dictionary bytes, 3.8 times the saving. That linear scaling is what identifies the cost as the checksum rather than the header parse: it is roughly 12 GB/s of redundant CRC-32. Setup overhead above the decode floor falls from 23.3 µs and 87.3 µs to 0.25 µs.

The two measurements agree once the arithmetic is done. A hundred row groups at 87 µs is 8.7 ms of CPU, spread across the four threads the scan keeps busy — about 2 ms of a 280 ms read, or 0.8%, far inside the ±18 ms band. So this is a per-column-chunk win that shows on files with many row groups and large dictionaries, and is invisible on a single-row-group scan.

Measured on an Intel N300, clock pinned at 1.50 GHz (`profiling-setup.sh calibrate 90 --pin`), Temurin 25.

## Benchmarks

Both are self-generating and documented in the micro-benchmarks README.

`DictionaryPageParseBenchmark` is the measurement above: one page, one thread, no file and no pool, with `decodeOnly` as the floor so the two gaps read as fractions of the work that had to happen anyway.

`DictionaryPageSetupBenchmark` puts the same work inside a real scan. It is a regression guard rather than a measurement, and its javadoc says so.

## Testing

`DictionaryParserTest` covers a page parsed from header and body alone — which only works if nothing re-reads a header out of those bytes — and each of the four things the parser refuses: a page of the wrong type, a body that is not the length the header describes, a missing `dictionary_page_header`, and a negative value count. The equivalence between the two entry points is asserted entry by entry rather than by dictionary size, since size is the header's `numValues` on both sides whatever bytes were decoded.

That the body is checked at all matters because a wrong body does not fail on its own: the uncompressed codec copies whatever remains of the buffer, so a region handed over whole decodes its own header bytes as dictionary entries, and the checksum that would catch it is one almost no writer emits.

`CrcValidationTest.testCorruptedDictionaryPageDetected` now corrupts the last byte before the first data page. It was corrupting the last byte of the column chunk, which is a data page byte — so it went through `PageDecoder`, duplicated `testCorruptedDataDetected`, and left the dictionary path untested on the failing side. Its fixture carries no page index, so it runs through `SequentialFetchPlan`, the path this changes.

Full `./mvnw verify`, integration tests included, is green.

## Notes for review

`parsePage` is `public` only so the benchmark module can call it; every caller inside `core` is same-package. Package-private would be the tighter choice, at the cost of the isolated measurement.

The indexed path's dictionary checksum has no corrupt-fixture test, because the only dictionary-CRC fixture in the repo lacks a page index. That predates this change and the code path is now shared, so I have left it alone.

## The build fix (#1124)

The first CI run of this branch failed in `hardwood-test-support`, resolving `hardwood-error-prone-checks` for its annotationProcessorPath, while the same reactor reported the checks built four milliseconds earlier. That is #1124's `-T 1C` against an ordering the reactor cannot derive: a processor path is compiler-plugin configuration, not a declared dependency. Only a local repository without the artifact exposes it, so a fresh clone or a cold CI cache hits it every time and everyone else never does — `./mvnw verify` on a clean checkout fails on current main.

The nine compiling modules now declare the checks as a `test`-scoped dependency, which is an edge the reactor reads. The parent cannot carry it: the checks module would depend on itself, which Maven rejects whatever the scope.

A `[docs]` commit also rewords CLAUDE.md's worktree rule, which said "before any branch work" and so had sessions building a worktree to read a diff, stated the reference checkout's branch as a property of the directory rather than a default for the session, and never said when a worktree stops being useful.

Also here: `MultiFilePlanningTest.aReadThatStopsEarlyDoesNotOpenEveryFile` failed once and passed on a re-run of the same tree. It read one row from an uncapped reader, abandoned the loop, and asserted fewer than all eight files had been touched — the shape `anUncappedFilteredReadYieldsWithoutPlanningTheWholeRead` warns against two tests below, since walking away leaves the retriever running. Stalling the calling thread for five seconds shows what was measured: uncapped it reaches 8 of 8 files and fails, and with `head(1)` it settles at 2 and stays there. The cap is the stop the surrounding javadoc already credits the test with, so it now has one. That commit is attributed to #1107.

Verified by deleting `~/.m2/repository/dev/hardwood/hardwood-error-prone-checks` — before the fix the `build-core` command failed 3/3 and `-T 1` passed; after it, both that command and a full `./mvnw verify` succeed from cold. A deliberate `var` still trips `NoVar`, so the checks are still running rather than quietly skipped.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -9,7 +9,9 @@ When a CLI tool is missing from the dev container (e.g. `cmp: command not found`
 
 Several Claude sessions may be running against this repository at once. Git allows one HEAD per working tree, so two sessions sharing a checkout also share HEAD, the index and the files: a `git checkout` in one silently relocates the other, and work gets committed to the wrong branch.
 
-Before any branch work, move into your own worktree. Treat `/workspace` as a reference checkout that stays on `main` — it is where every session lands by default, which is exactly why it must not be where any session commits. Name the worktree directory after its branch, so `git worktree list` reads as a registry of what is in flight.
+Before committing to a branch, move into your own worktree. Reading needs no worktree — review a PR or a branch from wherever you already are, with `gh pr diff`, or `git fetch` and then `git log` / `git diff` / `git range-diff` against `FETCH_HEAD`. None of those move HEAD. Treat `/workspace` as a reference checkout that stays on `main`, since it is where every session lands by default — but that is a default for you to hold to, not a rule to hold against the user. Asked for a branch or a PR in the workspace itself, put it there, and say once what it means for the other sessions. Name the worktree directory after its branch, so `git worktree list` reads as a registry of what is in flight.
+
+Remove your worktree once its work is pushed. Never leave one behind for a review that is over — a list that keeps merged branches and finished reviews is not a registry of anything.
 
 Stage explicit paths. Never `git add -A` or `git commit -a`: another session's uncommitted files may be sitting in the tree, and a catch-all stage sweeps them into your commit.
 
--- a/avro/pom.xml
+++ b/avro/pom.xml
@@ -54,6 +54,17 @@
   </dependencyManagement>
 
   <dependencies>
+    <!-- Not used by any code here. Declared so the reactor builds the Error Prone checks
+         before this module compiles: they reach javac through the compiler plugin's
+         annotationProcessorPaths, which the reactor does not read when ordering modules,
+         so under `-T` there is otherwise nothing stopping the two running at once. -->
+    <dependency>
+      <groupId>dev.hardwood</groupId>
+      <artifactId>hardwood-error-prone-checks</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
+
     <dependency>
       <groupId>dev.hardwood</groupId>
       <artifactId>hardwood-core</artifactId>
--- a/aws-auth/pom.xml
+++ b/aws-auth/pom.xml
@@ -61,6 +61,17 @@
   </dependencyManagement>
 
   <dependencies>
+    <!-- Not used by any code here. Declared so the reactor builds the Error Prone checks
+         before this module compiles: they reach javac through the compiler plugin's
+         annotationProcessorPaths, which the reactor does not read when ordering modules,
+         so under `-T` there is otherwise nothing stopping the two running at once. -->
+    <dependency>
+      <groupId>dev.hardwood</groupId>
+      <artifactId>hardwood-error-prone-checks</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
+
     <dependency>
       <groupId>dev.hardwood</groupId>
       <artifactId>hardwood-s3</artifactId>
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -121,6 +121,17 @@
   </dependencyManagement>
 
   <dependencies>
+    <!-- Not used by any code here. Declared so the reactor builds the Error Prone checks
+         before this module compiles: they reach javac through the compiler plugin's
+         annotationProcessorPaths, which the reactor does not read when ordering modules,
+         so under `-T` there is otherwise nothing stopping the two running at once. -->
+    <dependency>
+      <groupId>dev.hardwood</groupId>
+      <artifactId>hardwood-error-prone-checks</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
+
     <!-- Hardwood Core -->
     <dependency>
       <groupId>dev.hardwood</groupId>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -50,6 +50,17 @@
   </dependencyManagement>
 
   <dependencies>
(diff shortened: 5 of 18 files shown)


gunnarmorling:
Closing and reopening to re-trigger CI; the failure was a timing-dependent assertion, not a code change.

PR #1145, opened 2026-09-08
#1137 Decline to prune on an unrecognized bloom filter variant instea…

### LINKED ISSUE #1137: A bloom filter variant we cannot evaluate should skip pruning, not fail the read
A bloom filter whose header declares an algorithm, hash or compression this library does not implement fails the whole read. It should cost one pruning opportunity instead.

`BloomFilterHeaderReader` rejects any union variant it does not recognise:

```java
default -> throw notAVariantOf(ThriftStruct.BLOOM_FILTER_ALGORITHM, variant, "bloom filter algorithm");
```

The check itself is load-bearing and must stay. Nothing downstream consults `algorithm()`, `hash()` or `compression()` — `BloomFilterReader.readBitset` uses only `numBytes()`, and `BloomFilter.mightContain` goes straight to `SplitBlockBloomFilter` with XXH64 at seed 0. So a file declaring some future hash, read as if it were XXH64, would probe a filter built with something else and answer "definitely absent" for values that are present: dropped row groups and missing rows, with no error. Proceeding is not an option.

Two things are wrong with how it fails.

**It is the wrong exception.** `ParquetReadException` says the file is wrong. A bloom filter using a hash a later Parquet version defines is a correct file this library cannot evaluate, which [_designs/EXCEPTION_MODEL.md](_designs/EXCEPTION_MODEL.md) places under `UnsupportedOperationException`, alongside an encoding that is not implemented. The remedy is a newer Hardwood, not a different file.

**It should not fail at all.** `RowGroupBloomFilterSource` already has the answer for a filter it cannot use, and applies it three times — decline to prune, keep the row group, warn:

> The offset is present but points at or before the file's magic header, so it cannot name a real filter. Treat it as corruption but stay conservative — decline to prune rather than throw, keeping the row group (statistics still apply) — and warn so the malformed footer is visible instead of silently reducing pruning.

An unrecognised variant is the same situation: the filter cannot be used, the statistics still apply, and the row group is still readable. Failing the read is the one outcome that serves nobody, since every row the filter would have excluded is still obtainable by reading them.

Changing the exception type alone does not get there — nothing on that path catches `UnsupportedOperationException` (the one `catch` in the class is `requireSameFile`, which rethrows). It needs the decline-to-prune arm the invalid-offset case above it already has.

Found while making the Thrift readers consistent in #1136; separate because it changes behaviour rather than wording — a file that fails today would start reading.

nitrogen404:
Closes #1137

A bloom filter header naming an algorithm, hash, or compression variant this version of Hardwood doesn’t implement resulted in a complete read failure with ParquetReadException, even though such a file is perfectly valid Parquet — simply utilizing a feature this library predates. Since nothing downstream reads algorithm()/hash()/compression() after parsing the header, the fix raises UnsupportedOperationException instead (as per _designs/EXCEPTION_MODEL.md), and RowGroupBloomFilterSource catches it, refusing to prune that column while preserving the row group — mirroring the existing fallback for an invalid bloom_filter_offset.

- BloomFilterHeaderReader: unrecognized union variant → UnsupportedOperationException
- RowGroupBloomFilterSource: catches it, warns, returns no filter (statistics still apply)
- _designs/EXCEPTION_MODEL.md: documents the carve-out
- Tests: ThriftFieldNamingTest updated for the new exception type; new BloomFilterPushDownTest case reproduces the bug end-to-end

./mvnw verify -DskipIT's green.


gunnarmorling:
Thanks a lot, @nitrogen404! Expanded a bit, merging now.

Issue #1100, opened 2026-09-06
dive: show the current row group in the Data preview


gunnarmorling:
The Data preview title reports the absolute row range and the visible column window:

```
 Data preview (rows 1–38 of 1,000,000 · cols 1–7 of 19)
```

What it does not report is which row group the visible rows come from. That is the one piece of structural context the screen is missing: everywhere else in dive (Row groups, Column chunks, Pages, Column index) the row group is the organising unit, and when you page through the data you lose track of where you are relative to it. It also hides the boundaries — a page of rows can straddle two row groups and nothing on screen says so, even though that boundary is exactly what explains a jump in decode latency, a change in encoding, or a discontinuity in the values of a clustered column.

## Proposal

Extend the title with the row group the cursor is in, e.g.

```
 Data preview (rows 1–38 of 1,000,000 · RG 1 of 24 · cols 1–7 of 19)
```

Details to settle during implementation:

- The row group should follow the **cursor** (`firstRow + selectedRow`), not the first visible row, so that stepping across a boundary updates it.
- When the visible window spans more than one row group, either render a range (`RG 3–4 of 24`) or keep the cursor's row group and mark the boundary in the table body — a separator row, or a marker in the row-index gutter. Marking the boundary in the body is the more informative of the two and works for the single-row-group case as well.
- Mapping an absolute row to a row-group index is a cumulative walk over `RowGroup.numRows()`; `ParquetModel` should expose it once (`rowGroupOf(long row)`) rather than have the screen do it, since #1101 needs the inverse mapping.

## Files

- `cli/src/main/java/dev/hardwood/cli/dive/internal/DataPreviewScreen.java` (title, around the `Fmt.fmt(" Data preview (…) ")` call)
- `cli/src/main/java/dev/hardwood/cli/dive/ParquetModel.java` (row → row-group lookup)

SurajPatelPro:
I'd like to take this one — will add `ParquetModel.rowGroupOf(long row)` and extend the Data preview title with the cursor's row group, marking the boundary in the table body when the visible window spans more than one row group.

gunnarmorling:
Proposal to add to this issue's scope: `]` / `[` in the Data preview, moving the cursor to the first row of the next / previous row group.

[...]

PR #1185, opened 2026-09-11
1101 jump to row or row group

### LINKED ISSUE #1101: dive: jump to a given row group or record in the Data preview
Navigation in the Data preview is relative only: `↑/↓` step, `PgUp/PgDn` page, `g`/`G` jump to the first/last row. There is no way to say "take me to row 4,000,000" or "take me to the start of row group 17". On a file with millions of rows the only route to a specific record is holding `PgDn`, and reaching a specific row group means counting `numRows()` by hand off the Row groups screen.

This matters for the debugging workflow dive exists for: you look at a row group in the Row groups screen, see suspicious statistics or an odd encoding, and then want to look at the actual values behind it. Today that path does not exist.

## Proposal

Two entry points, both landing in the Data preview:

1. **Jump within the Data preview.** A key (`:` or `/`, to be picked) opens a small prompt overlay accepting either a row number or a row group (e.g. `12345` for the record, `rg 17` for the row group). Enter seeks the preview so that the target is the selected row; Esc cancels. Out-of-range input is rejected in the prompt rather than silently clamped.
2. **Jump from the Row groups screen.** On `RowGroupsScreen` (and `RowGroupDetailScreen`), a key that opens the Data preview positioned at the first row of the selected row group. This is the common case and needs no typing at all.

The prompt overlay should follow the existing overlay pattern (`HelpOverlay`, `ReadFailureOverlay`) and the navigation rules in `_designs/DIVE_NAVIGATION_MODEL.md`.

## Seek cost

`ParquetModel.readPreviewPage` already builds a fresh bounded cursor per call with `skip(firstRow).head(pageSize)`, and #361 made the underlying reader row-group-aware, so a jump costs one row group's worth of projected-column bytes rather than a walk from row 0. The `PreviewWindow` buffer must be invalidated on a jump — it is a sliding ±10×viewport window around the previous position and holds nothing useful after an arbitrary seek.

Row group → first row is the cumulative `RowGroup.numRows()` walk; it is the inverse of the row → row group mapping needed by #1100, so both should live on `ParquetModel` as one pair of methods.

## CLI parity

`hardwood print` has the same gap: `--rows/-n` only heads or tails, so there is no way to print rows starting at an offset or to print the rows of one row group. The same navigation vocabulary should reach the non-interactive side — `--skip <n>` and `--row-group <i>` on `print` (and `convert`) — so that what you find in dive is reproducible on the command line. Documented under `docs/content/reference/` alongside the existing `print` options.

## Files

- `cli/src/main/java/dev/hardwood/cli/dive/internal/DataPreviewScreen.java`
- `cli/src/main/java/dev/hardwood/cli/dive/internal/RowGroupsScreen.java`, `RowGroupDetailScreen.java`
- `cli/src/main/java/dev/hardwood/cli/dive/internal/PreviewWindow.java` (invalidation)
- `cli/src/main/java/dev/hardwood/cli/dive/ParquetModel.java`
- `cli/src/main/java/dev/hardwood/cli/command/PrintCommand.java`, `ConvertCommand.java`

nitrogen404:
Closes #1101.

Navigation in the dive Data preview was relative only: reaching a known record meant holding PgDn, and reaching a row group's first record meant adding up numRows() by hand. This adds three ways in, and makes the position reproducible outside the TUI.

- : in the Data preview opens a prompt taking either a row (12345) or a row group (rg 17). Enter seeks so the target is the selected row, Esc leaves the view where it was.
- d on the Row groups and Row group detail screens opens the Data preview at that group's first row, no typing. Enter keeps its existing meaning, so the ▶ marker still marks what Enter acts on.
- --skip <n> and --row-group <i> on print and convert, so a position found in dive can be passed on the command line. print --row-index now numbers rows by their position in the file, so the number carries back.

ParquetModel gains the row ↔ row group mapping (firstRowOf, rowGroupOf) over one prefix array of cumulative row counts, which #1100 can reuse for the preview's row-group indicator. Design doc: _designs/DIVE_JUMP_TO_ROW.md.

Assumptions worth confirming:

1. Zero-based everywhere, following inspect columns --row-group 0 and print --row-index. The preview's title still counts from one (rows 1-20 of 300), so a number copied off the title lands one row late — left to #1100, which revisits that line anyway.
2. : rather than /, since / is inline filter on the Schema, Column index and Dictionary screens; / stays free for a value search over the preview.
3. PreviewWindow needed no invalidation path, contrary to the issue text: its buffer is keyed by an absolute row range and is fully replaced whenever the requested page falls outside it, so an arbitrary seek either lands in a buffer still correct for it or triggers a refill.
4. Nothing is clamped. A row or row group past the end, --skip past the end, and d on an empty row group are all refused with a message — an empty row group starts where its successor does, so opening it would show the next group's rows under the wrong number.
5. --skip/--row-group combine with neither each other nor a negative -n, which counts from the end of the file and so has no starting point to move. With --row-group, a positive -n narrows within the group and never crosses into the next.

### DIFF
--- a/_designs/DIVE_JUMP_TO_ROW.md
+++ b/_designs/DIVE_JUMP_TO_ROW.md
@@ -1,0 +1,103 @@
+# Design: jumping to a row or row group in dive
+
+**Status: Implemented.** Tracking issue: #1101. Related: #1100.
+
+## Goal
+
+A reader looking at row group 17 on the Row groups screen — suspicious
+statistics, an odd encoding — should be able to see the values behind it, and a
+reader who knows the record they want should be able to name it. Navigation in
+the Data preview is relative only, so today the route to a specific record is
+holding `PgDn`, and the route to a row group's first record is adding up
+`numRows()` by hand.
+
+The same vocabulary reaches the non-interactive side, so that what a reader
+finds in dive is reproducible on the command line.
+
+## Row ↔ row group mapping
+
+`ParquetModel` owns the mapping in both directions, over one prefix array of
+cumulative `RowGroup.numRows()` built when the model opens:
+
+- `long firstRowOf(int rowGroupIndex)` — the absolute index of a row group's
+  first row. Drives the jump.
+- `int rowGroupOf(long row)` — the row group an absolute row falls in, by binary
+  search over the same array. Drives the Data preview's row-group indicator
+  (#1100), which asks once per keystroke.
+
+Both reject an out-of-range argument rather than clamping: a jump that silently
+lands somewhere else is worse than one that does not happen.
+
+## Jumping within the Data preview
+
+`:` opens a prompt on the Data preview. It accepts either a record — `12345` —
+or a row group — `rg 17`. `Enter` seeks so that the target is the selected row,
+`Esc` closes the prompt and leaves the view where it was.
+
+`:` rather than `/`, because `/` is inline search on the Schema, Column index
+and Dictionary screens, and a Data preview that answers `/` with a jump prompt
+would make the key mean two different things one screen apart. `/` stays
+reserved for a value search over the preview.
+
+Both numbers are counted from zero, as the rest of the tool counts them:
+`hardwood inspect columns --row-group 0` names the first row group, and
+`hardwood print --row-index` numbers the first record 0. A row typed into the
+prompt is therefore the row `--skip` takes, which is what makes a position
+found in dive reproducible on the command line.
+
+The Data preview's title counts from one — `rows 1-20 of 300` — so a reader
+copying a number off the title lands one row past the record they pointed at.
+Aligning the title on zero is a change to what every dive reader already sees
+and is left to #1100, which revisits that line to name the current row group.
+
+Input out of range for the file — a row past the last row, a row group past the
+last row group — is reported in the prompt, which stays open with the text
+intact. Nothing is clamped and no seek happens.
+
+The prompt is a single line above the table, in the shape `DictionaryScreen`
+already uses for its filter, rather than a modal box. Typed input in dive is a
+bar; a modal is for content a reader scrolls (`HelpOverlay`) or dismisses
+(`ReadFailureOverlay`).
+
+The seek itself goes through the Data preview's existing absolute-position move,
+so a jump is the same operation as `G` with a different target, and the
+navigation model's two rules (`_designs/DIVE_NAVIGATION_MODEL.md`) hold
+unchanged: the prompt is a mode, not a new kind of cursor.
+
+`PreviewWindow` needs no invalidation path. Its buffer covers a range of
+absolute row indices and is fully replaced whenever the requested page falls
+outside that range, so an arbitrary seek either lands in a buffer that is still
+correct for it or triggers a refill.
+
+## Jumping from the Row groups screens
+
+`d` on `RowGroupsScreen` and `RowGroupDetailScreen` opens the Data preview at
+the first row of the selected row group. This is the common case and takes no
+typing.
+
+`Enter` on those screens keeps its meaning — drilling into the row group's
+detail and its column chunks — so the `▶` marker continues to mark what `Enter`
+acts on.
+
+## Command-line parity
+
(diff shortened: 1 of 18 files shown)


gunnarmorling:
Hey @nitrogen404, thanks a lot for picking this up! Reviewed this with the help of Claude and have come to realize that jump-to-row should be a general feature for all list screens, not just data preview. Specifically, I'd suggest to lay out the foundation for this in a general way in this PR, wiring it up for data preview only. In a follow-up, we could then also bring this to other screens. Please see below for this and a few other findings.

_Via Claude_

The `d` key, the prefix-sum mapping on `ParquetModel` and the CLI parity all work well, and the screen-level tests are thorough. Before merge there are two bugs to fix and one change to how the prompt is built.

### Design: make the prompt screen-independent

The prompt's jump is the same operation as `G` with a different target, and every list screen shares `G` through `CursorPane.select`. Row groups, Column chunks, Pages, Offset index, Column index, Dictionary and Column across row groups all show a zero-based `#` and can run to thousands of entries, so they will want `:` too. So will Schema: it labels each leaf `[col N]`, and on a schema with thousands of columns `:N` should open up and select leaf column N. A prompt built into `DataPreviewScreen` would have to be copied onto each of them.

Please move the prompt into `DiveApp` as an overlay, next to the help and read-failure overlays. Each screen supplies a resolver that turns the typed number into a new screen state or a refusal. A position in the list alone is not enough, because Schema has to expand the leaf's parent groups, and under a `/` filter the target may be hidden. For this PR, wire it to the Data preview and the Row groups screen. The remaining screens, and `d` on Pages / Offset index via the page's first row, can follow in a separate issue.

With `:` on the Row groups screen, the `rg N` form can go. The prompt then means "go to the item this screen numbers N" everywhere, and only needs to accept digits. A row group's rows are `:N` then `d` from the Row groups screen, and `--row-group` covers the command line.

Moving the prompt into `DiveApp` also fixes the first bug below.

### Bugs

1. **Typing `q` in the prompt quits dive.** `DiveApp.dispatchKey` handles `q`/`Q`, `o` and `?` before the screen sees them, because `isTopInInputMode()` doesn't know about the prompt. The tests call `DataPreviewScreen.handle` directly, so they miss it. Please add a test that types into the prompt through `dispatchKey`.
2. **`--row-group` on an empty row group prints the rest of the file.** `resolveWindow` returns `RowWindow(firstRow, 0)`, and `buildRowReader` reads a limit of `0` as "no limit". Please refuse an empty row group, as `d` does, and make `RowWindow` able to tell "no rows" apart from "no limit". `resolveWindow` takes a `FileMetaData`, so this can be unit-tested with a synthetic one.

### Other changes

3. **Row index with a tail.** `print --row-index -n -5` still numbers the rows from 0. The new SKILL.md text says `--row-index` numbers rows by their position in the file, so please make that hold for a negative `-n` as well (start at `numRows - n`), and drop "keeps" from the sentence.
4. **Preview title.** The prompt and `--skip` count from zero, but the title reads `rows 1–20 of 300`, so a number read off the title lands one row late. Please make the title zero-based in this PR.
5. **Docs.** The dive Keys table in `docs/content/reference/cli.md` needs rows for `:` and `d`, and the Data preview entry under "Available screens" should mention the jump.
6. **`Keys.java`.** `isOpenDataPreview` was inserted below the `g` doc comment. That comment now documents `isOpenDataPreview`, and `isJumpTop` has none.
7. **Design doc.** `_designs/DIVE_JUMP_TO_ROW.md` says `--row-group` is resolved through `firstRowOf`, but `RowLimits.resolveWindow` sums `numRows()` itself. Please bring the doc in line with the code and drop the process wording ("today the route…", "left to #1100"). It will also need updating for the prompt changes above.
8. **Tests.** Please assert the CLI error messages in full with `isEqualTo`, as the neighbouring tests do (`rejectsNegativeMaxWidth`). `skipAndRowGroupAreRefusedTogether` currently checks only the first half of its message. Please also cover `--row-group -1`, and `--row-group` combined with a negative `-n`.
9. **History.** Please fold d7357722 into 76933ec1. Its message describes an interim state of this branch.

Issue #859, opened 2026-08-08
Dictionary-space predicate evaluation: evaluate the filter once per dictionary, answer rows by index lookup


gunnarmorling:
## Motivation

A single dictionary lookup answers a predicate for a whole column chunk. Take `foo = "abcdef"` against a chunk whose dictionary maps `"abcdef" → 42`: scan the dictionary once for the encoded needle, and one of two things happens.

- **Not found** — no row in the chunk can match, so the chunk is skipped without decoding a page. That is #196.
- **Found at 42** — rewrite the predicate to `index == 42` and evaluate it against the raw dictionary-index stream. No value is ever decoded or compared. That is this issue.

One pass over the dictionary, two payoffs. #196 tracks the skip; this tracks the rewrite.

The gain is that per-row work collapses from a value comparison to an integer comparison. The batch matchers (`ColumnBatchMatcher` and its typed subclasses) currently compare every row against the predicate operand, but a dictionary-encoded page holds at most `dictionarySize` distinct values and usually far fewer than it holds rows. For the classic case — a low-cardinality string column with a few dozen distinct values across a million-row page — this turns a million UTF-8 byte comparisons into a few dozen comparisons plus a million integer compares.

`index == 42` is the single-match EQ fast path. The general form is a mask over dictionary IDs: `IN` over k values leaves k surviving indices, and a range predicate leaves an arbitrary subset — dictionary IDs follow the writer's insertion order and the format guarantees no sorting, so `foo < "m"` cannot become an index range.

DuckDB, Velox and arrow-rs/DataFusion all do this. #500 counts "encoded-space eval" among the axes where Hardwood is behind the best columnar readers.

## Proposal

When a page is dictionary-encoded and the resolved predicate targets that column:

1. Evaluate the predicate against the `Dictionary`'s entries once, producing a `dictionarySize`-wide mask of which dictionary IDs satisfy it.
2. Drive the row-level match from the dictionary index stream: the answer for a row is `mask.get(index)` — no value comparison, and the value need not be materialized at all. Specialize the single-set-bit case (EQ with one match) to a plain `index == id` compare.
3. Fall back to the existing per-value matcher for non-dictionary pages, since encoding is per page and a chunk may mix dictionary and plain pages.

## Relationships

[...]

Issue #485, opened 2026-05-16
Drain-side filtering: close remaining gaps so RecordFilterCompiler can be deleted


iifawzi:
## Background

`BatchFilterCompiler` (drain-side, per-column matchers in `FlatColumnWorker`) is a strict subset of what `RecordFilterCompiler` (consumer-side, per-row) handles. Today any predicate that `BatchFilterCompiler.tryCompile` cannot accept falls through to the record-filter path, so `RecordFilterCompiler` plus `FilteredRowReader` must remain in the tree to cover those cases.

The goal of this umbrella issue is to enumerate every remaining shape that forces consumer-side evaluation. Once drain-side covers them, the consumer-side filter path (and the duplication between the two compilers) can be deleted outright.

`Or` (and mixed `And`/`Or`) is already tracked in [#457](https://github.com/hardwood-hq/hardwood/issues/457) and is not repeated below.

### Leaf-type gaps (rejected by `BatchFilterCompiler.isSupported`)

`RecordFilterCompiler` handles these; `BatchFilterCompiler` does not. Each needs a per-column matcher and a `compileLeaf` arm.

- [ ] **`BinaryPredicate`** — string / `BYTE_ARRAY` / `FIXED_LEN_BYTE_ARRAY` comparisons for `EQ`, `NOT_EQ`, `LT`, `LT_EQ`, `GT`, `GT_EQ`. This is the highest-impact gap in practice — most real-world filters are string equality / range (`name = 'x'`, `tag IN (...)`, prefix-style ranges).
- [ ] **`BinaryInPredicate`** — `IN (...)` for strings / byte arrays. Mirrors the existing `IntInBatchMatcher` / `LongInBatchMatcher`.
- [ ] **`Float16Predicate`** — half-float comparisons (decoded as `float` for both values and stats). Small surface but trivial to add alongside `FloatPredicate`.
- [ ] **`GeospatialPredicate`** — spatial filters. `RecordFilterCompiler` currently treats this as `row -> true` (pass-through), so the bar for drain is the same; needs a no-op or proper-eval matcher to keep the consumer fallback off the hot path.

### Structural gaps (rejected before `isSupported` runs)

- [ ] **Nested-column leaves** — `BatchFilterCompiler.java:91` requires `isTopLevel(schema, fileIdx)`. Any predicate on a struct member or list element falls through. Lifting this requires the per-column matcher pipeline to operate on nested column workers, not just `FlatColumnWorker`.
[...]

Issue #454, opened 2026-05-11
Fused range matchers for same-column value ranges in drain-side filtering


iifawzi:
## Background

`BatchFilterCompiler` now supports multiple leaves on the same column by folding them into an `AndBatchMatcher` composite (introduced alongside the drain-side filtering work — see `_designs/DRAIN_SIDE_RECORD_FILTERING.md` and #250). This closes the eligibility gap for range-shaped predicates like `id >= x AND id <= y` so they take the drain-side path rather than falling back to `FilteredRowReader`.

## Tradeoff

The composite is type-agnostic: it runs each child's `test()` separately (one writing into `outWords`, the other into a per-instance scratch buffer) and then word-ANDs the active range. For value-range predicates on a single column this is functionally correct but does roughly **twice the work of a fused matcher** — two full passes over the batch's typed value array plus a merge pass, instead of a single pass with an inline `lo <= v && v <= hi` test.

## Proposed work

Introduce fused per-type range matchers (e.g. `LongRangeBatchMatcher`, `IntRangeBatchMatcher`, `DoubleRangeBatchMatcher`, `FloatRangeBatchMatcher`) that take `(low, lowInclusive, high, highInclusive)` and pack one bit per row in a single loop. Teach `BatchFilterCompiler` to recognise the pattern \"two comparisons on the same column, same primitive type, complementary directions\" before falling back to `AndBatchMatcher`.

Scope:
- Six combinations of inclusive/exclusive bounds per primitive type. Hand-written like the existing per-`(type, op)` matchers, or one matcher per type with the inclusivity carried as a final enum field (see the A/B discussion around shared scalar helpers for the tradeoff between fewer classes and vectorization risk).
- The pattern recognition can be cheap: when about to compose into `AndBatchMatcher`, check if both children are typed comparison matchers on the same primitive that bound the value from opposite sides.

## When to do this

Driven by perf data. The composite already gets `range` predicates onto the drain-side path; the question is how much extra throughput a fused matcher buys on real workloads. Rerun `RecordFilterBenchmarkTest` after the composite change lands, look at the `range` scenario, and use that delta to decide whether the additional matcher classes are worth the maintenance cost.

## Related

- #250 — drain-side record filtering, the umbrella issue
- `_designs/DRAIN_SIDE_RECORD_FILTERING.md` — eligibility and performance discussion

PR #1210, opened 2026-09-13
#485 Filter byte-array leaves on the drain side

iifawzi:
## Authorship

Check this box to confirm:

## Checklist

- [x] If this PR adds or changes user-facing behaviour, `docs/` has been updated — no user-facing API or behaviour change; filters return the same rows, only faster

## Summary

Part of #485. Closes the first two leaf-type gaps listed there — `BinaryPredicate` and `BinaryInPredicate` — so string and decimal filters run on the drain side instead of sending the whole filter back to per-row evaluation on the consumer thread.

### What changes

- **Byte-array matchers.** `Binary{Eq,NotEq,Lt,LtEq,Gt,GtEq,In}BatchMatcher` compare each value's `bytes[offsets[i], offsets[i + 1])` slice in place through new slice overloads on `BinaryComparator`, with no per-row `byte[]`, skipping null rows.
- **Comparison modes.** The literal's `Comparison` decides order and equality: `BYTE_STRING` (unsigned, byte equality), `FIXED_DECIMAL` (signed over equal widths, byte equality), `VARIABLE_DECIMAL` (signed with sign extension; equality compares by value, since a `BYTE_ARRAY` decimal can spell the same number with padding — `0x7F` = `0x00 0x7F`).
- **Short-value equality.** For byte-exact `EQ` / `NOT_EQ` / `IN` with a literal of at most eight bytes, `BinaryShortInBatchMatcher` compares each row as one masked big-endian `long` — branch-free length and value checks against every member, with a byte-wise fallback within eight bytes of the array's end. Below eight bytes `Arrays.equals` compares byte by byte, which otherwise made drain-side `IN` over short strings slower than the consumer path. It is a separate class so the byte-wise matchers keep their C2 profile: sharing a class made their per-row comparison look cold and stop being inlined (confirmed with `PrintInlining`).
- **Physical-type guard.** Binary leaves are drain-eligible only on `BYTE_ARRAY` / `FIXED_LEN_BYTE_ARRAY`; `BatchExchange` hands INT96 the same `BinaryBatchValues`, so without the guard a binary leaf there would compare timestamp bytes.

### Performance

`RecordFilterBenchmarkTest`, 10M rows, alternating main/branch runs (two rounds × 5 iterations). Each contender reads a fixed base projection (`id, value, tag, flag`) plus its filter's columns. Row counts identical on every contender.

| Contender | main (ms) | this PR (ms) | Speedup |
|---|---:|---:|---:|
| `category = 'cat_42'` | 179.0 | 76.3 | 2.34x |
| `category < 'cat_50'` | 242.9 | 152.6 | 1.59x |
| `value < 500 AND category < 'cat_50'` | 241.7 | 150.9 | 1.60x |
| `category IN` (5 values) | 372.1 | 99.1 | 3.76x |
| `category NOT IN` (3 values) | 539.2 | 121.5 | 4.44x |
| 32-byte string eq / IN / NOT IN (`url_tail`) | 196.6 / 412.3 / 551.3 | 123.2 / 354.1 / 258.2 | 1.60x / 1.16x / 2.13x |
| 32-byte string eq / IN / NOT IN (`url_head`) | 187.1 / 351.8 / 515.7 | 106.2 / 256.6 / 221.5 | 1.76x / 1.37x / 2.33x |
| `DECIMAL` on `FIXED_LEN_BYTE_ARRAY(8)` range / eq / IN | 103.7 / 11.2 / 32.1 | 55.5 / 5.0 / 9.9 | 1.87x / 2.25x / 3.23x |
| `DECIMAL` on `BYTE_ARRAY` range / eq / IN | 98.9 / 10.8 / 24.1 | 46.7 / 7.4 / 22.4 | 2.12x / 1.45x / 1.08x |

Numeric contenders are within 5% of main (unchanged code); the no-filter baseline is identical (38.0 ms).

### Notes

- Out of scope and still open under #485: dictionary-space evaluation (#859), `Float16Predicate`, `GeospatialPredicate`, nested leaves, non-projected predicate columns.
- `NOT IN` lowers to a conjunction of `NOT_EQ` leaves, so each member is a separate pass chained through `AndBatchMatcher` — still 4.4x faster than main. This is the same per-leaf composite cost #454 tracks for same-column ranges; fusing a `NOT IN` into one negated membership pass is the byte-array counterpart of that work.
- `BYTE_ARRAY` decimal `IN` (value equality, no short-value path) is at rough parity with main (1.08x).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


iifawzi:
Hello @gunnarmorling, this PR brings binary predicates and InBinary predicates to the drain side, I left the dictionary part out so it can land as part of #859. 

The matchers follow the usual per-operator shape, however, after the first iteration, just following normal matcher design, BinaryIn turned out to be slower than the record matcher for short values, because of how Arrays.equals works: from 8 bytes up it compares 8 bytes at a time (vectorizedMismatch), but below 8 bytes it compares byte by byte. The short variant avoids that for literals of at most 8 bytes. It reads each value's first 8 bytes as one long, masks off the bytes past its length, and compares length and value against each member without a branch.

It's a separate matcher because in the first attempt, putting that logic as an early branch inside the regular binary matchers caused the long cases to regress (long NOT IN about 13% slower). Checking it with PrintInlining (with Claude's help), the per-row comparison call inside the byte-wise loop stopped being inlined, which's because C2 profiles per class and the short-literal calls that returned early made that call site look cold, and C2 only inlines a cold call site up to MaxInlineSize bytes, which wasn't holding there. Keeping the short path in its own class leaves the byte-wise matchers' profile untouched, saving the wins of both sites

gunnarmorling:
Hey @iifawzi, could you rebase to latest main to address the merge conflicts?

iifawzi:
> Hey @iifawzi, could you rebase to latest main to address the merge conflicts?
Hello Hello, rebased and fixed

gunnarmorling:
Great stuff, @iifawzi! Earlier today I've added support for another TIMESTAMP physical mapping, now also based on byte arrays, which required some adjustments to the "can be filtered drain side" decision. So I'm gonna close this one in favor of #1237, which has your commit, rebased to main, and my amendments. Thanks a lot!

Issue #1229, opened 2026-09-15
Cut filler phrasing from the docs and check for it in the PR build


gunnarmorling:
A survey of `docs/content/` found a recurring set of low-signal phrasing: sentences that evaluate instead of stating a fact (\"genuinely different\", \"convenient\", \"useful for …\"), lead-ins that announce a point instead of making it (\"Two consequences are worth knowing.\", \"once you hold one idea:\"), page-kind labels and duplicated cross-links in intros, filler intensifiers (\"actually\", \"simply\", \"just\"), the \"Want to run it or explore the capabilities yourself?\" callout on ten pages, paragraphs repeated between a how-to and its reference page, and a high density of em-dash asides on several pages.

Scope:
- Rewrite those passages so each sentence states something a reader can check or act on.
- Give duplicated facts one home, linking to it from the other pages.
- Bring em-dash density on the densest pages down.
- Add a small check to the PR build that flags the mechanical patterns, so they don't come back.
- Add the prose rule to CLAUDE.md's Documentation section.

Issue #921, opened 2026-08-13
Support FLBA(12) timestamps per the extended-precision timestamp proposal


gunnarmorling:
parquet-format allows a second physical carrier for the `TIMESTAMP` logical type after its 2.14.0 release ([apache/parquet-format#601](https://github.com/apache/parquet-format/pull/601), from the proposal [*Extended Precision Timestamps in Parquet via FLBA(12)*](https://docs.google.com/document/d/1H43RQZhWKcg9c4tJO5W87YxhbUxN7qGSFxJBrrwWOfY/edit)):

- `FIXED_LEN_BYTE_ARRAY(12)` — 96 bits, a **signed two's complement little-endian** integer counting the logical type's `TimeUnit` since the Unix epoch.
- All three `TimeUnit` values and both `isAdjustedToUTC` settings are valid, exactly as for the `INT64` form.
- The sort order is the signed value; the deprecated `TIMESTAMP_MILLIS` / `TIMESTAMP_MICROS` converted types annotate `INT64` only.
- No new encodings: `PLAIN`, `RLE_DICTIONARY`, `DELTA_BYTE_ARRAY` and `BYTE_STREAM_SPLIT` already apply to FLBA.

It exists because `INT64` nanos spans only ~585 years, which cannot cover the ANSI SQL `TIMESTAMP(9)` range of years 0001–9999. It is unrelated to the deprecated `INT96`, which packs a Julian day plus nanos-of-day rather than a flat count.

apache/parquet-testing ships `flba12_timestamp.parquet` with this carrier. Hardwood drops the annotation on it and reads the columns as opaque bytes, and no parquet-java release up to 1.18.1 opens the file (support is in apache/parquet-java#3680), so the parquet-testing comparison run skips it.

### Scope

- **Read:** every accessor path (flat, nested, lists, `getValue`) decodes to `Instant` / `LocalDateTime`; a value outside their range throws rather than wrapping. A legacy timestamp converted type on a non-`INT64` column is dropped.
- **Filter:** per `_designs/PREDICATE_LITERALS.md` — `Instant` / `LocalDateTime` literals on every operator, `byte[]` for equality and membership only; statistics, page index, dictionary and Bloom filter compared in the signed little-endian order.
- **Write:** the row and column writers accept the carrier, statistics are recorded in signed order, and the footer carries the union member without a converted type.
[...]

PR #1231, opened 2026-09-15
#921 #1229 Logical-type-first type tables and a prose pass over the docs

gunnarmorling:
Two related docs changes.

**#921 follow-up.** With `TIMESTAMP` gaining a `FIXED_LEN_BYTE_ARRAY(12)` carrier, the predicate literal table in `reference/query-controls.md` repeated each timestamp annotation across physical-type rows. That table now leads with the logical type and puts each physical carrier on its own line (`<br>`), giving one row per annotation variant. The accessor table (`reference/accessors.md`) and the row-setter table (`reference/writer.md`) get the same one-carrier-per-line treatment.

**#1229.** A survey of `docs/content/` for low-signal phrasing, applied in four commits:
- Sentences framed as "not X, but Y" where nobody would assume X are stated plainly, including one introduced by #1224.
- Sentences that list what a feature is useful for, announce a point before making it, restate as a slogan, or label the page's own kind are cut. The "Try it yourself" callouts stay as they are.
- Passages duplicated between pages (key-value metadata in both write how-tos, the S3 retry message format, the multi-file schema rules) now live in one place.
- Em-dash asides on the densest pages (up to 37 per 1,000 words) are rewritten as sentences, colons or parentheses. The CLI's `—` absent-value marker and empty table cells are untouched.

The last commit adds `tools/docs-prose-check.py`, a dependency-free check run as a `docs-prose-check` job in the PR build, and adds the prose rule to CLAUDE.md's Documentation section. It matches over whole paragraphs, list items, headings and table cells, case-insensitively, and flags filler intensifiers, announcing lead-ins, signposts, self-labelling pages, questions to the reader (outside `!!! example` callouts), and more than 15 em-dashes per 1,000 words on a page.

The docs site builds and the check passes on the result.


docs/content/concepts/concurrency-model.md
```diff
@@ -11,7 +11,7 @@
 -->
 # The Concurrency Model
 
-Hardwood is multi-threaded at its core: a single reader, driven by a single consumer thread,
+Hardwood is multi-threaded: a single reader, driven by a single consumer thread,
```
gunnarmorling: should stay.

docs/content/concepts/nested-columns.md
```diff
@@ -18,8 +18,7 @@ model** is how `ColumnReader` expresses that structure without boxing or per-row
 explains the model; for worked code against it, see
 [Column-Oriented Reading](../how-to/column-reader.md).
 
-!!! example "Try it yourself"
-    Want to run it or explore the capabilities yourself?
+!!! example "Runnable examples"
```
gunnarmorling: Should stay.

docs/content/how-to/column-reader.md
```diff
@@ -16,8 +16,8 @@ The `ColumnReader` provides batch-oriented columnar access with typed primitive
 !!! warning "Experimental API"
     The `ColumnReader` is under active development; the shape of the batch accessors and layer representation may change in future releases without prior deprecation.
 
-!!! example "Try it yourself"
```
gunnarmorling, line 19: Should stay. Same for the other occurances, this was a established pattern.

docs/content/how-to/row-reader.md
```diff
@@ -11,10 +11,10 @@
 -->
 # Row-Oriented Reading
 
-The `RowReader` provides a convenient row-oriented interface for reading Parquet files with typed accessor methods for type-safe field access.
+`RowReader` reads a Parquet file one record at a time, returning each field through a typed accessor addressed by name or index.
```
gunnarmorling: Should stay.

docs/content/reference/cli.md
```diff
@@ -11,7 +11,7 @@
 -->
 # CLI
 
-The `hardwood` CLI lets you inspect and convert Parquet files from the command line — useful for exploring datasets, debugging file structure, and quick format conversions without writing Java code. It reads local files and S3 URIs, and ships as a GraalVM native binary with instant startup.
+The `hardwood` CLI inspects and converts Parquet files from the command line. It reads local files and S3 URIs, and ships as a GraalVM native binary with instant startup.
```
gunnarmorling: here we should mention non-interactive (scripting, agents) and TUI (interactive usage) modes.

docs/content/how-to/s3.md
```diff
@@ -163,7 +163,7 @@ S3Source source = S3Source.builder()
 
 Set `maxRetries(0)` to disable retries entirely.
 
-When the retry budget is exhausted, the last failure is re-thrown as an `IOException` to the caller. For HTTP errors the message has the form `GET s3://<bucket>/<key> failed: HTTP <status>`; for network errors the original `IOException` message is preserved.
+When the retry budget is exhausted, the last failure is re-thrown as an `IOException`; its message is described in the [S3 Reference](../reference/s3.md#s3source-builder-options).
```
gunnarmorling: is it even worth stating this here?

docs/content/concepts/concurrency-model.md
gunnarmorling: Restored "multi-threaded at its core", and dropped that phrase from the prose check.

docs/content/concepts/nested-columns.md
gunnarmorling: Restored the "Try it yourself" callout here.

docs/content/how-to/column-reader.md
gunnarmorling: Restored the "Try it yourself" callout with its opening question on all ten pages, including the tutorial's "Want to run this as a standalone example?". The prose check exempts the body of an `!!! example` admonition from the question rule, and the CLAUDE.md rule names the callout as the exception.

docs/content/how-to/row-reader.md
gunnarmorling: Restored the original intro sentence.

docs/content/reference/cli.md
gunnarmorling: The intro now says the commands run non-interactively, for scripts and AI coding agents (linking the agents section), and that `hardwood dive` opens an interactive terminal UI for exploring a file by hand.

docs/content/how-to/s3.md
gunnarmorling: No. The S3 reference owns it (options table plus the paragraph below it), and the how-to task doesn't need it, so the sentence is gone rather than linked.

PR #1237, opened 2026-09-15
#485 Filter byte-array leaves on the drain side

gunnarmorling:
## Summary

Part of #485. Supersedes #1210 by @iifawzi, rebased onto current main with a follow-up commit on top; @iifawzi's commit is kept as authored.

**`#485 Filter byte-array leaves on the drain side`** (@iifawzi) makes `BinaryPredicate` and `BinaryInPredicate` leaves drain-eligible. Byte-array matchers compare each value's slice in place, in the order and with the equality the literal's `Comparison` gives (unsigned byte strings, signed fixed-width decimals, sign-extending variable-width decimals whose equality compares by value). A separate `BinaryShortInBatchMatcher` decides byte-exact equality against literals of at most eight bytes as one masked big-endian `long`.

**`#485 Decide a binary leaf's batch eligibility by its slice order alone`** fixes a semantic conflict with #921, which landed after #1210 was opened and merges without textual conflicts. #921 added `Comparison.FIXED_TIMESTAMP` for `TIMESTAMP` over `FIXED_LEN_BYTE_ARRAY(12)`. Such a column passed #1210's physical-type guard, so its leaves reached matchers that have no little-endian order: a filter on it either threw at query time (780 `PredicatePathAgreementTest` cases) or, answered as a byte order, would have returned wrong rows. `BinaryComparator.sliceOrder(Comparison)` is now the single, default-free switch that both the compiler's eligibility check and the matchers read, with the instant orders (`FIXED_TIMESTAMP`, `INT96_INSTANT`) mapping to `NONE` and falling back. A future `Comparison` fails to compile in that one place. Every column a binary leaf can name batches as its stored bytes, so the physical-type guard is gone, and equality on an `INT96` column's stored bytes also runs on the drain side.

### Performance

`RecordFilterBenchmarkTest`, 10M rows, alternating main/branch runs (two rounds × 5 iterations). Each contender reads a fixed base projection (`id, value, tag, flag`) plus its filter's columns. Row counts identical on every contender.

| Contender | main (ms) | this PR (ms) | Speedup |
|---|---:|---:|---:|
| `category = 'cat_42'` | 179.0 | 76.3 | 2.34x |
| `category < 'cat_50'` | 242.9 | 152.6 | 1.59x |
| `value < 500 AND category < 'cat_50'` | 241.7 | 150.9 | 1.60x |
| `category IN` (5 values) | 372.1 | 99.1 | 3.76x |
| `category NOT IN` (3 values) | 539.2 | 121.5 | 4.44x |
| 32-byte string eq / IN / NOT IN (`url_tail`) | 196.6 / 412.3 / 551.3 | 123.2 / 354.1 / 258.2 | 1.60x / 1.16x / 2.13x |
| 32-byte string eq / IN / NOT IN (`url_head`) | 187.1 / 351.8 / 515.7 | 106.2 / 256.6 / 221.5 | 1.76x / 1.37x / 2.33x |
| `DECIMAL` on `FIXED_LEN_BYTE_ARRAY(8)` range / eq / IN | 103.7 / 11.2 / 32.1 | 55.5 / 5.0 / 9.9 | 1.87x / 2.25x / 3.23x |
| `DECIMAL` on `BYTE_ARRAY` range / eq / IN | 98.9 / 10.8 / 24.1 | 46.7 / 7.4 / 22.4 | 2.12x / 1.45x / 1.08x |

Numeric contenders are within 5% of main (unchanged code); the no-filter baseline is identical (38.0 ms).

Measured in #1210 on its original base; the follow-up commit does not touch the matchers' loops.

### Notes

- Out of scope and still open under #485: dictionary-space evaluation (#859), `Float16Predicate`, `GeospatialPredicate`, nested leaves, non-projected predicate columns, and a slice comparison for `FIXED_TIMESTAMP`.
- `NOT IN` lowers to a conjunction of `NOT_EQ` leaves, each a separate pass through `AndBatchMatcher`; fusing it into one negated membership pass is the byte-array counterpart of #454.


gunnarmorling:
Hey @iifawzi, as I was wrapping up the predicate work today and added support for a new TIMESTAMP physical mapping, your PR #1210 needed another rebase, and more critically, conflict resolution, so I'm creating this one with your change and my amendments. Claude does a good job of explaining why it was needed:

> His PR decided whether a byte-array filter could run on the drain side by physical type: BYTE_ARRAY and FIXED_LEN_BYTE_ARRAY in, INT96 out. That worked when he wrote it, because INT96 was the only byte-backed column that doesn't sort by its bytes.
>
> #921 added a second such column. Its timestamps are 12-byte little-endian integers, so their byte order doesn't match time order. Being FIXED_LEN_BYTE_ARRAY, they passed his check, so a filter like ts < X was sent to matchers that only compare big-endian bytes, signed or unsigned.
>
> - Merged as-is: the code doesn't compile. The matchers' order switch has no case for the new FIXED_TIMESTAMP comparison.
> - With the obvious one-line fix (make it throw, as INT96 does): it compiles, but timestamp filters that work on main throw at query time. That's the 780 failing PredicatePathAgreementTest cases.
> - With the other obvious fix (treat it as signed or unsigned): it compiles, and range filters on those timestamps silently return wrong rows.
>
> The underlying cause was that two places decided eligibility, and a new column type could get through one while failing the other.

With that out of the way, do you think we can merge this one?

iifawzi:
Hey @gunnarmorling nice catch, and thanks for taking care of adapting it, it looks good to me

iifawzi (approved):

gunnarmorling:
Rebased and applied; thanks a lot, @iifawzi!

Issue #1244, opened 2026-09-17
The PR build serializes four jobs to pass jars between them


gunnarmorling:
The PR build's wall clock is set by a chain of four jobs that run one after another, not by how long any of them takes. `build-core` -> `build-s3` -> `build-cli` -> `Verify native CLI build` each wait for the one before it only to get its jars, downloaded as a `maven-repo-*` artifact.

From run [35020637896](https://github.com/hardwood-hq/hardwood/actions/runs/35020637896), 369 s end to end:

| Job | Starts | Runs |
|---|---|---|
| `build-core` | 0 s | 110 s |
| `build-s3` | 112 s | 52 s |
| `build-cli` | 166 s | 62 s |
| `Verify native CLI build` | 230 s | 139 s |

The last job in the chain spends four minutes waiting for jars that cost seconds to produce. A `-Dquick` build of a job's dependency closure takes 7-11 s locally, against 43 s for the same modules with tests and Error Prone, because `-Dquick` skips the tests and the `qa` profile. The jars are the same either way: `-Dquick` changes what is verified, not what is compiled, and every check it skips still runs in the job that owns the module.

## Proposal

Give every job a `./mvnw -ntp -B -Dquick install -pl <its modules> -am` step and drop its `needs:`, along with the `maven-repo-*` uploads and downloads. `-am` bounds each build to that job's own closure, so no job compiles or resolves the dependencies of modules it does not use. The two BOMs have to be named in `-pl` explicitly, because `-am` does not follow an imported BOM.

The build's wall clock then becomes its slowest single job rather than the sum of a chain: about 165 s, set by `Verify native CLI build` and its 105 s native-image step.

`compile-performance-tests` needs care. `-pl :hardwood-performance-testing -amd` selects an aggregator whose own closure is empty, and `-amd` adds its children only after `-am` has run, so a single `-am -amd` command compiles the benchmarks without core in the reactor and picks up whatever `hardwood-core` the cache happens to hold. It keeps a separate dependency step.

## Cost

Each job repeats a few seconds of compilation that used to happen once, and 13 jobs start at the same time where the graph used to stagger them. Public repositories get 20 concurrent runners, so one PR fits and two in flight do not.

[...]

PR #1246, opened 2026-09-17
SCRATCH: measure three native CLI build variants (do not merge)

gunnarmorling:
Scratch branch for #1244. Not for merge, and it will be deleted once read.

The PR build is cut down to three copies of the native CLI job so a control, `-Dgroups=native` plus an explicit image-builder heap, and `-Ob` run side by side, sharing a runner pool and a moment in time. Every other job is removed, including the predicate audit.


gunnarmorling:
Measured: `-Dgroups=native` and `-J-Xmx6g` are wins and have gone into #1245; `-Ob` came out 11% slower over two attempts and is rejected. Numbers are on #1245. Closing — this branch was only ever a measurement harness.

PR #1247, opened 2026-09-17
Cut the predicate audit's CI time, fixing two reader defects it exposed

### LINKED ISSUE #1241: The predicate audit takes 4-8 minutes of every PR build that runs it
The `predicate-audit` job is the longest job of the PR build whenever it runs, and it runs for every change under `core/src/main`. Over the last 14 runs its `Run the predicate audit` step took between 224 s and 509 s, median 389 s. The spread is 2.3x, so a change to the reader can add anything from four to nine minutes to the PR build depending on which runner it lands on.

Example: <https://github.com/hardwood-hq/hardwood/actions/runs/35264106892/job/105346824503?pr=1240>.

## Where the time goes

From run [35023386244](https://github.com/hardwood-hq/hardwood/actions/runs/35023386244), whose step took 407 s:

| Phase | Wall |
|---|---|
| Build Hardwood and the tool | 36 s |
| Write the fixtures with parquet-java | 2 s |
| Derive footer-rewritten variants | 0.4 s |
| **Matrix** | **352 s** |
| Resolver | 0.2 s |
| Consultation checks | 0.6 s |
| Accessor round-trip | 8 s |
| Engines | 7 s |

The matrix is 96 % of the audit. Everything else together is under 20 s.

## What the matrix costs

The matrix puts 73,454 cells through 5 read paths, so 367,270 reads. Each read builds a `RowReader` or `ColumnReader` on an already-open `ParquetFileReader`, reads, and closes it. The fixtures hold 600 rows and the reads project `__row__` plus the filter's column, so a read decodes almost nothing: at the median that is ~1.06 ms wall and roughly 4 ms of thread time per read. The cost is per-read setup and teardown, not decoding.

Two structural suspects, both unmeasured:

- **Fan-out granularity.** `Matrix.run` submits one task per (group, layout), 27 in all, but 20 of them carry ~4,300 cells and 7 carry 145–750. On a 4-vCPU `ubuntu-latest` runner that is five rounds of heavy tasks with a single-task tail, against 73,454/4 if the cells were split evenly.
- **Nested pools.** The audit's own pool is `availableProcessors()` threads, and each of those drives a reader that submits to `HardwoodContext`'s own `availableProcessors()`-sized executor and starts a pair of virtual threads per `ColumnWorker`. Four audit threads times per-read worker threads on 4 vCPU is oversubscribed, which would explain the cost per read and the 2.3x run-to-run spread.

## Scope

Two of the three causes turned out to be reader defects rather than audit ones, and are split out:

- **#1242** — `RowReader` throws mid-iteration when a filter references a column outside the projection. Blocks the projection change below.
- **#1243** — batch sizing ignores the rows available, so each of the audit's reads allocates batches hundreds of times the fixture's row count.

What stays here, once those land:

1. Project `__row__` plus the predicate's columns instead of every column, which removes 32 workers, 32 exchanges and 64 virtual threads per `RowReader` build over the `flat` group.
2. Silence the audit's `MinMaxStats` warnings. One run writes over 112,000 of them through `StreamHandler.publish`, which is synchronized, so every audit thread serialises on that lock and 35 MB goes to `audit-stderr.log`.
3. Fan out the matrix over predicates rather than over (group, layout), and size the audit's pool against `HardwoodContext`'s instead of both taking every core. Secondary, and worth revisiting once the reader fixes have moved the profile.
4. The audit's coverage must not shrink to buy the time. A cell dropped is a rule the audit stops checking, and `baseline.tsv` has to come out of the change unchanged.

## Done when

The `Run the predicate audit` step is under two minutes at the median, `baseline.tsv` is unchanged, and `tools/predicate-audit/README.md` states the new runtime.

### LINKED ISSUE #1242: RowReader throws mid-iteration when a filter references a column outside the projection
A `RowReader` whose filter references a column outside its projection throws from inside `hasNext()`, and whether it throws depends on the data.

```
projection = ["zz"], filter on "__row__", which is not projected

gt(__row__, 10000)  ->  OK, rows=0                     statistics prune every row group
gt(__row__, 300)    ->  IllegalArgumentException       a row group needs row-level evaluation
```

```
java.lang.IllegalArgumentException: [flat_single.parquet] Column not in projection: __row__
	at dev.hardwood.internal.reader.FlatRowReader.resolveIndex(FlatRowReader.java:998)
	at dev.hardwood.internal.reader.FlatRowReader.isNull(FlatRowReader.java:497)
	at dev.hardwood.internal.predicate.RecordFilterCompiler.lambda$26(RecordFilterCompiler.java:414)
	at dev.hardwood.internal.reader.FlatRowReader.hasNextMatching(FlatRowReader.java:393)
	at dev.hardwood.internal.reader.FlatRowReader.hasNext(FlatRowReader.java:352)
```

The nested reader has it too, with its own message:

```
projection = ["l"], filter on "s.x"
gt(s.x, 10000)  ->  OK, rows=0
gt(s.x, 3)      ->  IllegalArgumentException: Field 's' not in projection
```

When row-group and page statistics settle the predicate, the record matcher never runs and the read succeeds. As soon as one row group needs row-level evaluation it throws, after the reader was built and partway through the results. The same code therefore works against one file and fails against the next.

## Why

`RecordFilterCompiler` compiles a predicate into matchers that call the reader's own accessors (`RecordFilterCompiler.java:410-415`). The name-keyed ones go through `FlatRowReader.resolveIndex`, which looks the column up in `nameToIndex`, built over the projected schema. The record-level path can only see projected columns.

The `ColumnReader` resolves this at `ParquetFileReader.java:641` by carrying two schemas: `augProjected` for what is decoded, `payloadProjected` for what is exposed. `augmentWithPredicateColumns` (`:694`) builds the first. The `RowReader` path at `:519` builds one schema from the caller's projection and no more.

The seam for the fix is already there: `RecordFilterCompiler.compile` takes a `topLevelFieldIndex` callback, wired to `projectedSchema::toProjectedIndex` at `FlatRowReader.java:303`. It returns `-1` for an unprojected column, which is what selects the name-keyed leaf that throws. A projected schema augmented with the predicate's columns makes it return a real index, and the indexed accessor path never consults `nameToIndex`.

## Documented behaviour

`docs/content/how-to/column-reader.md:55` states the contract the `ColumnReader` honours:

> The predicate may reference the column being read, another column, or a column that is not read at all. Predicate columns that are not part of the projection are decoded internally to evaluate the filter but are not exposed.

`docs/content/how-to/query-controls.md:247`, in a section whose worked example is a `RowReader` and which `how-to/index.md:18` says applies to both reader types:

> The filter column does not need to be in the projection; Hardwood reads the filter column's statistics for pushdown regardless.

## Scope

Decode the predicate's columns without exposing them, so both readers mean the same thing by a projection. The augmented columns must stay out of `nameToIndex` and out of whatever else enumerates a row's fields, or the projection stops bounding what a caller can read.

Found while measuring #1241, which needs this to narrow the audit's projection.

### LINKED ISSUE #1243: Batch sizing ignores the rows available, so a small read allocates batches hundreds of times its row count
`BatchSizing.computeOptimalBatchSize` derives a batch size from a 6 MB value-memory budget and the projected columns' widths, clamped to `MAX_BATCH = 524288`. It has no term for how many rows the read can produce, so a file smaller than one batch still gets a batch sized for the budget.

Measured against a 600-row fixture by calling the function with its real schema:

| Projection | Columns | Batch size | Oversize |
|---|---|---|---|
| all columns | 34 | 22,153 | 37x |
| one `INT64` column | 1 | 524,288 (`MAX_BATCH`) | 874x |
| two columns | 2 | 524,288 (`MAX_BATCH`) | 874x |

`BatchExchange.recycling` (`BatchExchange.java:109`) then eagerly allocates three batches per column, each through `allocateArray` (`:329`): `new long[524288]` is 4 MB, allocated and zeroed to carry 600 values. The single-column `ColumnReader` path uses `detaching` (`ColumnReader.java:919`) and allocates a fresh one per batch produced.

Sampling a workload of many small reads (#1241's audit, `flat` group, 2400 cells, JFR at a 1 ms period) puts `BatchExchange.allocateArray` and `FlatRowReader.lambda$1`, the batch factory at `FlatRowReader.java:273`, as the top two frames at 16 % of samples between them. Capping the batch at the row count removes both from the profile.

## The fix

`min(computed, rowsAvailable)`.

The cap binds only when the whole read fits inside one batch, and in that case the batch already holds every row there is — the allocation reserves space for rows that cannot arrive. For any read larger than a batch, `min` returns the computed size and nothing differs. It cannot regress a large read.

Capping is safe: `FlatColumnWorker:95` computes `spaceInBatch = batchCapacity - rowsInCurrentBatch` and flushes at `:114`, so capacity is an upper bound on a batch's contents and never a floor. `FlatRowReader:58`'s `ALL_PRESENT` sentinel stays sized to `MAX_BATCH`.

Both call sites already hold the metadata. `ParquetFileReader.buildRowReader` (`:515`) takes `firstFileRowGroups` and forwards it to `NestedRowReader.create` for fan-out sizing; `FlatRowReader.create` is the one signature that does not take it. The `ColumnReader` path already threads `rowGroups` into `resolveBatchSize` (`:661`).

Three things decide the bound:

- **Batches span row groups.** `FlatColumnWorker.flushOnFilterAlwaysMatchesTransition` exists to force a flush at a row-group boundary when a filter is installed, which means batches otherwise run through boundaries. The bound is total rows across the row groups the read visits, not the largest row group.
- **Multi-file reads.** The variable is `firstFileRowGroups` and `inputFiles` (`ParquetFileReader.java:82`) is a list. Capping a multi-file read by the first file's rows stays correct but shrinks batches for every later file. Apply the cap only where the bound covers the whole read.
- **`head(N)` under a filter.** `maxRows` counts matching rows, so the reader may scan far more than `N`. The cap comes from available rows, not from `maxRows`.

A bound that is not an upper bound on the read is the only way this change can cost anything, so that is what the tests need to pin.

Same class as #748, where `BatchSizing` computed from one dimension of the input and ignored another.

Found while measuring #1241; it is a reader-wide defect and not specific to that tool.

gunnarmorling:
The `predicate-audit` job was the PR build's long pole whenever it ran: 224-509 s over the last 14 runs, median 389 s, of which the matrix was 96 %. Measuring it turned up two reader defects, which are the first two commits here; the audit change is the third and depends on the first.

Closes #1242, closes #1243, closes #1241.

## #1242 Decode a row reader's predicate columns without exposing them

A `RowReader` whose filter referenced a column outside its projection threw from inside `hasNext()`, and only where a row group needed row-level evaluation. Where statistics settled the predicate the read succeeded, so the same code worked against one file and failed against the next.

`ParquetFileReader` now resolves the projection twice for a filtered row read, as the column-reader path already did: `ProjectedSchema.createAugmented` appends the predicate's leaves after the projected ones, so every exposed column keeps the index it would have had and `RecordFilterCompiler`, which addresses columns by projected index, is untouched.

The projection bounds the row's fields: `getFieldCount()` and `getFieldName(int)` cover the projection alone, so a schema-driven consumer sees nothing else. A name-keyed accessor reaches any decoded column, so a predicate column outside the projection reads its value. Half of `RecordFilterCompiler`'s leaf kinds, and every nested path, resolve a column by name through those same accessors, so bounding them too would mean giving the matcher an accessor surface of its own. `_designs/ROW_READER_AUGMENTED_PROJECTION.md` has the model.

## #1243 Bound the batch size by the rows a read can produce

`BatchSizing.computeOptimalBatchSize` sized a batch from a 6 MB budget and the projected columns' widths, with no term for how many rows the read could produce. A single `INT64` column over a 600-row file budgeted `MAX_BATCH`: a 4 MB `long[]` per batch to hold 600 values, three of them per column from the recycling exchange.

The size is now capped at the rows available. The cap binds only where the whole read fits in one batch, which is the case where the batch already holds every row there is, so no batch a read could have filled is shortened and a large read is sized exactly as before. The bound is the first file's row count, since pruning, a row limit and a skip only remove rows; a multi-file read passes `ROWS_UNKNOWN`, because its later files are not planned when the batch is sized.

Measured on a workload of many small reads (the audit matrix over `flat`, 2400 cells, one thread, JFR at a 1 ms period): 24.1 s to 8.3 s, with `BatchExchange.allocateArray` and the batch factory gone from the top of the profile, where they had been the top two frames.

## #1241 Read one column per audit cell

The matrix's three row-reader paths projected every column of the fixture to read `__row__`, so each cell stood up 34 workers, 34 exchanges and 68 virtual threads over the `flat` group. They now project `__row__` alone and let the reader decode the predicate's columns beside it, which is what #1242 makes possible.

The fixtures also carry NaN bounds and sort orders the reader does not read, by design, so every read reported statistics it would not prune by: more than 100 000 records per run through one synchronized handler, and 35 MB of stderr.

Fanning out over predicates rather than over (group, layout), and sizing the audit's pool against the context's, are both in #1241's description and both turned out not to be worth doing. The job now spends more time building Hardwood than auditing it.

## Result

| | Matrix | Full audit |
|---|---|---|
| `main`, CI | 352 s | 367 s |
| this branch, four cores | 14 s | 20 s |

`baseline.tsv` is unchanged and the matrix still puts all 73 454 cells through five read paths.

## Testing

`./mvnw verify -DskipITs` is green, and the audit runs clean against its baseline. The integration tests have not run locally — this container has no Docker daemon — and this change touches the read path they cover, so the PR build is the first run of them.


gunnarmorling:
Confirmed on this PR's own build: the `Run the predicate audit` step took **83 s** (20:37:33 to 20:38:56 in [run 35271899340](https://github.com/hardwood-hq/hardwood/actions/runs/35271899340)), against a median of 389 s and a worst case of 509 s over the 14 runs measured in #1241. Most of what is left is building Hardwood, not auditing it.

The integration tests passed on all three JDKs, which is their first run for this change: the dev container has no Docker daemon, so they could not run locally.

PR #1251, opened 2026-09-18
#1047 Discard the output of a failed writer, and add abort()

### LINKED ISSUE #1047: close() after a failed write publishes the rows that got through, and there is no way to abandon a writer

gunnarmorling:
Builds on #1075 by @mohamedibrahim54, whose commit is kept as the first of two, rebased on main.

A write that throws fails the writer: `ColumnWriter.writeBatch` and `RowWriter.writeRow` reject further data, and `close()` discards the output instead of publishing the rows written before the failure. This covers both a rejected batch and a failure while writing (a destination `IOException`, a codec `ParquetWriteException`). A record `RowWriter` rejects before staging it still leaves the writer usable, as the row API documents.

A failure the writer never sees, such as one in the caller's data source, would otherwise let try-with-resources publish a truncated file. The new `ParquetFileWriter.abort()` discards the output for that case; a later `close()` does nothing. This is the same shape as Hadoop's `Abortable` and object_store's `MultipartUpload::abort`.

Compared to #1075, the `WriteFailurePolicy` / `COMMIT_PREFIX` option is dropped. Committing the prefix cannot work on object storage: a multipart upload with a failed part cannot be completed, and with concurrent part uploads the failure can surface on a later call than the one that lost the bytes. A discard that fails is now thrown rather than swallowed, since it leaves a temporary file or an uncompleted upload behind.

Docs: new how-to "Handle Write Failures", a "Finishing and Abandoning a File" section in the writer reference, and the failure bullet in the write model.

Closes #1047

## Authorship

Check this box to confirm:

## Checklist


mohamedibrahim54:
This approach LGTM! However, does the abort method really need to be public? I'm trying to think of a scenario where it would be useful. Since close already discards the file when the state is failed, a public abort might be unnecessary. although it give user more control to manually discard the file.

gunnarmorling:
> I'm trying to think of a scenario where it would be useful.

Hey @mohamedibrahim54, fair question; `abort()` needs to be public if there's failures higher up in the call stack, which aren't visible as failures in the writer itself. Say, a producer reads data from Kafka or a database, and that connection dies before the output file has been completely written. Or it has a given timeout and doesn't manage to write the file before that timeout. In cases like this, the write so far was fine from a Hardwood side, but the caller needs a way to abort it. This is what the method gives. Other, higher-level writers like Spark or Flink also have abort routines (e.g. when canceling a job), so without `abort()`, their only way for aborting would be to make a write fail on purpose, which feels hacky.

gunnarmorling:
Applied; thanks a lot, @mohamedibrahim54!

PR #1257, opened 2026-09-20
#859 implement dictionary predicate evaluation

shril:
## Authorship

Check this box to confirm:

## Checklist

## Summary

Implements the binary portion of #859: dictionary-encoded binary values are compared once per referenced dictionary entry, then matching rows are answered from their dictionary IDs.

This is the deliberate replacement for #972, not an independent implementation which overlooks it.

#972 originally added a standalone dictionary-aware equality matcher. During review, two changes were requested:

1. Decide dictionary entries lazily, instead of scanning the complete dictionary when first encountered.
2. Extend dictionary-space evaluation to every binary operator by delegating entry comparisons to the drain-side matchers being developed in #1210.

#1210 was subsequently closed in favor of #1237, which retained its binary matcher work and added the comparison eligibility changes required by the new byte-backed timestamp mapping. Once #1237 merged, its matchers superseded the equality implementation in #972. This PR therefore starts from the merged architecture and adds dictionary evaluation as a wrapper around those matchers.

Fixed-width primitive dictionary evaluation remains open under #859.

### Dictionary-aware matcher wrapper

`BatchFilterCompiler` wraps every eligible `BinaryBatchMatcher` in `DictionaryBinaryBatchMatcher`.

`BinaryBatchMatcher` now exposes a scalar operation which tests one non-null byte slice with the same semantics as its optimized whole-batch operation. The existing EQ, NOT_EQ, LT, LT_EQ, GT, GT_EQ, IN, and short-value IN matchers implement that operation without replacing their specialized batch loops.

The wrapper behaves as follows:

- A batch without dictionary metadata goes directly to the delegate's optimized whole-batch loop.
- A dictionary-backed row reads its result from a cached state indexed by dictionary ID.
- A row without a usable dictionary ID is tested from its packed byte slice through the delegate.
- A new dictionary object resets the cache, using dictionary identity as the column-chunk boundary.

This preserves the comparison semantics merged in #1237. Strings, stored bytes, fixed-width decimals, and variable-width decimals all use the delegate's existing definition of equality and order. In particular, padded and minimal encodings of the same variable-width decimal continue to compare by value rather than by byte spelling.

`NOT IN` continues to lower to its existing conjunction of `NOT_EQ` matchers, so each leaf receives dictionary-space evaluation without introducing a second membership implementation.

### Lazy per-entry decisions

This directly addresses the first unresolved review comment on #972.

Each dictionary entry has one primitive state:

- `UNKNOWN`
- `MATCH`
- `NO_MATCH`

For each batch, the wrapper walks only present dictionary IDs whose entries remain unknown. Each referenced entry is evaluated once and its result remains cached for later rows and batches using the same dictionary.

Consequences:

- page-index pruning, row masks, and early termination do not pay to compare unreferenced entries;
- repeated IDs are answered from the cache;
- null rows neither decide entries nor produce matches; and
- the discovery pass stops permanently once every entry has been decided.

A full scan still compares each entry at most once. Until all entries are known, it additionally walks each batch's IDs to discover newly referenced entries.

### Dictionary-index retention

The compiled matcher advertises whether it needs dictionary IDs. `AndBatchMatcher` and `OrBatchMatcher` propagate that requirement from their children.

The flat row-reader and exact column-reader paths compile the filter before allocating binary batches, then retain dictionary IDs only when the selected matcher needs them. This extends retention beyond interned string columns to every byte-array-shaped physical type supported by the binary matchers.

The exact-column path reuses the same compiled filter in `SelectionEngine`, avoiding a second compilation after reader allocation.

Nested leaves remain outside this optimization because nested values do not map one-to-one to top-level records. The nested allocation path rejects an impossible retention request rather than silently discarding metadata.

## Correctness and edge cases

- Null rows are skipped before dictionary lookup or packed-value fallback.
- Plain pages continue through the ordinary per-value matcher.
- A batch containing dictionary and plain values represents plain values with ID `-1`.
- A batch crossing a column-chunk boundary adopts the first dictionary; rows from the second dictionary use packed-value fallback until a later batch adopts it.
- Dictionary identity changes reset all cached outcomes.
- Cache storage is reused across dictionaries while only the active range is cleared.
- Compound same-column `AND` and `OR` matchers retain IDs when either child needs them.
- Variable-width decimal equality and membership retain sign-extending value semantics.
- Unsupported binary orders, including timestamp orders without a slice implementation, continue to use the existing record-filter fallback.
- Fixed-width primitive physical types are unchanged.

## Tests

Coverage includes:

- lazy decisions for referenced entries only;
- repeated IDs across batches, proving one decision per entry and predicate leaf;
- cache reset on dictionary identity changes;
- null rows;
- mixed dictionary and packed-value rows;
- missing retained-index validation;
- output crossing a 64-bit bitmap boundary;
- EQ, NOT_EQ, LT, LT_EQ, GT, GT_EQ, IN, and short-value IN;
- byte-string, stored-byte, fixed-decimal, and variable-decimal comparison modes;
- padded variable-width decimal equality;
- retention requirements for individual leaves and same-column `AND`/`OR`;
- row-reader and exact column-reader agreement with an unfiltered oracle;
- dictionary changes across column chunks; and
- `FIXED_LEN_BYTE_ARRAY` dictionary-ID retention.

Full `./mvnw verify` passes, including the Docker-backed S3 and CLI integration tests.

### JMH

`DictionarySpacePredicateBenchmark` is single-threaded, uses 65,536-row batches, and separates sparse access—four referenced entries—from full access. The measured run used one fork, three one-second warmups, and five one-second measurements.

The available measured run covered cold equality evaluation: a fresh wrapper per invocation, including entry discovery and decisions, against the packed-value equality loop.

| Access | Cardinality | Dictionary space | Packed values | Speedup |
|---|---:|---:|---:|---:|
| Sparse | 16 | 131.4 µs | 285.5 µs | 2.17× |
| Sparse | 4,096 | 117.8 µs | 286.3 µs | 2.43× |
| Full | 16 | 55.8 µs | 251.3 µs | 4.50× |
| Full | 4,096 | 69.7 µs | 258.1 µs | 3.70× |

The benchmark source now also contains cold and hot cross-batch arms for EQ, LT, and IN. The table above reports only the completed measured run; it does not claim results for the subsequently expanded matrix.

### `RecordFilterBenchmarkTest`

The end-to-end comparison reads a 10-million-row, single-row-group Parquet file. Each contender reads the same base projection plus its filter columns. The comparison used five runs per contender on the same tree and JVM, temporarily disabling only the dictionary wrapper for the baseline.

All paired contenders returned identical row counts.

| Contender | Packed matcher | Dictionary space | Ratio |
|---|---:|---:|---:|
| Category EQ | 93.0 ms | 116.1 ms | 0.80× |
| Category range | 135.3 ms | 109.8 ms | 1.23× |
| Category compound | 129.5 ms | 108.1 ms | 1.20× |
| Category IN, 5 values | 90.3 ms | 64.4 ms | 1.40× |
| Category NOT IN, 3 values | 90.1 ms | 79.6 ms | 1.13× |
| 32-byte tail EQ | 92.3 ms | 62.0 ms | 1.49× |
| 32-byte tail IN | 221.0 ms | 68.6 ms | 3.22× |
| 32-byte tail NOT IN | 156.9 ms | 81.6 ms | 1.92× |
| 32-byte head EQ | 81.8 ms | 71.7 ms | 1.14× |
| 32-byte head IN | 191.1 ms | 70.6 ms | 2.71× |
| 32-byte head NOT IN | 131.6 ms | 90.5 ms | 1.45× |
| Fixed-decimal range | 69.0 ms | 55.9 ms | 1.23× |
| Fixed-decimal EQ | 6.0 ms | 7.5 ms | 0.80× |
| Fixed-decimal IN | 10.5 ms | 11.9 ms | 0.88× |
| Variable-decimal range | 53.3 ms | 50.5 ms | 1.06× |
| Variable-decimal EQ | 8.4 ms | 7.6 ms | 1.11× |
| Variable-decimal IN | 22.3 ms | 23.4 ms | 0.95× |

The strongest gains are on repeated long values and membership predicates, where dictionary caching removes repeated multi-byte comparisons. Very selective decimal equality and membership cases are already dominated by page pruning and short execution times, so the added dictionary bookkeeping does not consistently improve them.

The decimal-IN benchmark was also corrected to pass `BigDecimal` operands through the public decimal predicate API. The previous helper converted them to raw two's-complement byte arrays, bypassing schema-aware decimal resolution and, for `FIXED_LEN_BYTE_ARRAY`, the required width padding. The corrected contenders each return 49 rows and exercise the intended fixed- and variable-decimal semantics.

## Documentation

- Adds `_designs/DICTIONARY_SPACE_EVALUATION.md`.
- Updates the drain-side filtering design to describe the implemented wrapper.
- Updates the roadmap to mark binary dictionary-space evaluation complete while retaining fixed-width primitive support as open work under #859.


shril:
cc: @gunnarmorling @iifawzi

iifawzi:
Good morning @shril, thanks for working on it! I'm going to check it tonight

gunnarmorling:
Thanks a lot, @shril!

> Contender 	Packed matcher 	Dictionary space 	Ratio
> Category EQ 	93.0 ms 	116.1 ms 	0.80×

This looks like quite a regression; do we know why happens?

gunnarmorling:
One more general question, instead of the discovery pass, could we replace the lazy tri-state with an eager match table: evaluate every entry once when a dictionary arrives, then look up each row's ID. I think that should simplify things quite a bit.

iifawzi:
> One more general question, instead of the discovery pass, could we replace the lazy tri-state with an eager match table: evaluate every entry once when a dictionary arrives, then look up each row's ID. I think that should simplify things quite a bit.

We started discussing it @ https://github.com/hardwood-hq/hardwood/pull/972#discussion_r3999254225, I believe we could benchmark it so we're backed by numbers, but doing it eagerly seemed quite aggressive based on an earlier discovery that I made, because if, finally, only a few values are being referenced, in the pages that we had to read, we paid the price of the entire dictionary. The tradeoff is that doing it lazily will definitely slow down full scans

gunnarmorling:
Ah, I see, sorry, I had missed that discussion. Gotta admit the lazy calculation seems quite complex to me, which is why I was/am preferring the eager approach. But your point on not all values being required is well taken. Hum, hum. Intuitively, I'd start with the eager solution for the sake of simplicity, but if we have convincing results that the lazy approach is advantageous, I'm open for it.

shril:
Thank you @gunnarmorling and @iifawzi.

My interpretation is that eager evaluation should be the default choice for simplicity unless the lazy approach demonstrates a material advantage on realistic sparse reads.

I’ll benchmark three implementations:
1. the current packed-value matcher;
2. an eager per-dictionary outcome table; and
3. the lazy `UNKNOWN` / `MATCH` / `NO_MATCH` table.

I’ll cover:
1. small and large dictionaries;
2. four referenced entries versus full dictionary coverage;
3. first/cold and later/hot batches;
4. short equality, long byte-string equality, ranges, and `IN`
5. full scans and reads where pruning or early termination touches only a small part of the dictionary.

If eager is close to lazy on sparse reads and better on common full scans, I’ll simplify the implementation to eager. I’ll retain lazy only if it shows a convincing advantage that justifies the additional lifecycle and discovery-pass complexity.

Let me know if this sounds okay to you, and please correct me if I have made any mistake in my understanding.

shril:
> Thanks a lot, @shril!
> 
> > Contender 	Packed matcher 	Dictionary space 	Ratio
> > Category EQ 	93.0 ms 	116.1 ms 	0.80×
> 
> This looks like quite a regression; do we know why happens?

I changed order of execution in benchmarks with Category EQ measured last.

I made three dedicated unmeasured warmups, ten measured runs, median/range reporting, and complete runs in `lazy → eager → packed → packed → eager → lazy` order.

The earlier Category EQ regression did not reproduce:
- packed medians: 86.3 / 86.7 ms
- eager medians: 77.1 / 77.1 ms
- lazy medians: 74.9 / 77.4 ms

Eager and lazy were effectively tied end-to-end for this contender.
JMH confirmed the trade-off. Eager is best for hot/full dictionaries, but on a cold sparse long-`IN` shape over 4,096 rows it measured 58.7 µs versus 8.0 µs for the current lazy approach.

The benchmark also exposed a weakness in my current lazy implementation: its separate discovery pass makes large sparse batches with cheap predicates slower than necessary.

I tested a simpler one-pass lazy form which decides an unknown entry directly in the output loop. It removes the discovery pass and `undecidedEntries`. It measured:
- large sparse short EQ: 65.2 µs, versus eager 64.9 µs and current lazy 119.6 µs;
- sparse long IN: 4.6 µs, versus eager 58.7 µs and current lazy 8.0 µs;
- full short EQ: 70.4 µs, versus eager 57.2 µs and current lazy 59.9 µs.

Based on these results, I propose replacing the current two-pass lazy implementation with this simpler one-pass lazy form. It preserves the sparse-read benefit while addressing both the complexity and extra-pass concerns. Does that direction sound reasonable?

_Disclaimer: I used AI to quickly run these benchmarks._

cc: @gunnarmorling @iifawzi

iifawzi:
Thanks for sharing it @shril. Two questions, what's `packed` approach? and what's the definition of cheap predicates mentioned here `its separate discovery pass makes large sparse batches with cheap predicates slower than necessary.`? also, can we have the benchmark of all contenders, through the different variants? `lazy`, `eager`, `lazy-one-pass`, `packed?` this perhaps makes the decision quite simpler. 

Overall, we're talking about 4x speedups when it comes to sparse Long IN (but is associated with "cheap predicate"), and at best, ~2x on large sparse short EQ, aside from rest of numbers for lazy vs eager are quite similar not to justify the lazy approach if it holds technical complexities as @gunnarmorling pointed out

shril:
Hi @iifawzi, "Packed" means running the ordinary binary matcher over every row’s decoded byte slice. It ignores the retained dictionary IDs, so a value repeated across many rows is compared repeatedly. It is the pre-optimization baseline. _Not the best terminology, I'll avoid future use._

By “cheap predicate,” I meant a low per-entry comparison cost relative to scanning the row-ID array. Specifically, this referred to the short-EQ benchmark, where each decision compares one short value such as "c00042" with one short literal.

shril:
I reran a bounded comparison of all four strategies. `Hot` means the same dictionary and referenced subset were processed once before measurement.
Packed has no cache, so its value is the same baseline in both tables.

### Cold:

| Shape | Packed | Eager | Two-pass lazy | One-pass lazy |
|---|---:|---:|---:|---:|
| Full short `EQ`, 100 entries, 65,536 rows | 364.1 µs | 57.0 µs | 59.6 µs | 69.0 µs |
| Sparse short `EQ`, 4/4,096 entries, 65,536 rows | 402.6 µs | 64.3 µs | 116.9 µs | 64.4 µs |
| Sparse long `IN`, 4/4,096 entries, 4,096 rows | 46.8 µs | 57.5 µs | 7.8 µs | 4.5 µs |
| Early-stop sparse short `EQ`, 4/4,096 entries, 64 rows | 0.406 µs | 14.042 µs | 0.576 µs | 0.562 µs |

### Hot:

| Shape | Packed | Eager | Two-pass lazy | One-pass lazy |
|---|---:|---:|---:|---:|
| Full short `EQ`, 100 entries, 65,536 rows | 364.1 µs | 55.6 µs | 58.7 µs | 67.7 µs |
| Sparse short `EQ`, 4/4,096 entries, 65,536 rows | 402.6 µs | 50.1 µs | 116.5 µs | 64.0 µs |
| Sparse long `IN`, 4/4,096 entries, 4,096 rows | 46.8 µs | 5.0 µs | 6.2 µs | 4.4 µs |
| Early-stop sparse short `EQ`, 4/4,096 entries, 64 rows | 0.406 µs | 0.053 µs | 0.091 µs | 0.084 µs |

This confirms that one-pass lazy is not universally faster. When the full dictionary is resolved, current two-pass lazy is about 15% faster because it no longer runs discovery and does not check `UNKNOWN` per row. For a large partially resolved dictionary, current two-pass lazy is about 1.8 times slower because it repeats discovery while unreferenced entries remain unknown.

One-pass lazy preserves the important cold sparse behavior: in sparse long `IN`, it is about 12.8 times faster than eager cold. It also removes `undecidedEntries` and the separate discovery pass, leaving dictionary identity reset, tri-state storage, and one `UNKNOWN` check in the output loop.

My would prefer the one-pass lazy road, because it avoids eager's whole-dictionary cold cost and the current lazy implementation's repeated discovery scans with less lifecycle complexity. The measured cost is about 15% versus current two-pass lazy in the fully covered short-`EQ` kernel; one-pass takes about 21% longer than eager there.

Please guide if I am missing something here, or if you want to to explore some other mechanism.

cc: @gunnarmorling @iifawzi

iifawzi:
Thanks for sharing the numbers @shril, I'm quietly leaning toward the lazy one pass approach. How do you see that @gunnarmorling?

gunnarmorling:
Thanks a lot for the explorations here! Yes, agreed, one-pass lazy seems like the right approach forward in light of these numbers.

shril:
Thank you @iifawzi and @gunnarmorling.
I'll make changes to implement one-pass lazy approach here.

